PR #265, opened 2019-10-14
Calculate epoch reward prorated

macpie:


src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -263,6 +256,43 @@ get_reward_vars(Ledger) ->
         poc_version => POCVersion
     }.
 
+-spec calculate_epoch_reward(non_neg_integer(), non_neg_integer(), blockchain_ledger_v1:ledger()) -> float().
+calculate_epoch_reward(Start, End, Ledger) ->
+    Version = case blockchain:config(?reward_version, Ledger) of
```
vihu: We should handle the case of `?reward_version` being `{error, not_found}` explicitly I think for clarity. I don't fully understand why it's being defaulted to 1 here?

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -230,31 +230,22 @@ get_txns_for_epoch(Current, End, Chain, Txns) ->
...
-    BlockPerMin = 60/BlockTime1,
-    % Convert to blocks per hour
-    BlockPerHour = BlockPerMin*60,
-    % Calculate number of elections per hour
-    ElectionPerHour = BlockPerHour/ElectionInterval,
+    EpochReward = calculate_epoch_reward(Start, End, Ledger),
     #{
         monthly_reward => MonthlyReward,
-        epoch_reward => MonthlyReward/30/24/ElectionPerHour,
+        epoch_reward => EpochReward,
```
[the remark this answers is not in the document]
macpie: consensus is not prorated look above

src/transactions/v1/blockchain_txn_rewards_v1.erl
Vagabond: We should hold this until my chain var branch lands and add the newly required validation for that var as well.

macpie: cause 1 is the starting point, we use that method and all the other chain vars versions

PR #281, opened 2019-10-24
add cache for valdation of txn at height

evanmcc:
this seems to be the safest/most conservative way to cache this.  it's possible that we could expand it later with some awareness of when something is going to expire, although that makes it a lot harder to figure out cache keys.


evanmcc:
closing this for now, it's not clear that we need it.

PR #286, opened 2019-11-02
Switch to ETS backed cache

vihu:
TODO:
- Cache invalidation
- Handle clauses when there is no chain
- Fix eunits


src/blockchain_score_cache.erl
```diff
@@ -0,0 +1,123 @@
...
+        {error, not_found} ->
+            set({Address, Height}, ScoreFun());
+        {ok, Score} ->
+            Score
+    end.
+
+-spec find({Address :: libp2p_crypto:pubkey_bin(),
+            Height :: pos_integer()}) -> {error, not_found} | {ok, score()}.
+find({Address, Height}) ->
+    gen_server:call(?MODULE, {find, {Address, Height}}, infinity).
```
evanmcc: do the fetch directly here.

src/blockchain_score_cache.erl
```diff
@@ -0,0 +1,123 @@
...
+
+-spec set({Address :: libp2p_crypto:pubkey_bin(),
+           Height :: pos_integer()},
+          Score :: score()) -> score().
+set({Address, Height}, Score) ->
+    gen_server:call(?MODULE, {set, {{Address, Height}, Score}}, infinity).
```
evanmcc: do the set directly here (i.e. no call).

src/blockchain_score_cache.erl
```diff
@@ -0,0 +1,123 @@
...
+%% ------------------------------------------------------------------
+%% gen_server Function Definitions
+%% ------------------------------------------------------------------
+init(_Args) ->
+    ok = blockchain_event:add_handler(self()),
+    Cache = ets:new(score_cache,
+                    [named_table,
+                     {write_concurrency, true},
+                     {read_concurrency, true}]),
+    case  blockchain_worker:blockchain() of
```
evanmcc: wire into blockchain_event here.

src/blockchain_score_cache.erl
```diff
@@ -0,0 +1,123 @@
...
+        undefined ->
+            %% XXX: need to handle clauses when there is no chain
```
evanmcc: you should still save the cache here, and set up a tick to set the chain and register for events later when there's a chain.

src/blockchain_score_cache.erl
```diff
@@ -0,0 +1,123 @@
...
+    true = ets:insert(Cache, {{Address, Height}, Score}),
+    {reply, Score, State};
+handle_call(_Msg, _From, State) ->
+    {reply, ok, State}.
+
+handle_cast(_Msg, State) ->
+    {noreply, State}.
+
+handle_info({blockchain_event, {add_block, _Hash, _Sync, _Ledger}}, State) ->
+    %% Do cache invalidation here?
```
evanmcc: use [select_delete](http://erlang.org/doc/man/ets.html#select_delete-2) to remove anything that's older than the delayed ledger, which is what, block height - 50? 30?

PR #294, opened 2019-11-19
Add support for txn bundles

vihu:
The goal here is to allow bundled txn support.

- The bundled transactions are first speculatively absorbed to check whether they are valid in the order they appear in the bundle.
- Introduce `max_bundle_size` chain var. Default: 5

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -158,3 +158,10 @@
 -define(poc_witnesses_percent, poc_witnesses_percent).
 -define(poc_challengers_percent, poc_challengers_percent).
 -define(dc_percent, dc_percent).
+
+%%%
+%%% bundle txn vars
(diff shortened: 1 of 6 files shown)


src/transactions/v1/blockchain_txn_payment_v1.erl
```diff
@@ -179,7 +179,6 @@ absorb(Txn, Chain) ->
                     blockchain_ledger_v1:credit_account(Payee, Amount, Ledger)
             end
     end.
-
```
Vagabond: Undo unneeded whitespace changes?

vihu:
Rethinking this to allow bundled transactions on chain. Will update.

vihu:
Tests are in miner: https://github.com/helium/miner/pull/256

PR #295, opened 2019-11-22
add direct context cache test, fix cache deletes

evanmcc:
this test should take care of most issues with lagging ledger determinism, but requires a lot of testing; we need to make sure whether or not the insert-witness change means we need to reset all the ledgers, and if so, if there's a way around it.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -361,11 +361,21 @@ fingerprint(#ledger_v1{mode = Mode} = Ledger, Extended) ->
            securities = SecuritiesCF,
            routing = RoutingCF
           } = SubLedger,
-        L = [DefaultVals, GWsVals, EntriesVals, DCEntriesVals, HTLCs,
-             PoCs, Securities, Routings]
-            =  [cache_fold(Ledger, CF, fun(X, Acc) -> [X | Acc] end, [])
-                || CF <- [DefaultCF, AGwsCF, EntriesCF, DCEntriesCF, HTLCsCF,
-                          PoCsCF, SecuritiesCF, RoutingCF]],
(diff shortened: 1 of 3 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1054,37 +1054,26 @@ request_poc(OnionKeyHash, SecretHash, Challenger, BlockHash, Ledger) ->
...
-            ok = cache_put(Ledger, AGwsCF, Challenger, GwBin),
-
-            PoC = blockchain_ledger_poc_v2:new(SecretHash, OnionKeyHash, Challenger, BlockHash),
-            PoCBin = blockchain_ledger_poc_v2:serialize(PoC),
-            BinPoCs = erlang:term_to_binary([PoCBin|lists:map(fun blockchain_ledger_poc_v2:serialize/1, PoCs)]),
-            PoCsCF = pocs_cf(Ledger),
-            cache_put(Ledger, PoCsCF, OnionKeyHash, BinPoCs);
+            ok;
         LastOnionKeyHash  ->
             case delete_poc(LastOnionKeyHash, Challenger, Ledger) of
```
Vagabond: No need for a case here at all now?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1732,12 +1721,7 @@ cache_get(#ledger_v1{db=DB}=Ledger, CF, Key, Options) ->
 -spec cache_delete(ledger(), rocksdb:cf_handle(), any()) -> ok.
 cache_delete(Ledger, CF, Key) ->
     {Context, Cache} = context_cache(Ledger),
-    case ets:lookup(Cache, {CF, Key}) of
```
Vagabond: Nice catch

PR #297, opened 2019-11-25
Keep retrying assert location transactions

vihu:
Far from ideal, but it would make the on-boarding potentially suck less as we'll keep retrying assert location transactions, even after F+1 members have rejected them. This is temporary while we make txn bundles work solid enough.

### DIFF
--- a/src/transactions/blockchain_txn_mgr.erl
+++ b/src/transactions/blockchain_txn_mgr.erl
@@ -239,7 +239,7 @@ rejected(Txn, Member, State=#state{txn_map=TxnMap, chain=Chain}) ->
             F = (N - 1) div 3,
             case length(lists:usort([Member|Rejections])) > F of
                 true ->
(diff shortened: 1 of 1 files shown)


src/transactions/blockchain_txn_mgr.erl
```diff
@@ -239,8 +239,16 @@ rejected(Txn, Member, State=#state{txn_map=TxnMap, chain=Chain}) ->
             F = (N - 1) div 3,
             case length(lists:usort([Member|Rejections])) > F of
                 true ->
-                    %% too many rejections
-                    State#state{txn_map=maps:remove(Txn, TxnMap)};
+                    %% Temporary patch to keep retrying assert location
```
Vagabond:
```suggestion
                    %% TODO XXX Temporary patch to keep retrying assert location
```

PR #298, opened 2019-11-26
mod the fingerprint height so we get more reports

evanmcc:


src/blockchain_worker.erl
```diff
@@ -231,7 +231,10 @@ signed_metadata_fun() ->
                         {error, _} ->
                             #{}
                     end,
-                FPMD = case blockchain_ledger_v1:fingerprint(blockchain:ledger(Chain)) of
+                Ht0 = maps:get(<<"height">>, HeightMD, 1),
+                Ht = max(1, Ht0 - (Ht0 rem 20)),
+                {ok, Ledger} = blockchain:ledger_at(Ht, Chain),
+                FPMD = case blockchain_ledger_v1:fingerprint(Ledger) of
```
Vagabond: put this in an e2qc cache?

PR #301, opened 2019-12-01
use snapshots for ledger_at, when available

Vagabond:
Since we already create ledger snapshots on block absorb, cache them in
    an ETS table so that if we need to compute a ledger at that height
    later, we have it cached. Snapshots older than the lagging ledger will
    be deleted from the cache and be garbage collected as normal. This
    provides a very sigificant speed increase when syncing blocks or
    when the consensus group is validating proposed transactions.

    In addition remove a log message from poc receipt validation that was
    accidentally merged.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -13,7 +13,7 @@
     check_key/2, mark_key/2,
 
     new_context/1, delete_context/1, reset_context/1, commit_context/1, context_cache/1,
-    new_snapshot/1, release_snapshot/1, snapshot/1,
+    new_snapshot/1, has_snapshot/2, release_snapshot/1, snapshot/1,
 
     current_height/1, increment_height/2,
     transaction_fee/1, update_transaction_fee/1,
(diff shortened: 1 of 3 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -283,16 +285,31 @@ context_cache(#ledger_v1{mode=delayed, delayed=#sub_ledger_v1{context=Context, c
...
                         snapshot=undefined,
+                        snapshots=Cache,
                         active=#sub_ledger_v1{context=undefined, cache=undefined},
                         delayed=#sub_ledger_v1{context=undefined, cache=undefined}}=Ledger) ->
     case rocksdb:snapshot(DB) of
         {ok, SnapshotHandle} ->
+            {ok, Height} = current_height(Ledger),
+            DelayedLedger = blockchain_ledger_v1:mode(delayed, Ledger),
+            {ok, DelayedHeight} = current_height(DelayedLedger),
+            ets:delete(Cache, DelayedHeight),
```
vihu: Would it be wise to do `true = ets:delete(Cache, DelayedHeight)` here? Same for the insert next line?

Vagabond: Why? maybe we don't have a snapshot at that height

PR #305, opened 2019-12-03
Support for separate absorbed predicate function

andymck:
NOTE:  this is dependant upon proto PR: https://github.com/helium/proto/pull/19

Adds support for a new txn specific absorbed/2 function via which we can identify if a txn has already been seen before ( ie a replay ) and preventing a double spend.  It is called as part of and before is_valid and thus if we have seen a txn previously we will avoid having to run the more computationally heavy validations.

absorbed/2 functions have been added to those txns for which I have been able to identify some means to detect potential replays/double spends ( it is included for the heavy POC receipt txns ).

Where a txn specifies a nonce, I have used the approach of checking that.  Where a nonce is not present I have looked at other approaches based on the data available and the actions performed by absorb/2 or is_valid/2.  

If a txn module exports absorbed/2 then it will be called from blockchain_txn:is_valid/2.

Some minor modification have been made to a number of is_valid/2 implementations.  Mostly these are to remove checks/validations which are now in absorbed/2 but more importantly to allow is_valid/2 to be deterministic and to always return the same result for the same txn irrespective of whether the txn has been absorbed or not.  

Further Considerations 
--------------------
This PR is predominantly concerned with being able to determine if a txn is a replay.  It does not attempt to rewrite the is_valid/2 functions.  There is scope for improving the existing is_valid/2 implementations and breaking them out into well-formedness checks and rules based checks ( ie is the member field populated vs is the election too early or too late ).  As this PR has taken so long as it is and touches a fair amount of code and has sufficient value in itself, i think its best to tackle the wider is_valid/2 improvements in a follow up task/PR.  

Txns which have absorbed/2 support
--------------------
- blockchain_txn_add_gateway_v1
- blockchain_txn_assert_location_v1
- blockchain_txn_create_htlc_v1
- blockchain_txn_gen_gateway_v1
- blockchain_txn_oui_v1
- blockchain_txn_payment_v1
- blockchain_txn_poc_receipts_v1
- blockchain_txn_routing_v1
- blockchain_txn_security_exchange_v1
- blockchain_txn_token_burn_v1
- blockchain_txn_vars_v1

txns which do not have absorbed/2 support as no clear means to facilitate 
--------------------
- blockchain_txn_poc_request_v1
- txn_coinbase 
- txn_dc_coinbase ( only passes is_valid/2 if height at genesis block )
- txn_redeem_htlc
- txn_security_coinbase ( only passes is_valid/2 if height at genesis block )
- txn_token_burn_exchange_rate

txns which do not have absorbed/2 support and likely wont ever need it
--------------------
- blockchain_txn_bundle_v1
- blockchain_txn_reward_v1 ( doesnt have an absorb function, why ? )

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -22,7 +22,7 @@
     spend/3, spend/4,
     payment_txn/5, payment_txn/6,
     submit_txn/1, submit_txn/2,
-    create_htlc_txn/6,
+    create_htlc_txn/7,
     redeem_htlc_txn/3,
     peer_height/3,
     notify/1,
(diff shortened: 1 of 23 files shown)


evanmcc:
I think that this is irretrievably bitrotted at this point, and we'll go forwards with #1236

PR #307, opened 2019-12-04
initial version of poc gc

evanmcc:
in order to have ledger fingerprints, we need to clean up old poc stuff that has never been written over.  This is a first, kind of ugly, attempt at a cleanup.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -651,7 +651,14 @@ add_block_(Block, Blockchain, Syncing) ->
                                                             lager:error("Error absorbing transaction, Ignoring Hash: ~p, Reason: ~p", [blockchain_block:hash_block(Block), Reason]),
                                                             Error;
                                                         ok ->
-                                                            case blockchain_ledger_v1:new_snapshot(blockchain:ledger(Blockchain)) of
+                                                            try
+                                                                ok = blockchain_ledger_v1:maybe_gc_pocs(Blockchain)
+                                                            catch C:E ->
+                                                                    lager:info("poc gc failed with ~p:~p", [C,E]),
+                                                                    ok
+                                                            end,
+                                                            Ledger = blockchain:ledger(Blockchain),
+                                                            case blockchain_ledger_v1:new_snapshot(Ledger) of
                                                                 {error, Reason}=Error ->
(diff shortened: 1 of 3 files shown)


src/blockchain.erl
```diff
@@ -552,7 +552,9 @@ add_block_(Block, Blockchain, Syncing) ->
                                                             lager:error("Error absorbing transaction, Ignoring Hash: ~p, Reason: ~p", [blockchain_block:hash_block(Block), Reason]),
                                                             Error;
                                                         ok ->
-                                                            case blockchain_ledger_v1:new_snapshot(blockchain:ledger(Blockchain)) of
+                                                            Ledger = blockchain:ledger(Blockchain),
+                                                            ok = blockchain_ledger_v1:maybe_gc_pocs(Ledger, Blockchain),
```
Vagabond: Wny not just pass the chain here and extract the ledger in the function?

evanmcc: it was originally written to just need the ledger, but then it needed the chain for some reason and I didn't clean it up very well.  will fix.

Vagabond:
Can the draft status be removed?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1150,6 +1151,65 @@ delete_pocs(OnionKeyHash, Ledger) ->
...
+                                fun(PoC) ->
+                                        H = blockchain_ledger_poc_v2:block_hash(PoC),
+                                        case H of
+                                            <<>> ->
+                                                %% pre-upgrade pocs are ancient
+                                                false;
+                                            _ ->
+                                                {ok, B} = blockchain:get_block(H, Chain),
+                                                BH = blockchain_block:height(B),
+                                                (Height - BH) < 100
```
vihu: Maybe the `100` should be configurable via some chain var? `poc_challenge_interval` perhaps?

src/blockchain_utils.erl
```diff
@@ -26,6 +26,8 @@
 -include_lib("eunit/include/eunit.hrl").
 -endif.
 
+-include("blockchain_vars.hrl").
+
 -define(CHALLENGE_INTERVAL, poc_challenge_interval).
```
vihu: Maybe rip this out if it's not getting used anymore?

PR #309, opened 2019-12-04
Normalize floats and add more witness checks

vihu:
- Normalize all floats in path v2 and target v2

These  filters would kick in when we set `poc_version` to 5:
- Add witness inclusion filter
- Add bad witness rssi filter


vihu:
I'm closing this and will open two separate PRs, one for just normalization and another one for witness filteration.

PR #310, opened 2019-12-05
Normalize poc-v4 floats

vihu:
Hide the normalization behind poc-version > 4.

### DIFF
--- a/src/blockchain_poc_target_v2.erl
+++ b/src/blockchain_poc_target_v2.erl
@@ -11,6 +11,8 @@
 %%%-----------------------------------------------------------------------------
 -module(blockchain_poc_target_v2).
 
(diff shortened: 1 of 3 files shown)


src/blockchain_poc_target_v2.erl
```diff
@@ -23,7 +25,7 @@
 -define(POC_V4_TARGET_EXCLUSION_CELLS, 6000).
 
 -export([
-         target/3, filter/5
+         target/3, filter/5, challenge_age/1
```
Vagabond: Why was this exported?

vihu: Good catch, seems to have been a left-over from the previous conflated WIP branch.

PR #315, opened 2019-12-08
Attempt to use the snapshots more efficently

Vagabond:
* Check for a snapshot first in ledger_at
* Allow gateway scores to use cache if the ledger is snapshotted

### DIFF
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -163,16 +163,16 @@ hex_adjustment(Loc) ->
 
 score_gateways(Ledger) ->
     {ok, Height} = blockchain_ledger_v1:current_height(Ledger),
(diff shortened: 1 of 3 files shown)


src/blockchain_utils.erl
```diff
@@ -163,16 +163,16 @@ hex_adjustment(Loc) ->
 
 score_gateways(Ledger) ->
     {ok, Height} = blockchain_ledger_v1:current_height(Ledger),
-    case blockchain_ledger_v1:mode(Ledger) of
-        delayed ->
+    case blockchain_ledger_v1:snapshot(Ledger) of
+        {error, undefined}  ->
```
evanmcc: won't this also be true when we're using the stopgap overlays?

PR #317, opened 2019-12-09
Small snapshot improvements

Vagabond:
* Ensure snapshots can only be taken from an active ledger
* Delete the right old record from the cache
* Context cache updates should not override existing cache entries
* When restoring a context cache make sure the ledger is in delayed mode
  to prevent the context being installed onto the wrong subledger

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -304,6 +304,7 @@ context_cache(#ledger_v1{mode=delayed, delayed=#sub_ledger_v1{context=Context, c
 new_snapshot(#ledger_v1{db=DB,
                         snapshot=undefined,
                         snapshots=Cache,
+                        mode=active,
                         active=#sub_ledger_v1{context=undefined, cache=undefined},
                         delayed=#sub_ledger_v1{context=undefined, cache=undefined}}=Ledger) ->
     case rocksdb:snapshot(DB) of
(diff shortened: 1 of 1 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -304,14 +304,15 @@ context_cache(#ledger_v1{mode=delayed, delayed=#sub_ledger_v1{context=Context, c
 new_snapshot(#ledger_v1{db=DB,
                         snapshot=undefined,
                         snapshots=Cache,
+                        mode=active,
```
evanmcc: this will function clause.  should we add another clause to the function that throws a sensible error?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -327,7 +328,7 @@ context_snapshot(Context, #ledger_v1{db=DB, snapshots=Cache} = Ledger) ->
         _ ->
             case rocksdb:snapshot(DB) of
                 {ok, SnapshotHandle} ->
-                    ets:insert(Cache, {Height, {context, SnapshotHandle, Context}});
+                    ets:insert_new(Cache, {Height, {context, SnapshotHandle, Context}});
```
evanmcc: nice

src/ledger/v1/blockchain_ledger_v1.erl
Vagabond: Addressed

PR #318, opened 2019-12-10
Hierarchical POC targeting

vihu:


src/blockchain_utils.erl
```diff
@@ -204,6 +207,52 @@ vars_binary_keys_to_atoms(Vars) ->
...
+                               zones_(Ledger, Vars)
+                       end);
+        active ->
+            %% recalculate in active ledger mode
+            zones_(Ledger, Vars)
+    end.
+
+-spec zones_(Ledger :: blockchain_ledger_v1:ledger(),
+             Vars :: map()) -> zone_map().
+zones_(Ledger, Vars) ->
```
evanmcc: So, this isn't actually any cheaper.

What we should be doing, IMO, is to add an upgrade step that constructs the zones from the ledger on startup.  Once the zone structure is in the ledger (we'll need appropriate code to update it as machines are moved or new ones are asserted), then we should be able to just pull some ledger values and a few spots, making zones pretty cheap.  We can also cache zones at a height, but I suspect that it won't be a huge help.

Let me know if this isn't clear enough, and I can clarify, or alternately I can just add some commits to do what I said.

vihu: I see, right, probably gonna need some advice on the ledger upgrade code. We can discuss it offline. This was just a speculative thing anyway.

src/blockchain_poc_path_v3.erl
```diff
@@ -0,0 +1,397 @@
...
+    maps:get(poc_v4_randomness_wt, Vars).
+
+-spec poc_version(Vars :: map()) -> pos_integer().
+poc_version(Vars) ->
+    maps:get(poc_version, Vars).
+
+%% we assume that everything that has made it into build has already
+%% been asserted, and thus the lookup will never fail. This function
+%% in no way exists simply because
+%% blockchain_ledger_v1:find_gateway_info is too much to type a bunch
```
vihu: Hah! :smile:

src/blockchain_poc_target_v2.erl
```diff
@@ -55,6 +58,40 @@ target(Hash, GatewayScoreMap, Vars) ->
     {RandVal, _} = rand:uniform_s(Entropy),
     select_target(ScaledProbs, RandVal, Vars).
 
+-spec target_v2(Hash :: binary(),
```
vihu: Minor nitpick, but can we add some comments in/around this function please?

src/blockchain_poc_target_v2.erl
```diff
@@ -195,6 +245,16 @@ select_target([{GwAddr, Prob}=_Head | _], Rnd, _Vars) when Rnd - Prob =< 0 ->
...
 
+-spec select_hex(TaggedProbs :: [{h3:index(), float()}],
+                 Rnd :: float()) ->
+                        {error, no_hex} | {ok, h3:index()}.
+select_hex([], _)  ->
+    {error, no_hex};
+select_hex([{Idx, Prob}=_Head | _], Rnd) when Rnd - Prob =< 0 ->
+    {ok, Idx};
+select_hex([{_Idx, Prob} | Tail], Rnd) ->
+    select_hex(Tail, (Rnd - Prob)).
```
vihu: `(Rnd-Prob)` should probably be normalized, maybe we should just make a `select/2 or select/3` helper function..

src/blockchain_utils.erl
Vagabond: Given that the zone map is in the ledger now, is this function still needed?

src/transactions/v1/blockchain_txn_assert_location_v1.erl
```diff
@@ -363,12 +363,25 @@ absorb(Txn, Chain) ->
...
     end,
 
-    Gateways = blockchain_ledger_v1:active_gateways(Ledger),
-    Neighbors = blockchain_poc_path:neighbors(Gateway, Gateways, Ledger),
-    {ok, Gw} = blockchain_ledger_v1:find_gateway_info(Gateway, Ledger),
-    ok = blockchain_ledger_v1:fixup_neighbors(Gateway, Gateways, Neighbors, Ledger),
-    Gw1 = blockchain_ledger_gateway_v2:neighbors(Neighbors, Gw),
-    ok = blockchain_ledger_v1:update_gateway(Gw1, Gateway, Ledger).
+    case blockchain:config(?poc_version, Ledger) of
+        {ok, V} when V >= 6 ->
```
Vagabond: This var is already live, so we can't use that here?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -217,7 +217,19 @@ is_valid(Txn, Chain) ->
                                                                     {ok, OldLedger} = blockchain:ledger_at(blockchain_block:height(Block1), Chain),
                                                                     maybe_log_duration(ledger_at, StartLA),
                                                                     Path = case blockchain:config(?poc_version, OldLedger) of
-                                                                               {ok, V} when V > 3 ->
+                                                                               {ok, V} when V >= 6 ->
```
Vagabond: PoC v6 is already live

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1972,6 +1977,42 @@ maybe_use_snapshot(#ledger_v1{snapshot=Snapshot}, Options) ->
...
+        {ok, BinList} ->
+            {ok, binary_to_term(BinList)};
+        not_found ->
+            {error, not_found};
+        Error ->
+            Error
+    end.
+
+hex_name(Hex) ->
+    <<?hex_prefix, (integer_to_binary(Hex))/binary>>.
```
Vagabond: Why not do a big endian 64 bit integer in its own column family?

src/transactions/v1/blockchain_txn_assert_location_v1.erl
vihu: Oh yeah, good catch. We should just make a separate chain var `poc_target_zones` or something as a bool.

src/blockchain_poc_target_v2.erl
evanmcc: done

src/blockchain_poc_target_v2.erl
evanmcc: making a helper fn in utils is a great idea.  will normalize this

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: I'm not sure that there's any gain here to be had that would be worth the trouble here of adding another column family.

src/blockchain_poc_path_v3.erl
```diff
@@ -0,0 +1,397 @@
+%%%-----------------------------------------------------------------------------
```
Vagabond: How does this module differ from path_v2?

src/blockchain_poc_target_v2.erl
```diff
@@ -36,6 +58,56 @@ target(Hash, GatewayScoreMap, Vars) ->
...
+-spec target_v2(Hash :: binary(),
+                Ledger :: blockchain:ledger(),
+                Vars :: map()) -> {ok, libp2p_crypto:pubkey_bin()} | {error, no_target}.
+target_v2(Hash, Ledger, Vars) ->
+    %% Grab the list of parent hexes
+    {ok, Hexes} = blockchain_ledger_v1:get_hexes(Ledger),
+    HexList = maps:to_list(Hexes),
+
+    %% assemble the probabilities per hex for the CDF (each spot is
+    %% equally likely to get targeted at this stage)
```
Vagabond: Aren't denser hexes more likely to be targeted because of `Ct/GwCt`?

evanmcc: yes.  there's no curve here.  "each *spot* is equally likely" not hex

src/blockchain_poc_path_v3.erl
evanmcc: it... doesn't?  looks like I added it and then never needed to change it.

include/blockchain_vars.hrl
```diff
@@ -221,4 +221,8 @@
 %%% POC V5 vars
 %% Dictates how much randomness we want in the target selection
 -define(poc_v5_target_prob_randomness_wt, poc_v5_target_prob_randomness_wt).
+
+%% Target Zone Vars
+%% Create zones at this resolution for all the hotspots on the network.
+-define(poc_v5_target_zone_parent_res, poc_v5_target_zone_parent_res).
```
Vagabond: Reminder there was some discussion to change this name

src/blockchain_poc_target_v2.erl
Vagabond: oh yes, I misread

src/blockchain_poc_target_v2.erl
```diff
@@ -47,27 +119,34 @@ target(Hash, GatewayScoreMap, Vars) ->
...
-                                    true ->
-                                        %% Allow to be included in testing
-                                        true
-                                end
-                        end
+                        valid(Gateway, ChallengerLoc, Height, Vars)
                 end,
                 maps:without([ChallengerAddr], GatewayScoreMap)).
 
+valid(Gateway, Height, Vars) ->
```
Vagabond: This was used by pathv3 which looks like it got removed?

src/transactions/v1/blockchain_txn_assert_location_v1.erl
```diff
@@ -363,12 +363,27 @@ absorb(Txn, Chain) ->
...
+            {ok, Res} = blockchain:config(?poc_v5_target_zone_parent_res, Ledger),
+            Hex = h3:parent(Location, Res),
+
+            {ok, Hexes} = blockchain_ledger_v1:get_hexes(Ledger),
+            Hexes1 = maps:update_with(Hex, fun(X) -> X + 1 end, 1, Hexes),
+            ok = blockchain_ledger_v1:set_hexes(Hexes1, Ledger),
+
+            {ok, OldAddrs} = blockchain_ledger_v1:get_hex(Hex, Ledger),
+            ok = blockchain_ledger_v1:set_hex(Hex, [Gateway | OldAddrs], Ledger);
+        _ ->
```
Vagabond: Guessing this will need a rebase?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -217,7 +217,19 @@ is_valid(Txn, Chain) ->
...
+                                                                                   Vars = blockchain_utils:vars_binary_keys_to_atoms(blockchain_ledger_v1:all_vars(OldLedger)),
+                                                                                   StartFT = erlang:monotonic_time(millisecond),
+                                                                                   %% If we make it to this point, we are bound to have a target.
+                                                                                   {ok, Target} = blockchain_poc_target_v2:target_v2(Entropy, OldLedger, Vars),
+                                                                                   maybe_log_duration(target, StartFT),
+                                                                                   StartB = erlang:monotonic_time(millisecond),
+                                                                                   Time = blockchain_block:time(Block1),
+                                                                                   RetB = blockchain_poc_path_v3:build(Target, OldLedger, Time, Entropy, Vars),
```
Vagabond: Pathv3 is used here?

src/transactions/v1/blockchain_txn_assert_location_v1.erl
```diff
@@ -387,6 +400,31 @@ absorb(Txn, Chain) ->
             ok = blockchain_ledger_v1:update_gateway(Gw1, Gateway, Ledger)
     end.
 
+add_to_hex(Hex, Gateway, Ledger) ->
+    {ok, Hexes} = blockchain_ledger_v1:get_hexes(Ledger),
```
vihu: You could hoist this line in the caller and pass `Hexes` in to `remove_from_hex` and `add_to_hex` instead of doing it twice presumably?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -677,6 +677,8 @@ validate_var(?poc_typo_fixes, Value) ->
         false -> ok;
         _ -> throw({error, {invalid_poc_typo_fixes, Value}})
     end;
+validate_var(?poc_target_hex_parent_res, Value) ->
+    validate_int(Value, "poc_v4_parent_res", 3, 7, false);
```
vihu:
```suggestion
    validate_int(Value, "poc_target_hex_parent_res", 3, 7, false);
```

PR #320, opened 2019-12-11
Only fetch and score the gateways once

evanmcc:
currently, due to some code paths that were incompletely renovated during the poc_v4 update, we fetch the active_gateways column off the disk twice.  doing this twice adds roughly 30% to the receipt validation time.  it's fairly easy to make the code that uses the unscored gateways map use the scored version, so we can just skip doing this work.

### DIFF
--- a/src/blockchain_poc_target_v2.erl
+++ b/src/blockchain_poc_target_v2.erl
@@ -17,16 +17,14 @@
          target/3, filter/5
         ]).
 
--type gateway_score_map() :: #{libp2p_crypto:pubkey_bin() => {blockchain_ledger_gateway_v2:gateway(), float()}}.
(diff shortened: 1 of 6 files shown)


src/blockchain_poc_target_v2.erl
```diff
@@ -17,9 +17,11 @@
          target/3, filter/5
         ]).
 
--type gateway_score_map() :: #{libp2p_crypto:pubkey_bin() => {float(), blockchain_ledger_gateway_v2:gateway()}}.
```
Vagabond: Move this to to score_gateways or whatever and make sure we use this in its spec

vihu: Addressed here: ae6c4ff

PR #321, opened 2019-12-11
Implement print/1 in all transaction types

mrallen1:
Problems to solve:
* The eqc plugin should be deprecated in favor of upstream
* While there are some `print/1` functions in txn types, they are not fully implemented
* Find locations in the code where transactions are logged "naked" convert those calls to use the shiny new `print` function.

Solutions:
* Point eqc plugin to upstream
* Implement `print/1` for the rest of the transaction types.
* Update log directives to call the `print` function.

[story](https://app.clubhouse.io/hlm/story/5576/add-print-callback-to-all-transaction-types)

### DIFF
--- a/src/transactions/v1/blockchain_txn_coinbase_v1.erl
+++ b/src/transactions/v1/blockchain_txn_coinbase_v1.erl
@@ -17,7 +17,8 @@
     fee/1,
     is_valid/2,
     absorb/2,
-    sign/2
+    sign/2,
+    print/1
 ]).
 
 -ifdef(TEST).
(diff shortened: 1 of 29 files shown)


src/transactions/v1/blockchain_txn_coinbase_v1.erl
```diff
@@ -108,6 +109,19 @@ absorb(Txn, Chain) ->
...
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec print(undefined|txn_coinbase()) -> [iodata()].
+print(undefined) ->
+    <<"type=coinbase, undefined">>;
+print(#blockchain_txn_coinbase_v1_pb{} = Txn) ->
+    io_lib:format("txn_coinbase: payee: ~p, amount: ~p",
+                  [?MODULE:payee(Txn), ?MODULE:amount(Txn)]);
+print(Other) ->
```
vihu: Odd that this is done only here, I don't _think_ it's required tbh.

mrallen1: I think it's the first one I did. I will update it to be like the rest.

vihu: Sounds good :+1:

mrallen1:
@vihu suggested that `print/1` become a required callback in the ch story. I think that is a good idea and I'm going to make it so.

mrallen1:
build failure is on rocksdb blah

vihu:
@mrallen1 

Just gonna leave a comment here since its wayy too much to comment on every file/line.

So, every time we print an owner, payer, payee, signature (anything which is a binary), we should aim for printing its b58 form. There's a `libp2p_crypto:bin_to_b58/1` function to do so, could even make a macro if you wish.

Secondly, anytime we print a `gateway`, we should use its animal name version. It's far easier to digest it that way. `erl_angry_purple_tiger:animal_name(libp2p_crypto:bin_to_b58(Bin))` should give you an `{ok, AnimalName}`.

Let me know if these suggestions make sense?

mrallen1:
I'll finish this up tonight... still don't have a good grasp of what fields need to B58 encoded though, so I'll definitely need some review help on that front...

PR #324, opened 2019-12-17
Add blockchain:analyze/1 and blockchain:repair/1

Vagabond:
There are many things that go wrong when syncing is interrupted by some
kind of hard power down (OTAs or unplugging). This patch intends to try
to add some repair facilities for the common scenarios observed during
sync corruption.

The common sync failures are:

* Missing blocks
* Ledger inconsistency

To detect these failures several checks have been added:

* Check the delayed ledger is the right number of blocks behind
* Check the blockchain and the ledger agree on the blockchain height
* Check the delayed ledger advanced to the leading ledger fingerprints
  the same as the leading ledger

And several repair tools have been added:

* Advance the lagging ledger
* Re-sync a missing block (only gap one block wide is doable right now)
* Apply all the differences from an advanced lagging ledger to the
  leading ledger

To analyze the problems, use blockchain:analyze. Analyze will report a
single problem at a time.

To repair the current analyzed problem (if it is one we can repair), use
blockchain:repair. It will only repair a single problem at a time. It is
recommended you use analyze after each repair to check if there are more
problems, and to repair them if needed.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -24,6 +24,8 @@
 
     last_block_add_time/1,
 
+    analyze/1, repair/1,
+
     reset_ledger/1, reset_ledger/2, reset_ledger/3
 ]).
 
@@ -52,6 +54,7 @@
 -define(DB_FILE, "blockchain.db").
 -define(HEAD, <<"head">>).
 -define(TEMP_HEADS, <<"temp_heads">>).
+-define(MISSING_BLOCK, <<"missing_block">>).
 -define(GENESIS, <<"genesis">>).
 -define(ASSUMED_VALID, blockchain_core_assumed_valid_block_hash_and_height).
 -define(LAST_BLOCK_ADD_TIME, <<"last_block_add_time">>).
@@ -233,17 +236,23 @@ head_hash(#blockchain{db=DB, default=DefaultCF}) ->
 
 -spec sync_hash(blockchain()) -> {ok, blockchain_block:hash()} | {error, any()}.
 sync_hash(Blockchain=#blockchain{db=DB, default=DefaultCF}) ->
-    case persistent_term:get(?ASSUMED_VALID, undefined) of
-        undefined ->
-            ?MODULE:head_hash(Blockchain);
+    case rocksdb:get(DB, DefaultCF, ?MISSING_BLOCK, []) of
+        {ok, Hash} ->
+            %% this is the hash of the last block before the gap
+            {ok, Hash};
         _ ->
-            case rocksdb:get(DB, DefaultCF, ?TEMP_HEADS, []) of
-                {ok, BinHashes} ->
-                    Hashes = binary_to_term(BinHashes),
-                    RandomHash = lists:nth(rand:uniform(length(Hashes)), Hashes),
-                    {ok, RandomHash};
+            case persistent_term:get(?ASSUMED_VALID, undefined) of
+                undefined ->
+                    ?MODULE:head_hash(Blockchain);
                 _ ->
-                    ?MODULE:head_hash(Blockchain)
+                    case rocksdb:get(DB, DefaultCF, ?TEMP_HEADS, []) of
+                        {ok, BinHashes} ->
+                            Hashes = binary_to_term(BinHashes),
+                            RandomHash = lists:nth(rand:uniform(length(Hashes)), Hashes),
+                            {ok, RandomHash};
+                        _ ->
+                            ?MODULE:head_hash(Blockchain)
+                    end
             end
     end.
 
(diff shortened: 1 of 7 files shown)


Vagabond:
This PR also fixes a silly bug when we did not report we were doing an 'assume valid' sync and the sync height looked like it was going backwards.

src/blockchain.erl
```diff
@@ -465,14 +493,40 @@ add_block(Block, Blockchain) ->
...
+add_block(Block, #blockchain{db=DB, blocks=BlocksCF, heights=HeightsCF, default=DefaultCF} = Blockchain, Syncing) ->
     blockchain_lock:acquire(),
     try
-        case persistent_term:get(?ASSUMED_VALID, undefined) of
-            undefined ->
-                add_block_(Block, Blockchain, Syncing);
-            AssumedValidHashAndHeight ->
-                add_assumed_valid_block(AssumedValidHashAndHeight, Block, Blockchain, Syncing)
+        PrevHash = blockchain_block:prev_hash(Block),
+        case rocksdb:get(DB, DefaultCF, ?MISSING_BLOCK, []) of
```
evanmcc: could we consider walling this off behind a config var?

src/blockchain.erl
```diff
@@ -233,17 +236,23 @@ head_hash(#blockchain{db=DB, default=DefaultCF}) ->
 
 -spec sync_hash(blockchain()) -> {ok, blockchain_block:hash()} | {error, any()}.
 sync_hash(Blockchain=#blockchain{db=DB, default=DefaultCF}) ->
-    case persistent_term:get(?ASSUMED_VALID, undefined) of
-        undefined ->
-            ?MODULE:head_hash(Blockchain);
+    case rocksdb:get(DB, DefaultCF, ?MISSING_BLOCK, []) of
```
evanmcc: when we sync this and we get repaired, what happens with block + 1?

src/blockchain.erl
```diff
@@ -904,6 +958,218 @@ last_block_add_time(#blockchain{default=DefaultCF, db=DB}) ->
...
+                                                                                ({deleted, _, _}, Acc) ->
+                                                                                     maps:update_with(deleted, fun(V) -> V+1 end, 1, Acc)
+                                                                             end, #{}, Changes),
+                                                           {M, maps:to_list(Sum)}
+                                                   end, Mismatches)}};
+      E ->
+          E
+  end.
+
+%% TODO it'd be nice if the fingerprint function was consistent with the column family names
```
evanmcc: yeah.

src/transactions/blockchain_txn.erl
```diff
@@ -521,26 +526,40 @@ absorb_delayed(Block0, Chain0) ->
...
+                                              absorb_delayed_(Block1, Chain1);
+                                         (_, Acc) ->
+                                              Acc
+                                      end,
+                                      ok,
+                                      lists:seq(DelayedHeight+1, DelayedHeight + Lag)),
+                    case Res of
+                        ok ->
+                            ok = blockchain_ledger_v1:commit_context(DelayedLedger1);
+                        Error ->
```
evanmcc: do we need to clear the context here?

Vagabond: No, the error path discards the context

src/blockchain.erl
```diff
@@ -372,12 +389,23 @@ ledger_at(Height, Chain0) ->
 
 fold_chain(Chain0, DelayedHeight, DelayedLedger, Height) ->
     lists:foldl(
-      fun(H, ChainAcc) ->
-              {ok, Block} = ?MODULE:get_block(H, Chain0),
-              {ok, Chain1} = blockchain_txn:absorb_block(Block, ChainAcc),
-              Chain1
+      fun(H, {ok, ChainAcc}) ->
```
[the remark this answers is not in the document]
evanmcc: ignore this

src/blockchain.erl
Vagabond: It will go back to the previous HEAD now that the hole is filled in

src/blockchain.erl
Vagabond: What kind of config var, application env, command line flag?

src/blockchain.erl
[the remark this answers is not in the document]
Vagabond: Ironically I thought throwing would be more controversial

src/blockchain.erl
evanmcc: the sync will just error out when the block after the missing one is reapplied?

src/blockchain.erl
evanmcc: application env, so we can edit the config and restart

src/blockchain.erl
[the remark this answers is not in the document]
evanmcc: it's a nit, really.  throwing I think is better in places where you'd really prefer to return early.

src/blockchain.erl
Vagabond: yeah, that is the intention

PR #326, opened 2019-12-18
Fix incorrect witness bad rssi check

vihu:
- Hide behind PoC V6


src/blockchain_poc_path_v2.erl
```diff
@@ -326,10 +326,16 @@ check_witness_bad_rssi(Witness, Vars) ->
                         0 ->
                             %% No bad RSSIs found, include
                             true;
+                        BadCount when is_integer(V), V > 5 ->
```
evanmcc: in my heart, I would prefer that this be `V >= 6`, but that is very minor.

PR #329, opened 2019-12-19
Repair CLI interface

mrallen1:
Problem to solve: It would be far more convenient (and safer) to expose repair operations to a CLI interface than having to use a remote_console to invoke them.

Solution: Implement a CLI interface for certain repair operations. These are:

* pause_sync
* cancel_sync
* resume_sync
* sync_state
* analyze
* repair


mrallen1:
There are a couple of xref fails here... let me fix those before merging.

PR #331, opened 2019-12-24
Resync the chain if the ledger is missing or at height 0

Vagabond:


src/blockchain_sup.erl
```diff
@@ -104,8 +104,8 @@ init(Args) ->
         ?WORKER(blockchain_lock, []),
         ?WORKER(blockchain_swarm, [SwarmWorkerOpts]),
         ?WORKER(?EVT_MGR, blockchain_event, [BEventOpts]),
-        ?WORKER(blockchain_worker, [BWorkerOpts]),
```
madninja: just moving a line around?

Vagabond:
This actually didn't work because of the automatic ledger reset (that I just removed). Trying again.

src/blockchain_sup.erl
Vagabond: Process dependencies are hell

Vagabond:
This seems to have worked for me, please try it.

Vagabond:
Don't merge this as it breaks the miners consensus manager.

PR #336, opened 2020-01-08
Enable atomic_sync for blockchain and ledger

Vagabond:
Bump libp2p to get the relcast and rocksdb bump which gives us the new
atomic_sync option for rocksdb which prevents column families flushing
to disk separately, which we think is a source of ledger corruption.


fvasquez:
Do it.

PR #340, opened 2020-01-10
Switch witness garbage collection to on read rather than on assert

Vagabond:
Prior to this change, gateway witnesses were garbage collected when
their nonce changed as part of a location assertion. This involved a
full scan of all the ledger gateways, which is at least O(N).

This change attempts to drastically reduce the scope by checking the
nonces of witnesses against the actual gateway record's nonce when the
gateway record is loaded from rocksdb. Since gateways have a much
smaller witness list than the total number of gateways this bounds the
compute complexity much lower.

note that this is potentially consensus affecting if the results of the garbage collection are different than the original code. Additionally this PR is based on top of #339 which is much more conservative and likely much safer as an emergency fix.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_gateway_v2.erl
+++ b/src/ledger/v1/blockchain_ledger_gateway_v2.erl
@@ -27,7 +27,8 @@
     clear_witnesses/1,
     remove_witness/2,
     witnesses/1,
-    witness_hist/1, witness_recent_time/1, witness_first_time/1
+    witnesses/2,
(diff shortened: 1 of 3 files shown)


Vagabond:
This appears to be too slow because we have to check if we need to GC witnesses every read. We need a faster approach. One idea I have is to have a 'gateway epoch' that is incremented every time we have a block with an assert location in it. Each gateway record would have the epoch it was last updated in, and if the global gateway epoch was higher than the per-gateway epoch, we'd do the GC.

Vagabond:
Closing for now

PR #348, opened 2020-01-17
Update to POC V8

vihu:
This PR adds "centrality" checks for the RSSI histograms we have collected in the ledger.

We split the histogram into two "buckets" and assign probability depending on whether the witnessing appears within an acceptable range which is controlled by introducing three new chain vars:

- `poc_good_bucket_low` -> lower bound for RSSI
- `poc_good_bucket_high` -> upper bound for RSSI
- `poc_centrality_wt` -> weightage given to the centrality probability

I've also added another eqc for the same.


vihu:
I'm closing this in favor of https://github.com/helium/blockchain-core/pull/365

PR #349, opened 2020-01-17
Move macros to util header file

vihu:


mrallen1:
LGTM - sorry I reversed the order of the arguments. You could've just told me.

vihu:
> LGTM - sorry I reversed the order of the arguments. You could've just told me.

No worries, I just noticed it when I was running some tests and it would fail to boot. We should've caught it with the prior PR but such is life. I don't _think_ this bug would have a catastrophic impact so it's okay :)

PR #350, opened 2020-01-17
various changes for cleaner process termination and app shutdown

andymck:


src/handlers/blockchain_fastforward_handler.erl
```diff
@@ -45,9 +45,11 @@ server(Connection, Path, _TID, Args) ->
 %% libp2p_framed_stream Function Definitions
 %% ------------------------------------------------------------------
 init(client, _Conn, [Blockchain]) ->
+    erlang:process_flag(trap_exit, true),
```
vihu: Just for my info, why add the `trap_exit` here? Are we assuming that some upper level libp2p supervisor/server is gonna crash if we don't do this?

andymck: this handler is a libp2p_framed_stream behaviour and the terminate of that performs some cleanup such as closing libp2p_connections.  If the process is not trapping exits, then during a shutdown this terminate will not be called and thus those connections not closing cleanly can potentially result in unexpected errors.

So in general what I am doing is trapping exits on most supervised processes.  This is especially important to do on processes where cleanup/controlled shutdown is required but I am also trapping exits by default even on those top level processes which dont have cleanup requirements.  This enables any msg currently being processed to complete orderly.  My rationale here is that its better to trap exits until you know its not required or its going to cause issues ( such as would be the case if the process was likely to have a high volume mailbox or individual msg handling takes a long time ).

src/blockchain_lock.erl
```diff
@@ -45,6 +45,7 @@ start_link() ->
     gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
 
 init([]) ->
+    erlang:process_flag(trap_exit, true),
```
Vagabond: Why trap exits if there's no cleanup being done in terminate?

andymck: It mostly personal preference.  I have found that as systems develop it less error prone and less likely to lead to mis assumptions to get into the habit of trapping exits by default on top level supervised processes than not.  As per the cases in this PR its easy to start adding cleanup to terminate ( especially if the server is exporting it ) and assume it will be triggered when in most default flows it wont.

On the functional side it also has the benefit of permitting the server too orderly complete its current msg handling and thus much less likely to result in any errors during shutdown ( unless of course the mailbox count is likely to be high and/or msg handling takes a long time )

PR #352, opened 2020-01-22
Add support for var hooks

evanmcc:
this PR adds hooks that are invoked when a chain var goes into affect.  they're meant to give a mode-agnostic place for ledger upgrades/downgrades/format changes.

This is a WIP, as I'm not sure how well this will play with the delayed ledger.


vihu:
Closing this as it's very stale and the rebase is a pain to go over 1.5 years+ of commit history. Since the changes are relatively minimal here, I'll open a new PR and we can review that.

PR #353, opened 2020-01-22
Maybe start new sync handler client if hit disjoint_chain error when adding block

andymck:
Remove the starting up off a new blockchain_sync_handler stream in the event we hit disjoint_chain error when adding a block from gossip handler.  Instead call into blockchain worker and have it start a sync handler only if one is not already running.

End result is that we only ever have a single running client sync handler which is managed from blockchain worker.

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -31,6 +31,7 @@
 
     new_ledger/1,
 
+    maybe_sync/0,
     sync/0,
     cancel_sync/0,
     pause_sync/0,
@@ -106,6 +107,9 @@ cancel_sync() ->
 
 pause_sync() ->
     gen_server:call(?SERVER, pause_sync, infinity).
+
+maybe_sync() ->
+    gen_server:call(?SERVER, maybe_sync, infinity).
 
 sync_paused() ->
     try
@@ -344,6 +348,9 @@ handle_call({new_ledger, Dir}, _From, State) ->
     %% snapshot cache ETS table can be owned by an ephemeral process.
     Ledger1 = blockchain_ledger_v1:new(Dir),
     {reply, {ok, Ledger1}, State};
+
+handle_call(maybe_sync, _From, State) ->
+    {reply, ok, maybe_sync(State)};
 handle_call(sync, _From, State) ->
     %% if sync is paused, unpause it
     {reply, ok, maybe_sync(State#state{sync_paused = false})};
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -109,7 +109,7 @@ pause_sync() ->
     gen_server:call(?SERVER, pause_sync, infinity).
 
 maybe_sync() ->
-    gen_server:call(?SERVER, maybe_sync, infinity).
+    gen_server:cast(?SERVER, maybe_sync).
 
 sync_paused() ->
     try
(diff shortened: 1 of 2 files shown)


Vagabond:
What is the advantage to only having a single sync handler?

andymck:
@Vagabond this came out of a slack conversation.  I have posted that below for easy reference, I do not know exactly whether this is an advantage or not ( outside of cleaner shutdown ) but it does seem strange that we maintain a sync handler in blockchain_worker and then spawn one or more additional sync handlers from gossip handler but yet dont make any attempt to monitor them.

In terms of cleaner shutdown the fact that gossip handler is started unmonitored/unlinked sync_handlers results in an un orderly shutdown.  The gossip handler cannot stop the sync handlers during termination, unlike blockchain_worker which does stop the sync handler during shutdown and ensures there is only one active at any one time.  The end result of this is that during shutdown the gossip spawned sync handlers could have msgs in its mailbox and will continue to process these resulting in spurious errors ( for instance it may process a msg after rocksdb and or the ledger is closed )

If we *do* want gossip handler to start its own sync handler then we should update it to follow the convention used in blockchain worker and ensure we can orderly terminate them



slack conversation
-------------------
mack  4:54 PM
it is normal and expected for any one miner to potentially have multiple concurrent ‘blockchain_sync_handler’ processes running, right?

evan  4:54 PM
depends
4:55
but not really?
4:55
well
4:55
you should only have zero or one client, but you might be server to many (edited) 

mack  4:57 PM
as part of the miner shutdown task I could see multiple running when testing and carried that forward with my changes.  Today i started trying to put together a flow diagram ( initially to keep myself right ) and started to doubt whether this was expected
4:57
only one sync handler can be active via blockchain worker
4:57
but one can also be kicked off from the gossip handler
4:57
without impacting the running blockchain worker one
4:59
and I am seeing this in local playing of my miner cleaner shutdown branch
4:59
but now doubting whether this is a good thing

evan  4:59 PM
@mack client = syncing blocks in, server = syncing blocks out

mack  4:59 PM
yeah these are def both clients
5:02
so def on my branch two sync clients can be running at the same time…i am 99% sure same on master but will go back and confirm…maybe i missed something

evan  5:02 PM
I mean, it's a bit silly
5:02
one will fail to take the lock, wait for a long time, then fail when the lock is released (edited) 

mack  5:03 PM
k so if the expectation is that there is only one active sync client at any one time, then i def need some changes to my branch

evan  5:04 PM
I don't know that it's the expectation
5:04
hm.  can you figure out why the gossip handler would start one? (edited) 

mack  5:05 PM
it does so if blockchain:add_block returns disjoint_chain

vihu  5:06 PM
Right, because u can get a block gossiped to you while you’re behind and maybe syncing

mack  5:07 PM
so we have an active sync client running via blockchain_worker
5:08
and we get a gossiped block which results in the disjoint_chain error
5:08
and then end up with another sync client being spun up from the blockchain_gossip_handler
5:09
in fact gossip handler can in theory start multiple sync clients, one for each encounter with disjoint_chain error
5:09
it doesnt check if we already have an active sync

evan  5:11 PM
ok
5:11
maybe we should just stop doing that

vihu  5:11 PM
Yeah, I don’t remember why it was done this way at the time.

evan  5:11 PM
instead of attempting to "handle it properly"

mack  5:12 PM
so let blockchain_worker maintain the sync client? (edited) 

evan  5:12 PM
yes
5:13
just don't start it when you get disjoint_chain

mack  5:13 PM
yeah

evan  5:13 PM
can you do that in a separate PR?

mack  5:13 PM
sure

evan  5:13 PM
so we can talk it over without tying it to the rest of your work

PR #355, opened 2020-01-25
Compact the blockchain and ledger on bootup to free RAM

Vagabond:


fvasquez:
My understanding is that the compaction happens in the background so neither the blockchain or ledger databases are unavailable while compaction is occurring.  If we run out of memory during compaction does compaction continue to make progress until eventually we don't?

PR #360, opened 2020-01-31
Refresh witnesses for random subset of hotspots

vihu:
Attempt to refresh witnesses for a deterministically random subset of hotpots every X blocks.

`X` is configured via the `witness_refresh_interval` chain var.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -41,6 +41,7 @@
     fixup_neighbors/4,
     add_gateway_location/4,
     add_gateway_witnesses/3,
+    refresh_gateway_witnesses/2,
 
     gateway_versions/1,
 
@@ -149,6 +150,7 @@
 -type htlcs() :: #{libp2p_crypto:pubkey_bin() => blockchain_ledger_htlc_v1:htlc()}.
 -type securities() :: #{libp2p_crypto:pubkey_bin() => blockchain_ledger_security_entry_v1:entry()}.
 -type hexmap() :: #{h3:h3_index() => non_neg_integer()}.
+-type gateway_offsets() :: [{pos_integer(), libp2p_crypto:pubkey_bin()}].
 
 -export_type([ledger/0]).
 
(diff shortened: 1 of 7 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1136,6 +1138,53 @@ add_gateway_witnesses(GatewayAddress, WitnessInfo, Ledger) ->
...
+-spec remove_gateway_witness(GatewayPubkeyBin :: libp2p_crypto:pubkey_bin(),
+                             Ledger :: ledger()) -> ok | {error, any()}.
+remove_gateway_witness(GatewayPubkeyBin, Ledger) ->
+    case ?MODULE:find_gateway_info(GatewayPubkeyBin, Ledger) of
+        {error, _}=Error ->
+            Error;
+        {ok, GW0} ->
+            GW1 = blockchain_ledger_gateway_v2:clear_witnesses(GW0),
+            AGwsCF = active_gateways_cf(Ledger),
+            cache_put(Ledger, AGwsCF, GatewayPubkeyBin, blockchain_ledger_gateway_v2:serialize(GW1))
```
evanmcc: could replace this with `update_gateway/3`

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -561,6 +561,8 @@ validate_var(?election_restart_interval, Value) ->
 %% ledger vars
 validate_var(?var_gw_inactivity_threshold, Value) ->
     validate_int(Value, "var_gw_inactivity_threshold", 15, 2880, false);
+validate_var(?witness_refresh_interval, Value) ->
+    validate_int(Value, "witness_refresh_interval", 10, 20000, false);
```
evanmcc: 10 is probably way too low?  maybe approximately 1 day to two weeks, so something more like 1000 - 20000?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1306,6 +1355,40 @@ maybe_gc_pocs(Chain) ->
...
+                                 GatewayPubkeyBins)).
+
+-spec filtered_gateways_to_refresh(Hash :: blockchain_block:hash(),
+                                   RefreshInterval :: pos_integer(),
+                                   GatewayOffsets :: gateway_offsets()) -> gateway_offsets().
+filtered_gateways_to_refresh(Hash, RefreshInterval, GatewayOffsets) ->
+    RandState = blockchain_utils:rand_state(Hash),
+    %% NOTE: I believe this ensure that the random number gets seeded with a value
+    %% higher than the RefreshInterval
+    {RandVal, _NewRandState} = rand:uniform_s(2*RefreshInterval, RandState),
```
evanmcc: I'm concerned about the rand interval. maybe it could be 100? or 200? or bigger? the interval is never going to be huge, and I don't think that there's a reason to make it so tight.  I also think that it needs to be a chain var, or if we ever need to tune it we're going to need a var to switch between versions. might as well get it out of the way.

src/transactions/v1/blockchain_txn_vars_v1.erl
vihu: Right, yes. I put it as 10 for the testing to not be a pain heh.

src/ledger/v1/blockchain_ledger_v1.erl
vihu: Updated

src/ledger/v1/blockchain_ledger_v1.erl
vihu: Updated b4904ca

PR #363, opened 2020-01-31
Switch eqc plugin and fix hex_eqc

vihu:
We'll see if https://github.com/kellymclaughlin/rebar3-eqc-plugin/pull/33 pans out otherwise switching back to @Vagabond's fork.


mrallen1:
LGTM 👍

PR #365, opened 2020-02-05
POC v8

vihu:
This PR add support for POC v8. Most notable changes:

- The `target_v3:target` function now returns a `rand_state` which is then propagated through to the `path_v4:build` function making the `Hash` as a parameter in the `path_v4:build` function superfluous. `Hash` has been removed from the `path_v4:build` signature. This ensures that we are guaranteed to get a _new_ random value when selecting a potential next hop.

- Witness rssi centrality checks. We split the histogram into two "buckets" and assign probability depending on whether the witnessing appears within an acceptable range which is controlled by introducing three new chain vars:
-- `poc_good_bucket_low` -> lower bound for RSSI
-- `poc_good_bucket_high` -> upper bound for RSSI
-- `poc_centrality_wt` -> weight-age given to the centrality probability

- Limit distance of potential next hops. This is controlled by the chain var `poc_max_hop_cells`. This is to ensure we don't do paths which are extremely improbable to successfully complete.

- Scope poc related code into `src/poc` dir.


src/poc/blockchain_poc_target_v3.erl
```diff
@@ -0,0 +1,256 @@
...
+    GatewayParent = h3:parent(GatewayLoc, ParentRes),
+    %% Check that they are far
+    try h3:grid_distance(ChallengerParent, GatewayParent) > ExclusionCells of
+        Res -> Res
+    catch
+        %% Grid distance may badarg because of pentagonal distortion or
+        %% non matching resolutions or just being too far.
+        %% In either of those cases, we assume that the gateway
+        %% is potentially legitimate to be a target.
+        _:_ -> true
```
Vagabond: Is this the correct fallback decision here? Maybe false would be better choice?

Vagabond: Or giving vincenty a shot

src/poc/blockchain_poc_path_v4.erl
```diff
@@ -309,12 +306,32 @@ filter_witnesses(GatewayLoc, Indices, Witnesses, Ledger, Vars) ->
...
+check_witness_too_far(WitnessLoc, GatewayLoc, Vars) ->
+    POCMaxHopCells = poc_max_hop_cells(Vars),
+    try h3:grid_distance(WitnessLoc, GatewayLoc) of
+        Res ->
+            Res < POCMaxHopCells
+    catch
+        %% Grid distance may badarg because of pentagonal distortion or
+        %% non matching resolutions or just being too far.
+        %% In either of those cases, we assume that the gateway
+        %% is potentially legitimate to be a target.
```
Vagabond:
```suggestion
        %% is potentially illegitimate to be a target.
```

vihu:
Will remove the `io:format` after some eqc runs...

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -218,6 +218,18 @@ is_valid(Txn, Chain) ->
                                                                     {ok, OldLedger} = blockchain:ledger_at(blockchain_block:height(Block1), Chain),
                                                                     maybe_log_duration(ledger_at, StartLA),
                                                                     Path = case blockchain:config(?poc_version, OldLedger) of
+                                                                               {ok, V} when V >= 8 ->
+                                                                                   Vars = blockchain_utils:vars_binary_keys_to_atoms(blockchain_ledger_v1:all_vars(OldLedger)),
+                                                                                   StartFT = erlang:monotonic_time(millisecond),
+                                                                                   %% If we make it to this point, we are bound to have a target.
+                                                                                   {ok, {Target, TargetRandState}} = blockchain_poc_target_v2:target_v2(Entropy, OldLedger, Vars),
```
Vagabond: You'll need to update other uses of target_v2 to ignore the returned rand state.

src/poc/blockchain_poc_path_v4.erl
```diff
@@ -0,0 +1,567 @@
...
+    %% Try to find a next hop
+    case next_hop(TargetPubkeyBin, Ledger, HeadBlockTime, Vars, RandState, Indices) of
+        {error, _} ->
+            lists:reverse(Path);
+        {ok, WitnessPubkeyBin} ->
+            %% Try the next hop in the new path, continue building forward
+            NextHopGw = find(WitnessPubkeyBin, Ledger),
+            Index = blockchain_ledger_gateway_v2:location(NextHopGw),
+            NewPath = [WitnessPubkeyBin | Path],
+            {_, NewRandState} = rand:uniform_s(RandState),
```
Vagabond: Why?

vihu: We've gotten a next hop at this point, continue building the path with a new rand_state?

src/poc/blockchain_poc_path_v4.erl
```diff
@@ -0,0 +1,567 @@
...
+-spec find(libp2p_crypto:pubkey_bin(), blockchain_ledger_v1:ledger()) -> blockchain_ledger_gateway_v2:gateway().
+find(Addr, Ledger) ->
+    {ok, Gw} = blockchain_ledger_v1:find_gateway_info(Addr, Ledger),
+    Gw.
+
+-spec split_hist(Hist :: blockchain_ledger_gateway_v2:histogram(),
+                 Vars :: map()) -> {blockchain_ledger_gateway_v2:histogram(),
+                                    blockchain_ledger_gateway_v2:histogram()}.
+split_hist(Hist, Vars) ->
+    %% TODO: Maybe these can just be constants instead of vars?
```
Vagabond:
```suggestion
```

src/poc/blockchain_poc_path_v4.erl
```diff
@@ -0,0 +1,567 @@
...
+                        BadCount when is_integer(V), V > 5 ->
+                            %% Activate with PoC v6
+                            %% Check that the bad rssi count is less than
+                            %% the sum of other known good rssi
+                            BadCount < lists:sum(maps:values(maps:without([28], Hist)));
+                        BadCount ->
+                            %% If the bad RSSI count does not dominate
+                            %% the overall RSSIs this witness has, include,
+                            %% otherwise exclude
+                            %% XXX: This is an incorrect check
```
Vagabond: This is a new module, remove this

eqc/hex_eqc.erl
```diff
@@ -28,13 +28,13 @@ prop_hex_check() ->
                 %% Intiial acc for the counter, each node starts with a 0 count
                 InitAcc = maps:map(fun(_, _) -> 0 end, CumulativePopulationMap),
 
-                Fname = "/tmp/zones_" ++ libp2p_crypto:bin_to_b58(Hash),
+                %% Fname = "/tmp/zones_" ++ libp2p_crypto:bin_to_b58(Hash),
```
evanmcc: can we put these behind env vars or something?

eqc/zone_target_eqc.erl
```diff
@@ -62,11 +62,11 @@ gen_hash() ->
     binary(32).
 
 gen_challenger_index() ->
-    ?SUCHTHAT(S, int(), S < 2614 andalso S > 0).
```
evanmcc: can we pull this stuff into a utility module that handles everything with regards to the pickled ledger?  This is getting a bit repetitive/noisy.

src/poc/blockchain_poc_target_v2.erl
```diff
@@ -82,8 +84,17 @@ target_v2(Hash, Ledger, Vars) ->
     ProbTargetMap = target_prob(ProbScores, ProbEdges, Vars),
     %% Sort the scaled probabilities in default order by gateway pubkey_bin
     %% make sure that we carry the entropy through for determinism
-    {RandVal, _} = rand:uniform_s(Entropy1),
-    blockchain_utils:icdf_select(lists:keysort(1, maps:to_list(ProbTargetMap)), RandVal).
+    {RandVal, RandState} = rand:uniform_s(Entropy1),
+    io:format("randval: ~p, select_target~n", [RandVal]),
```
evanmcc: remove debugging

src/poc/blockchain_poc_path_v4.erl
evanmcc: why not just seed instead of iterating it?

eqc/zone_target_eqc.erl
vihu: I have one, will push in a bit :+1:

src/poc/blockchain_poc_path_v4.erl
```diff
@@ -0,0 +1,565 @@
...
+                    %% P(WitnessCount) = Probability that the witness is infrequent.
+                    PWitnessCount = witness_count_probs(FilteredWitnesses, Vars),
+                    %% P(RSSICentrality) = Probability that the witness rssi lies within a good range
+                    PWitnessRSSICentrality = witness_rssi_centrality_probs(FilteredWitnesses, Vars),
+                    %% P(Witness) = RSSIWeight*P(WitnessRSSI) + TimeWeight*P(WitnessTime) + CountWeight*P(WitnessCount)
+                    PWitness = witness_prob(Vars, PWitnessRSSI, PWitnessTime, PWitnessCount, PWitnessRSSICentrality),
+                    PWitnessList = lists:keysort(1, maps:to_list(PWitness)),
+                    %% Select witness using icdf
+                    {RandVal, _} = rand:uniform_s(RandState),
+                    io:format("randval: ~p, next_hop~n", [RandVal]),
```
evanmcc: debugging

src/poc/blockchain_poc_path_v4.erl
```diff
@@ -0,0 +1,565 @@
...
+witness_rssi_centrality_probs(Witnesses, Vars) ->
+    maps:map(fun(_WitnessPubkeyBin, Witness) ->
+                     try
+                         blockchain_ledger_gateway_v2:witness_hist(Witness)
+                     of
+                         Hist ->
+                             %% The closer these values are to 0.0, the more confident we
+                             %% are that this witness has a reasonable looking RSSI, therefore
+                             %% we bias _for_ picking that witness
+                             {MaxMetric, MeanMetric} = centrality_metrics(Hist, Vars),
```
evanmcc: it looks like bad witnesses are already filtered out, we're biasing against marginal ones towards ones that are clearly good?

vihu: Yeah, I was thinking about that as well. Given that we already have filtered out what we consider not to be in the good range, does it make much sense to even introduce a weight for it? I can change it to be just random and remove the `centality_wt` entirely.

evanmcc: Yeah, my concern here would be that marginal (but real) stuff wouldn't get explored in a reasonable amount of time, or borderline paths wouldn't get removed quickly enough by proving their badness.  but "quickly enough" is very vaguely defined here.

eqc/hex_eqc.erl
vihu: I can do that in a separate PR

src/poc/blockchain_poc_path_v4.erl
Vagabond: This ^

src/poc/blockchain_poc_path_v4.erl
```diff
@@ -0,0 +1,564 @@
...
+%%% Once we have a target we recursively find a potential next hop from the target
+%%% gateway by looking into its witness list.
+%%%
+%%% Before we calculate the probability associated with each witness in witness
+%%% list, we filter out potentially useless paths, depending on the following filters:
+%%% - Next hop witness must not be in the same hex index as the target
+%%% - Every hop in the path must be unique
+%%% - Every hop in the path must have a minimum exclusion distance
+%%%
+%%% The criteria for a potential next hop witness are biased like so:
```
Vagabond: These comments could use correcting

src/poc/blockchain_poc_path_v4.erl
vihu: I thought that we should seed _only_ once using the hash, which is done [here](https://github.com/helium/blockchain-core/pull/365/files#diff-7ba66bc9e6dded1f783877554d6e0e05R104). I'm not certain if iterating over a previously generated seed to get a new random state with `rand:uniform_s` is okay or reseeding from a previous random state is better?

```
{_, S2} = rand:uniform_s(..., S1)
S3 = rand:seed_s(S1)
S2 == S3.
false
```

vihu: Check bdadc9f ?

src/poc/blockchain_poc_path_v4.erl
vihu: Ok, since most of the simulations have looked reasonable so far, I'm leaning towards keeping the centrality_wt. If it doesn't work as intended (we can observe for a couple days), we can zero it out and do a `1.0` random_wt with a subsequent chain var txn.

src/poc/blockchain_poc_target_v3.erl
```diff
@@ -0,0 +1,107 @@
...
+                                        end,
+                                        #{},
+                                        FilteredList),
+            %% Sort the scaled probabilities in default order by gateway pubkey_bin
+            %% make sure that we carry the rand_state through for determinism
+            {RandVal, TargetRandState} = rand:uniform_s(HexRandState),
+            {ok, TargetPubkeybin} = blockchain_utils:icdf_select(lists:keysort(1, maps:to_list(ProbTargetMap)), RandVal),
+            {ok, {TargetPubkeybin, TargetRandState}};
+        _ ->
+            {error, no_target}
```
Vagabond: Why not call target again with this hex excluded?

PR #370, opened 2020-02-11
rework txn mgr

andymck:
There are a couple of diagrams which help to outline the main differences between the existing txn mgr and this new proposed txn mgr:

existing: 
https://docs.google.com/drawings/d/1Hl4hwjzGZZmlThxCDxTtk5lkaJLa79RcfcSiFIp-sRs/edit?usp=sharing

proposed: 
https://docs.google.com/drawings/d/1OReoD_uMtQq-S8oKm_66uNaBxfPZytS2Zb8CF_B0vps/edit?usp=sharing

**This PR addresses the following issues with the current txn mgr:**

1.  The rejected flow is basically an infinite loop within the context of a block epoch.  Unless the txn type is an assert location, the txn will keep being resent to another random CG member ad infinitum, it will remain in the cache and this loop until either it is accepted or marked invalid as part of validations during the add_block event or it is included in the new block.  Effectively the currently implemented 2F+1 rejection check is redundant and we are in a way spamming the CG,  if a txn cannot be accepted ( for example it has a bad nonce ) we will just keep bouncing it from one CG member to the next ad infinitum and forcing the CG member to run expensive validations each time.

2.  Txns are only sent to a single CG member at any one time

3.  A submitted txn is required to be accepted by only a single CG member.  Once we get that first acception it will not be sent to any other members until the next add_block event at which point it will be resent if not included in the block or not marked invalid.

4.  The resubmit msg handling has a bug which results in it inserting a badly formed tuple to the map.

5.  As part of the add_block event we are, unnecessarily, revalidating cached txns which were included in the newly added block

6.  Dialers are not well managed, with numerous cases of them being left open.

8.  The mgr  will never report a 'rejected' response for a txn, as txns can never actually be rejected.  The mgr can only report either 'ok' or 'invalid'

**Key points to note re the new txn mgr in this PR:**

1.  Utilizes ets for its cache rather than a map

2.  Upon receipt of a new txn received via submit_txn/2, the txn is immediately submitted to 2F+1 random members

3.  When inserting a txn into the cache we get the current block height and save this along with the txn data to the cache.  This will be used to determine as part of the rejection flow how long any one txn has existed in the cache.  If a txn keeps being rejected across multiple block epochs then we will give up on it and reject it.

4.  Upon receipt of an 'accepted' response from a CG member we append that member to the txn's accept list and stop its dialer.  The txn will remain in the cache

5.  Upon receipt of a 'rejected' response from a CG member we append that member to the txn's reject list and stop its dialer

- We then check if we have exceeded our rejection threshold of F+1.  

- If true we assume the txn is considered invalid for the current block/round but we dont automatically reject the txn at this point

- To allow for the fact that a txn may later be accepted we keep the rejected txn in the cache

- It will then be resubmitted to the CG but only after the next add_block event, essentially we are saying once its rejected it unlikely to be accepted in the same block epoch and so hold until the next add_block event and resend at that point

- If the txn continues to be rejected and does so across N blocks then we will give up on the txn, delete it from the cache and invoke its callback with rejected response
 
6.  Resubmits are now only performed as part of a new add_block event.  Within the context of a single block epoch a txn is sent to at most 2F+1 members, irrespective of how many rejections it receives

7.  Resubmits are now only performed after validations are run within the mgr and as such we will not be sending obviously invalid txns to the CG group

**Performance gains**

The rework has mostly tried to focus on correctness but there some gains to be had:

1.  Using ETS rather than maps will provide for some gain, mostly when we are pulling all txns from the cache

2.  The resend loop for rejected txns is removed.  Assuming we have 'rejected' txns on a regular basis it would have resulted in the mgr continually sending, handling rejection, resending.  Whilst validations were not run in this loop it still has some cost but I expect the greater gain will be on the CG with members no longer being continually sent a rejected txn and having to run expensive validations on this before rejecting it only for it to come right back again

3.  As part of handling the add_block event, the old mgr would revalidate all txns in its cache including those included in the new block.  Now we only run validations after removing the mined txns from the cache


src/transactions/blockchain_txn_mgr.erl
```diff
@@ -32,11 +33,15 @@
...
-          txn_map = #{} :: txn_map(),
+          submit_f :: undefined | integer(),
+          reject_f :: undefined | integer(),
+          cur_block_height :: undefined | integer(),
+          txn_cache :: undefined | ets:tid(),
           chain :: undefined | blockchain:blockchain()
          }).
 
--type txn_map() :: #{blockchain_txn:txn() => {fun(), [], undefined | pid()}}.
+-type cached_txn_type() :: {blockchain_txn:txn(), {fun(), undefined | integer(), [libp2p_crypto:pubkey_bin()], [libp2p_crypto:pubkey_bin()], undefined | [pid()]}}.
```
vihu: Can we add identifiers for this spec please? So like:

`{Txn :: blockchain_txn:txn(), {CallbackFun :: fun(), ... : undefined | integer()......

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +57,300 @@ submit(Txn, Callback) ->
...
-    gen_server:call(?MODULE, txn_map, infinity).
+-spec txn_list() -> [cached_txn_type()].
+txn_list() ->
+    gen_server:call(?MODULE, txn_list, infinity).
 
 %% ------------------------------------------------------------------
 %% gen_server Function Definitions
 %% ------------------------------------------------------------------
 init(_Args) ->
+    erlang:process_flag(trap_exit, true),
```
vihu: Why the trap exit here?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +57,300 @@ submit(Txn, Callback) ->
...
+    TxnCache = ets:new(?TXN_CACHE,
+                    [named_table,
+                     public,
```
vihu: Why is this ets public?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +57,300 @@ submit(Txn, Callback) ->
...
+
+handle_cast({set_chain, Chain}, State=#state{chain = undefined}) ->
+    NewState = initialize_with_chain(State, Chain),
+    {noreply, NewState};
+
+handle_cast({submit, Txn, Callback}, State=#state{chain = undefined}) ->
+    %% Got txn when there is no chain, keep it in the cache and process when its available
+    %% as no chain we dont have any height data, so cache it with height = undefined
+    %% we will update when the chain is set and we resubmit the cached txns
+    cache_txn(Txn, Callback, undefined, [], [],[]),
```
vihu: nitpicking, `ok = cache_txn...`

src/cli/blockchain_cli_txn.erl
```diff
@@ -63,25 +63,27 @@ txn_queue_usage() ->
...
+    lists:map(fun({Txn, {_Callback, AcceptedBlockHeight, Acceptions, Rejections, _Dialer}}) ->
                       TxnMod = blockchain_txn:type(Txn),
                       TxnHash = blockchain_txn:hash(Txn),
                       [
                        {txn_type, atom_to_list(TxnMod)},
                        {txn_hash, io_lib:format("~p", [libp2p_crypto:bin_to_b58(TxnHash)])},
-                       {retries, integer_to_list(Retries)}
+                       {acceptions, length(Acceptions)},
+                       {rejections, length(Rejections)},
+                       {accepted_block_height, AcceptedBlockHeight}
```
vihu: Would be nice to print `RecvBlockHeight` here too?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +57,300 @@ submit(Txn, Callback) ->
...
-                                            end
-                                    end,
-                                    TxnMap,
-                                    SortedTxns),
-
-            {noreply, State#state{txn_map=NewTxnMap}};
+            BlockHeight = blockchain_block:height(Block),
+            MinedTxns = blockchain_block:transactions(Block),
+            %% delete any txn which is included in the new block from our cache
+            lists:foreach(
```
vihu: nit, `ok = lists:foreach...`

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +57,300 @@ submit(Txn, Callback) ->
...
+                        false ->
+                            ok
+                    end
+                end, sorted_cached_txns()),
+
+            %% resubmit the txns remaining in cache
+            resubmit(Chain, BlockHeight, SubmitF),
+            {noreply, State#state{cur_block_height = BlockHeight}};
         _ ->
             lager:error("WTF happened!"),
```
vihu: Let's change this error to something meaningful while we're here, `{error, block_not_found}` or something like that maybe?

src/cli/blockchain_cli_txn.erl
andymck: its already there,  "accepted_block_height"

src/transactions/blockchain_txn_mgr.erl
andymck: can do

src/transactions/blockchain_txn_mgr.erl
andymck: The terminate has some cleanup and this will ensure it is performed in all scenarios including shutdown

src/transactions/blockchain_txn_mgr.erl
andymck: it has no current need to be public, in an older version i was spawning some functions off which would need access to the table.  This is no longer the case, will change

src/transactions/blockchain_txn_mgr.erl
andymck: yip, will do

src/transactions/blockchain_txn_mgr.erl
andymck: will do

src/transactions/blockchain_txn_mgr.erl
andymck: yeah good point, i have read that 100 times and thought the same but then overlooked it

src/transactions/blockchain_txn_mgr.erl
andymck: Done

src/transactions/blockchain_txn_mgr.erl
andymck: Done

src/transactions/blockchain_txn_mgr.erl
andymck: Done

src/transactions/blockchain_txn_mgr.erl
andymck: Done

src/transactions/blockchain_txn_mgr.erl
andymck: Done

src/cli/blockchain_cli_txn.erl
vihu: Ah sorry, I was confused by the naming there. We call it `RecvBlockHeight` in the `cache_txn_type` definition.

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +61,299 @@ submit(Txn, Callback) ->
...
             %% rescue block! no signatures we can use
             %% so use a random consensus member
             Ledger = blockchain:ledger(Chain),
             {ok, Members0} = blockchain_ledger_v1:consensus_members(Ledger),
-            Members = Members0 -- [blockchain_swarm:pubkey_bin()] -- Rejections,
-            Index = rand:uniform(length(Members)),
-            Member = lists:nth(Index, Members),
-            {ok, Member};
+            Members = Members0 -- [blockchain_swarm:pubkey_bin()],
+            RandomMembers = randomize_list(Members),
```
vihu: Why not use `blockchain_utils:shuffle/1` for this?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +61,299 @@ submit(Txn, Callback) ->
...
+    lists:sort(fun({TxnA, _}, {TxnB, _}) -> blockchain_txn:sort(TxnA, TxnB) end, Txns).
+
+-spec normalise_block_height(integer(), undefined | integer()) -> integer().
+normalise_block_height(CurBlockHeight, undefined)->
+    CurBlockHeight;
+normalise_block_height(_CurBlockHeight, RecvBlockHeight)->
+    RecvBlockHeight.
+
+submit_f(NumMembers)->
+    (NumMembers - 1) div 3 * 2.
```
vihu:
```suggestion
    ((NumMembers - 1) div 3 ) * 2.
```

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +61,299 @@ submit(Txn, Callback) ->
...
         [] ->
             %% rescue block! no signatures we can use
             %% so use a random consensus member
             Ledger = blockchain:ledger(Chain),
             {ok, Members0} = blockchain_ledger_v1:consensus_members(Ledger),
-            Members = Members0 -- [blockchain_swarm:pubkey_bin()] -- Rejections,
-            Index = rand:uniform(length(Members)),
-            Member = lists:nth(Index, Members),
-            {ok, Member};
+            Members = Members0 -- [blockchain_swarm:pubkey_bin()],
```
vihu: I think you do want to eliminate the members who have rejected the txn here too and try different members, like we were doing before?

andymck: In this new implementation we blast the txn out to 2F+1 members and then await responses.  Within the same block epoch we do not resend to any additional members, even if the block is rejected by all members.  At the next add_block event the block will either be included in the new block or resent.

I took this approach to avoid resending an obviously invalid or temporarily  unacceptable txn over and over again to the CG.  In the previous implementation we would sent to one CG member at a time and then keep sending to a new one each time it was rejected.  This felt kind of spammy to me and it forces the CG members to rerun the expensive validations each time.  
Now we blast from the outset to 2F+1 members and leave it at that until after the next add block event.  So this is a balance between sending to sufficient CG members in order to give it a chance to be accepted and not spamming them with txns which cannot be accepted either permanently or temporarily.

In an earlier version I did do resends to diff members upon rejections but changed to the above.  Let me know if you think this approach in sub optimal or if there is something I am maybe not considering and I can redo.

src/transactions/blockchain_txn_mgr.erl
andymck: I think i forgot it was there...will switch to using it

src/cli/blockchain_cli_txn.erl
andymck: Made it consistent, now using RecvBlockHeight in blockchain_cli_txn

src/transactions/blockchain_txn_mgr.erl
vihu: So, I don't _think_ we need to blast to `2F + 1` members as soon as we receive a txn. `2F+1` is the guarantee for txn acceptance. But in most situations you'd likely have a single honest cg member accept the txn, put it in its buffer and it will appear in a block (assuming it was valid of course).

Also note that we can send to `F+1` members and be pretty certain that at the very least one honest node got the txn.

That being said, I think even something like `F/2 + 1` would work reasonably well. So currently that would be 3 members (F=5). If a single one of them rejected the txn, just find a different new previous signatory member and resend. If you have `F+1` rejections drop that txn. I think this would be potentially faster than doing the current approach. Thoughts?

andymck: k, i will change to submit to F/2+1 and resend upon each rejection until we either have F/2+1 accepts or F+1 rejections.  At that point will cease submits.

andymck: @vihu fyi changes above have now been added

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +61,321 @@ submit(Txn, Callback) ->
...
+    spawn(fun() -> Callback(Msg) end),
+    ok.
 
-signatory_rand_member(Rejections, Chain) ->
+-spec signatory_rand_members(blockchain:blockchain(), integer(), [libp2p_crypto:pubkey_bin()], [libp2p_crypto:pubkey_bin()]) -> {ok, [libp2p_crypto:pubkey_bin()]}.
+signatory_rand_members(Chain, SubmitF, Acceptions, Rejections) ->
     {ok, PrevBlock} = blockchain:head_block(Chain),
-    Signatories = [Signer || {Signer, _} <- blockchain_block:signatures(PrevBlock), not lists:member(Signer, [blockchain_swarm:pubkey_bin()|Rejections])],
+    Signatories = [Signer || {Signer, _} <- blockchain_block:signatures(PrevBlock),
+        not Signer =:= blockchain_swarm:pubkey_bin()],
```
Vagabond: The precedence is wrong here
```suggestion
        not (Signer =:= blockchain_swarm:pubkey_bin())],
```

example:

```
1> not 1 =:= 1.
** exception error: bad argument
     in operator  not/1
        called as not 1
2> not (1 =:= 1).
false
```

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +61,321 @@ submit(Txn, Callback) ->
...
-                            retry(Txn, State#state{txn_map=maps:put(Txn, {Callback, [], Dialer}, TxnMap)})
-                    end;
-                false ->
-                    retry(Txn, State#state{txn_map=maps:put(Txn, {Callback, [Member|Rejections], Dialer}, TxnMap)})
+-spec resubmit(blockchain:blockchain(), undefined | integer(), integer()) -> ok.
+resubmit(Chain, CurBlockHeight, SubmitF)->
+    %% get a sorted list of the cached txns
+    CachedTxns = sorted_cached_txns(),
+    %% validate the cached txns
+    {ValidTransactions, InvalidTransactions} = blockchain_txn:validate(CachedTxns, Chain),
```
Vagabond: validate crashes here because CachedTxns are not a list of transactions, but a list of transactions with their metadata attached.
```suggestion
    {Txns, _} = lists:unzip(CachedTxns),
    {ValidTransactions, InvalidTransactions} = blockchain_txn:validate(Txns, Chain),
```

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +61,321 @@ submit(Txn, Callback) ->
...
+    ok = rejected(Chain, Txn, Member, Dialer, CurBlockHeight, RejectF),
+    {noreply, State};
+
 handle_info({blockchain_event, {new_chain, NC}}, State) ->
-    {noreply, State#state{chain = NC}};
-handle_info({blockchain_event, {add_block, BlockHash, _Sync, _Ledger}}, State=#state{chain=Chain, txn_map=TxnMap}) ->
+    NewState = initialize_with_chain(State, NC),
+    {noreply, NewState};
+
+handle_info({blockchain_event, {add_block, BlockHash, _Sync, _Ledger}}, State=#state{chain = Chain, submit_f = SubmitF}) ->
```
Vagabond: Chain can be undefined here, which causes get_block to go boom

Vagabond:
```suggestion
handle_info({blockchain_event, {add_block, BlockHash, _Sync, _Ledger}}, State0=#state{chain = Chain0}) ->
    State = case Chain0 of
        undefined ->
            NC = blockchain_worker:blockchain(),
            initialize_with_chain(State0, NC);
        _ ->
            State0
    end,
    #state{submit_f = SubmitF, chain = Chain} = State,
```

Vagabond:
```
2020-03-06 18:47:13 =ERROR REPORT====
** Generic server blockchain_txn_mgr terminating
** Last message in was {blockchain_event,{add_block,<<26,111,217,144,140,17,131,113,147,74,101,183,224,131,245,83,87,185,213,128,184,81,133,53,90,188,27,228,173,165,40,82>>,false,{ledger_v1,"/home/andrew/miner/_build/test/logs/ct_run.nonode@nohost.2020-03-06_18.41.51/lib.miner.logs/run.2020-03-06_18.41.51/log_private/data/miner_blockchain_SUITE_master_key_test_1tysve@system76",#Ref<0.2057296762.4183425041.85622>,#Ref<0.2057296762.4183425025.86680>,active,{sub_ledger_v1,#Ref<0.2057296762.4183425041.85623>,#Ref<0.2057296762.4183425041.85624>,#Ref<0.2057296762.4183425041.85625>,#Ref<0.2057296762.4183425041.85626>,#Ref<0.2057296762.4183425041.85627>,#Ref<0.2057296762.4183425041.85628>,#Ref<0.2057296762.4183425041.85629>,#Ref<0.2057296762.4183425041.85630>,undefined},{sub_ledger_v1,#Ref<0.2057296762.4183425041.85631>,#Ref<0.2057296762.4183425041.85632>,#Ref<0.2057296762.4183425041.85633>,#Ref<0.2057296762.4183425041.85634>,#Ref<0.2057296762.4183425041.85635>,#Ref<0.2057296762.4183425041.85636>,#Ref<0.2057296762.4183425041.85637>,#Ref<0.2057296762.4183425041.85638>,undefined},#Ref<0.2057296762.4183425032.107077>}}}
** When Server state == {state,2,3,47,txn_cache,{blockchain,"/home/andrew/miner/_build/test/logs/ct_run.nonode@nohost.2020-03-06_18.41.51/lib.miner.logs/run.2020-03-06_18.41.51/log_private/data/miner_blockchain_SUITE_master_key_test_1tysve@system76",#Ref<0.2057296762.4183425041.85616>,#Ref<0.2057296762.4183425041.85617>,#Ref<0.2057296762.4183425041.85618>,#Ref<0.2057296762.4183425041.85619>,#Ref<0.2057296762.4183425041.85620>,{ledger_v1,"/home/andrew/miner/_build/test/logs/ct_run.nonode@nohost.2020-03-06_18.41.51/lib.miner.logs/run.2020-03-06_18.41.51/log_private/data/miner_blockchain_SUITE_master_key_test_1tysve@system76",#Ref<0.2057296762.4183425041.85622>,#Ref<0.2057296762.4183425025.86680>,active,{sub_ledger_v1,#Ref<0.2057296762.4183425041.85623>,#Ref<0.2057296762.4183425041.85624>,#Ref<0.2057296762.4183425041.85625>,#Ref<0.2057296762.4183425041.85626>,#Ref<0.2057296762.4183425041.85627>,#Ref<0.2057296762.4183425041.85628>,#Ref<0.2057296762.4183425041.85629>,#Ref<0.2057296762.4183425041.85630>,undefined},{sub_ledger_v1,#Ref<0.2057296762.4183425041.85631>,#Ref<0.2057296762.4183425041.85632>,#Ref<0.2057296762.4183425041.85633>,#Ref<0.2057296762.4183425041.85634>,#Ref<0.2057296762.4183425041.85635>,#Ref<0.2057296762.4183425041.85636>,#Ref<0.2057296762.4183425041.85637>,#Ref<0.2057296762.4183425041.85638>,undefined},undefined}}}
** Reason for termination ==
** {{noproc,{gen_server,call,[blockchain_txn_mgr_sup,{terminate_child,<0.5239.0>},infinity]}},[{gen_server,call,3,[{file,"gen_server.erl"},{line,223}]},{blockchain_txn_mgr_sup,'-stop_dialers/1-lc$^0/1-0-',1,[{file,"/home/andrew/miner/_build/default/lib/blockchain/src/transactions/blockchain_txn_mgr_sup.erl"},{line,47}]},{blockchain_txn_mgr_sup,stop_dialers,1,[{file,"/home/andrew/miner/_build/default/lib/blockchain/src/transactions/blockchain_txn_mgr_sup.erl"},{line,47}]},{blockchain_txn_mgr,'-resubmit/3-fun-2-',6,[{file,"/home/andrew/miner/_build/default/lib/blockchain/src/transactions/blockchain_txn_mgr.erl"},{line,248}]},{lists,foreach,2,[{file,"lists.erl"},{line,1338}]},{blockchain_txn_mgr,resubmit,3,[{file,"/home/andrew/miner/_build/default/lib/blockchain/src/transactions/blockchain_txn_mgr.erl"},{line,245}]},{blockchain_txn_mgr,handle_info,2,[{file,"/home/andrew/miner/_build/default/lib/blockchain/src/transactions/blockchain_txn_mgr.erl"},{line,168}]},{gen_server,try_dispatch,4,[{file,"gen_server.erl"},{line,637}]}]}
2020-03-06 18:47:13 =CRASH REPORT====
  crasher:
    initial call: blockchain_txn_mgr:init/1
    pid: <0.334.0>
    registered_name: blockchain_txn_mgr
    exception exit: {{noproc,{gen_server,call,[blockchain_txn_mgr_sup,{terminate_child,<0.5239.0>},infinity]}},[{gen_server,call,3,[{file,"gen_server.erl"},{line,223}]},{blockchain_txn_mgr_sup,'-stop_dialers/1-lc$^0/1-0-',1,[{file,"/home/andrew/miner/_build/default/lib/blockchain/src/transactions/blockchain_txn_mgr_sup.erl"},{line,47}]},{blockchain_txn_mgr_sup,stop_dialers,1,[{file,"/home/andrew/miner/_build/default/lib/blockchain/src/transactions/blockchain_txn_mgr_sup.erl"},{line,47}]},{blockchain_txn_mgr,'-resubmit/3-fun-2-',6,[{file,"/home/andrew/miner/_build/default/lib/blockchain/src/transactions/blockchain_txn_mgr.erl"},{line,248}]},{lists,foreach,2,[{file,"lists.erl"},{line,1338}]},{blockchain_txn_mgr,resubmit,3,[{file,"/home/andrew/miner/_build/default/lib/blockchain/src/transactions/blockchain_txn_mgr.erl"},{line,245}]},{blockchain_txn_mgr,handle_info,2,[{file,"/home/andrew/miner/_build/default/lib/blockchain/src/transactions/blockchain_txn_mgr.erl"},{line,168}]},{gen_server,try_dispatch,4,[{file,"gen_server.erl"},{line,637}]}]}
    ancestors: [blockchain_sup,miner_sup,<0.269.0>]
    message_queue_len: 1
    messages: [{'EXIT',<0.272.0>,shutdown}]
    links: []
    dictionary: [{rand_seed,{#{jump => #Fun<rand.16.8986388>,max => 18446744073709551615,next => #Fun<rand.15.8986388>,type => exs1024},{[7837407313690587008,15056810970685313164,18030000405143231605,14795351861972892642,12124445732987077783,6873173150464907114,14984644242634745786,1382256195557956762,6599361407610914522,6071727814966107327,10482934821067028262,8600477661232989710,14432769999808943703,152560699368799132,766472080451672109],[17325062742087844586]}}}]
    trap_exit: true
    status: running
    heap_size: 17731
    stack_size: 27
    reductions: 1270569
  neighbours:
```

Vagabond:
```
diff --git a/src/transactions/blockchain_txn_mgr_sup.erl b/src/transactions/blockchain_txn_mgr_sup.erl
index 2d986eeb..0e93bc4d 100644
--- a/src/transactions/blockchain_txn_mgr_sup.erl
+++ b/src/transactions/blockchain_txn_mgr_sup.erl
@@ -44,5 +44,5 @@ stop_dialer(Pid) ->
     supervisor:terminate_child(?MODULE, Pid).

 stop_dialers(Pids) ->
-    [supervisor:terminate_child(?MODULE, Pid) || Pid <- Pids],
-    ok.
\ No newline at end of file
+    [catch supervisor:terminate_child(?MODULE, Pid) || Pid <- Pids],
+    ok.
```
propose we do something like this to make shutdown cleaner

Vagabond:
So, I've rebased this, applied my proposed fixes and I've gotten miner's test suite to pass with this new txn manager. I need to review the logic a bit more but things seem structurally correct now.

Vagabond:
My branch can be found at `adt/ch6193/rework-txn-manager-rebased`, for reference

andymck:
@Vagabond great, thanks for reviewing/testing...tbh i dont think really paid much attention to the test failures when pushing this at the time.  Kinda had it in my head they were just general test noise failures.  Had meant to come back too and then got distracted and forgot, so apols for you having to iron those issues out.  When you are done with your review we should just go ahead and merge your branch rather than accepting your suggestions here.

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +61,321 @@ submit(Txn, Callback) ->
...
-            {ok, Member};
+            Members = Members0 -- [blockchain_swarm:pubkey_bin()] -- Acceptions -- Rejections,
+            RandomMembers = blockchain_utils:shuffle(Members),
+            {ok, lists:sublist(RandomMembers, SubmitF)};
         _ ->
-            Index = rand:uniform(length(Signatories)),
-            Signer= lists:nth(Index, Signatories),
-            {ok, Signer}
+            %% we have signatories
+            RandomSignatories = blockchain_utils:shuffle(Signatories) -- Acceptions -- Rejections,
```
Vagabond: What is the desired precedence here?

```
2> [1, 2] -- [1,2] -- [1].
[1]
3> [1, 2] -- ([1,2] -- [1]).
[1]
4> ([1, 2] -- [1,2]) -- [1].
[]
```

andymck: aha well spotted, it should be the third in your list, so ([1, 2] -- [1,2]) -- [1].

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -52,204 +61,321 @@ submit(Txn, Callback) ->
...
+    delete_cached_txn(Txn);
+%% the txn has been rejected but has not yet exceeded the max number of rejections,
+%% so resend to another CG member
+reject_actions(Chain,
+                {Txn, {Callback, RecvBlockHeight, Acceptions, Rejections, Dialers}},
+                NewRejections,
+                RejectF,
+                _CurBlockHeight)
+    when length(NewRejections) < RejectF ->
+    NewDialer = submit_txn_to_cg(Chain, Txn, 1, Acceptions, Rejections),
```
Vagabond: Should this not be NewRejections here?

andymck: Yes, it should indeed be NewRejections

PR #375, opened 2020-02-19
lazily GC stale witnesses

andymck:
With this solution, stale witnesses ( ie witnesses added to a challengee and where the witness GW has subsequently reasserted its location ) are purged from the challengee GW witness map when the next witness is added to the challengee witness list.

It may mean the ledger will contain more stale data for longer but the solution is cheap and thus easier to scale.

The implementation adds new fields to the following records:

#gateway_v2.last_location_nonce
When a GW asserts its location, the value of the GWs ledger nonce is set as the value of last_location_nonce. 

#witness.added_location_nonce
When a witness is added, the last_location_nonce value of the witness GW is recorded here.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_gateway_v2.erl
+++ b/src/ledger/v1/blockchain_ledger_gateway_v2.erl
@@ -28,11 +28,11 @@
     beta/1,
     delta/1,
     set_alpha_beta_delta/4,
-    add_witness/1, add_witness/5,
-    has_witness/4,
+    add_witness/1, add_witness/7,
+    has_witness/2,
     clear_witnesses/1,
     remove_witness/2,
-    witnesses/1, witnesses/3,
+    witnesses/1,
     witnesses_plain/1,
     witness_hist/1, witness_recent_time/1, witness_first_time/1,
     oui/1, oui/2,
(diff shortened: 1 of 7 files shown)


src/ledger/v1/blockchain_ledger_gateway_v2.erl
```diff
@@ -393,13 +419,13 @@ clear_witnesses(Gateway) ->
...
-    maps:is_key(WitnessAddr, Witnesses).
+-spec has_witness(libp2p_crypto:pubkey_bin(), gateway(), libp2p_crypto:pubkey_bin(), blockchain_ledger_v1:ledger()) -> boolean().
+has_witness(GatewayBin, Gateway, WitnessAddr, Ledger) ->
+    maps:is_key(WitnessAddr, purge_stale_witnesses(GatewayBin, Gateway, Ledger)).
 
--spec witnesses(gateway()) -> #{libp2p_crypto:pubkey_bin() => gateway_witness()}.
-witnesses(Gateway) ->
-    Gateway#gateway_v2.witnesses.
+-spec witnesses(libp2p_crypto:pubkey_bin(), gateway(), blockchain_ledger_v1:ledger()) -> #{libp2p_crypto:pubkey_bin() => gateway_witness()}.
+witnesses(GatewayBin, Gateway, Ledger) ->
```
Vagabond, line 427: If we purged on `add_witness`, we'd be more deterministic, right?

Vagabond: And just filtered stale witnesses from the return here.

Vagabond: Because add_witness happens deterministically depending on what happens in a block, but reading the witness list can be done for other reasons that are not deterministic.

src/ledger/v1/blockchain_ledger_gateway_v2.erl
```diff
@@ -157,6 +161,21 @@ mode(Gateway) ->
 -spec mode(Mode :: mode(), Gateway :: gateway()) -> gateway().
 mode(Mode, Gateway) ->
     Gateway#gateway_v2{mode=Mode}.
+%%--------------------------------------------------------------------
```
vihu, line 164: nitpicking mostly, but either have the docs or don't :)

src/ledger/v1/blockchain_ledger_gateway_v2.erl
```diff
@@ -664,6 +735,32 @@ mask_for_gateway_mode(#gateway_v2{mode = full}, Ledger)->
...
+-spec filter_stale_witnesses(libp2p_crypto:pubkey_bin(), gateway(), blockchain_ledger_v2:ledger(), boolean()) -> gateway().
+filter_stale_witnesses(GatewayBin, GatewayToPurge = #gateway_v2{witnesses = Witnesses}, Ledger, PurgeGW)->
+    PurgedGW =
+        lists:foldl(fun
+                      ({WitnessPubkeyBin, Witness}, AccGW)->
+                          %% get the witness GW
+                          %% check when it was last updated location and whether it has done so since this witness report was added
+                          %% to the GW being purged
+                          {ok, WitnessGW} = blockchain_ledger_v1:find_gateway_info(WitnessPubkeyBin, Ledger),
+                          case WitnessGW#gateway_v2.last_location_nonce > Witness#witness.added_location_nonce of
```
vihu, line 748: Since both of these can be `undefined` do we need to handle that scenario?

src/ledger/v1/blockchain_ledger_gateway_v2.erl
evanmcc: don't, ideally

evanmcc (approved):
this looks good to me, an earlier version has been running problem free locally for quite a while and the performance improvement is noticable.

vihu:
@andymck I've added https://github.com/helium/blockchain-core/pull/886, plz double check if that looks ok

PR #378, opened 2020-02-24
support nonce in htlc txns

andymck:
NOTE: this PR is dependant on proto PR: https://github.com/helium/proto/pull/19

This adds a nonce to the create_htlc txn and uses this to avoid replays/double spends.

### DIFF
--- a/rebar.lock
+++ b/rebar.lock
@@ -33,7 +33,7 @@
   0},
  {<<"helium_proto">>,
   {git,"https://github.com/helium/proto.git",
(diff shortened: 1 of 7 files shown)


Vagabond:
This looks correct, nice job

vihu:
Can we run this with the nonce field proto change to get a clean run on kite? We can switch branches back to master(s) when merging

andymck:
@vihu done, tests passed when pointing to the proto commit

Vagabond:
Let's get this merged up

PR #382, opened 2020-02-27
Add payment_v2 transactions

vihu:
Adds a new transaction `blockchain_txn_payment_v2` which allows multi payee payment transactions. So we can eventually start supporting txns like: `payerA -> [payeeX, payeeY, payeeZ]`

Couple things to note:
1. The only nonce which we care about is the nonce for payerA
2. We check whether the whole txn is valid by summing the amounts being paid to X, Y and Z.
3. There is no internal ordering for the payout to X, Y or Z as it doesn't really matter I think.
4. A single payment looks like `payerA -> [payeeX]`
5. The max limit of number of payments in a single txn is configured via `max_payments` chain var.

Proto: https://github.com/helium/proto/pull/20

### DIFF
--- a/src/transactions/blockchain_txn.erl
+++ b/src/transactions/blockchain_txn.erl
@@ -30,7 +30,8 @@
              | blockchain_txn_token_burn_v1:txn_token_burn()
              | blockchain_txn_dc_coinbase_v1:txn_dc_coinbase()
              | blockchain_txn_token_burn_exchange_rate_v1:txn_token_burn_exchange_rate()
-             | blockchain_txn_bundle_v1:txn_bundle().
+             | blockchain_txn_bundle_v1:txn_bundle()
+             | blockchain_txn_payment_v2:txn_payment_v2().
 
 -type txns() :: [txn()].
 -export_type([hash/0, txn/0, txns/0]).
@@ -90,7 +91,8 @@
     {blockchain_txn_assert_location_v1, 15},
     {blockchain_txn_redeem_htlc_v1, 16},
     {blockchain_txn_poc_request_v1, 17},
-    {blockchain_txn_poc_receipts_v1, 18}
+    {blockchain_txn_poc_receipts_v1, 18},
+    {blockchain_txn_payment_v2, 19}
 ]).
 
 block_delay() ->
@@ -150,6 +152,8 @@ wrap_txn(#blockchain_txn_dc_coinbase_v1_pb{}=Txn) ->
     #blockchain_txn_pb{txn={dc_coinbase, Txn}};
 wrap_txn(#blockchain_txn_token_burn_exchange_rate_v1_pb{}=Txn) ->
     #blockchain_txn_pb{txn={token_burn_exchange_rate, Txn}};
+wrap_txn(#blockchain_txn_payment_v2_pb{}=Txn) ->
+    #blockchain_txn_pb{txn={payment_v2, Txn}};
 wrap_txn(#blockchain_txn_bundle_v1_pb{transactions=Txns}=Txn) ->
     #blockchain_txn_pb{txn={bundle, Txn#blockchain_txn_bundle_v1_pb{transactions=lists:map(fun wrap_txn/1, Txns)}}}.
 
@@ -376,7 +380,8 @@ absorb(Txn, Chain) ->
     Type = ?MODULE:type(Txn),
     try Type:absorb(Txn, Chain) of
         {error, _Reason}=Error ->
-            lager:info("failed to absorb ~p ~p", [Type, _Reason]),
+            lager:info("failed to absorb ~p ~p ~s",
+                       [Type, _Reason, ?MODULE:print(Txn)]),
             Error;
         ok -> ok
     catch
@@ -470,7 +475,9 @@ type(#blockchain_txn_dc_coinbase_v1_pb{}) ->
 type(#blockchain_txn_token_burn_exchange_rate_v1_pb{}) ->
     blockchain_txn_token_burn_exchange_rate_v1;
 type(#blockchain_txn_bundle_v1_pb{}) ->
-    blockchain_txn_bundle_v1.
+    blockchain_txn_bundle_v1;
+type(#blockchain_txn_payment_v2_pb{}) ->
+    blockchain_txn_payment_v2.
 
 %% ------------------------------------------------------------------
 %% Internal Function Definitions
@@ -578,6 +585,8 @@ nonce(Txn) ->
             blockchain_txn_payment_v1:nonce(Txn);
         blockchain_txn_security_exchange_v1 ->
             blockchain_txn_security_exchange_v1:nonce(Txn);
+        blockchain_txn_payment_v2 ->
+            blockchain_txn_payment_v2:nonce(Txn);
         _ ->
             -1 %% other transactions sort first
     end.
@@ -617,6 +626,8 @@ actor(Txn) ->
             blockchain_txn_token_burn_v1:payer(Txn);
         blockchain_txn_dc_coinbase_v1 ->
             blockchain_txn_dc_coinbase_v1:payee(Txn);
+        blockchain_txn_payment_v2 ->
+            blockchain_txn_payment_v2:payer(Txn);
         _ ->
             <<>>
     end.
--- a/src/transactions/v2/blockchain_payment_v2.erl
+++ b/src/transactions/v2/blockchain_payment_v2.erl
@@ -1,0 +1,64 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain Payment V2 ==
+%%%-------------------------------------------------------------------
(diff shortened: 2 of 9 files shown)


src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -0,0 +1,286 @@
...
+         print/1
+        ]).
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-type txn_payment_v2() :: #blockchain_txn_payment_v2_pb{}.
+
+%% TODO: Maybe chain var
```
evanmcc: definitely.  Otherwise if we need to make it smaller for any reason, old txns will stop validating.  I would also set this lower, maybe 20?  I dunno I don't feel super strongly, but I'd like to bias towards keeping block sizes down.

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -0,0 +1,286 @@
...
+                                    case blockchain_ledger_v1:transaction_fee(Ledger) of
+                                        {error, _}=Error0 ->
+                                            Error0;
+                                        {ok, MinerFee} ->
+                                            case (TotAmount >= 0) andalso (Fee >= MinerFee) of
+                                                false ->
+                                                    {error, invalid_transaction};
+                                                true ->
+                                                    case blockchain_ledger_v1:check_dc_balance(Payer, Fee, Ledger) of
+                                                        {error, _}=Error1 ->
```
evanmcc: my kingdom for a pipe operator

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -0,0 +1,286 @@
...
+
+-spec print(txn_payment_v2()) -> iodata().
+print(undefined) -> <<"type=payment_v2, undefined">>;
+print(#blockchain_txn_payment_v2_pb{payer=Payer,
+                                    fee=Fee,
+                                    payments=Payments,
+                                    nonce=Nonce,
+                                    signature = S}=Txn) ->
+    io_lib:format("type=payment_v2, payer=~p, payments=~p, tot_amount=~p, fee=~p, nonce=~p, signature=~p",
+                  [?TO_B58(Payer), print_payments(Payments), ?MODULE:total_amount(Txn), Fee, Nonce, S]).
```
evanmcc: move the total before the individuals so it doesn't get truncated in big txns

vihu: Added: a93d74b

src/transactions/v2/blockchain_txn_payment_v2.erl
vihu: Updated.

src/transactions/blockchain_txn.erl
```diff
@@ -376,7 +380,7 @@ absorb(Txn, Chain) ->
     Type = ?MODULE:type(Txn),
     try Type:absorb(Txn, Chain) of
         {error, _Reason}=Error ->
-            lager:info("failed to absorb ~p ~p", [Type, _Reason]),
+            lager:info("failed to absorb ~p ~p ~p", [Type, _Reason, Txn]),
```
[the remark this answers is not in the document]
vihu: Done

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -0,0 +1,301 @@
...
+nonce(Txn) ->
+    Txn#blockchain_txn_payment_v2_pb.nonce.
+
+-spec signature(txn_payment_v2()) -> binary().
+signature(Txn) ->
+    Txn#blockchain_txn_payment_v2_pb.signature.
+
+%%--------------------------------------------------------------------
+%% @doc
+%% NOTE: payment transactions can be signed either by a worker who's part of the blockchain
```
Vagabond: WHAT?

vihu: oh, cruft from copy, gonna remove

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -0,0 +1,294 @@
...
+    end.
+
+-spec print(txn_payment_v2()) -> iodata().
+print(undefined) -> <<"type=payment_v2, undefined">>;
+print(#blockchain_txn_payment_v2_pb{payer=Payer,
+                                    fee=Fee,
+                                    payments=Payments,
+                                    nonce=Nonce,
+                                    signature = S}=Txn) ->
+    io_lib:format("type=payment_v2, tot_amount=~p, payer=~p, payments=~p, fee=~p, nonce=~p, signature=~p",
```
Vagabond: Shouldn't payments be printed with ~s as they're a string from print_payments?

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -0,0 +1,294 @@
...
+                        false ->
+                            case lists:member(Payer, ?MODULE:payees(Txn)) of
+                                false ->
+                                    TotAmount = ?MODULE:total_amount(Txn),
+                                    Fee = ?MODULE:fee(Txn),
+                                    case blockchain_ledger_v1:transaction_fee(Ledger) of
+                                        {error, _}=Error0 ->
+                                            Error0;
+                                        {ok, MinerFee} ->
+                                            case (TotAmount >= 0) andalso (Fee >= MinerFee) of
```
Vagabond: Presumably there should also not be a payee with a 0 amount?

vihu: Yeah, also, negative amounts in payments shouldn't be allowed. Good catch.

src/transactions/v2/blockchain_txn_payment_v2.erl
vihu: Will fix.

src/transactions/v2/blockchain_payment_v2.erl
```diff
@@ -0,0 +1,64 @@
...
+-endif.
+
+-type payment() :: #payment_pb{}.
+-type payments() :: [payment()].
+
+-export_type([payment/0, payments/0]).
+
+-spec new(Payee :: libp2p_crypto:pubkey_bin(),
+          Amount :: non_neg_integer()) -> payment().
+new(Payee, Amount) ->
```
Vagabond:
```suggestion
new(Payee, Amount) when Amount > 0 ->
```

vihu: Yup, probably better to disallow such payment creations. Updated.

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -0,0 +1,296 @@
...
+    ?assertEqual(1, nonce(Tx)).
+
+signature_test() ->
+    Payments = [blockchain_payment_v2:new(<<"x">>, 10),
+                blockchain_payment_v2:new(<<"y">>, 20),
+                blockchain_payment_v2:new(<<"z">>, 30)],
+    Tx = new(<<"payer">>, Payments, 1, 10),
+    ?assertEqual(<<>>, signature(Tx)).
+
+sign_test() ->
```
Vagabond: It would be good to have a test that tested some negative conditions, like 0 or negative amounts, or a bad signature

test/blockchain_payment_v2_SUITE.erl
```diff
@@ -0,0 +1,247 @@
...
+    ConsensusMembers = ?config(consensus_members, Config),
+    Balance = ?config(balance, Config),
+    BaseDir = ?config(base_dir, Config),
+    Chain = ?config(chain, Config),
+    Swarm = ?config(swarm, Config),
+
+    %% Test a payment transaction, add a block and check balances
+    [_, {Payer, {_, PayerPrivKey, _}}|_] = ConsensusMembers,
+
+    %% Create a payment to a single payee TWICE
```
Vagabond: Arguably this should be invalid, because it's just wasting space

src/transactions/v2/blockchain_txn_payment_v2.erl
vihu: Added both

test/blockchain_payment_v2_SUITE.erl
Vagabond: Thoughts @vihu ?

vihu: Yeah, I believe if A _really_ wants to pay B twice just make the txn with the sum of the amounts instead of n separate payments. Otherwise it's possible a notorious entity just starts flooding with several minuscule payments.

I'll make it so.

vihu: Updated: 4e22608

PR #383, opened 2020-02-27
Implement a checkpoint cli

mrallen1:
Problem to solve: We want a way to write a periodic snapshot of the rocksdb so that we can store it in S3 (for backup and cloning purposes)

Solution: Implement a `rocksdb checkpoint` command line interface.

### DIFF
--- a/scripts/extensions/checkpoint
+++ b/scripts/extensions/checkpoint
@@ -2,5 +2,5 @@
 if [ -t 0 ] ; then
     export CLIQUE_COLUMNS=`stty size 2>/dev/null | cut -d ' ' -f 2`
 fi
-relx_nodetool rpc blockchain_console command rocksdb $@
+relx_nodetool rpc blockchain_console command checkpoint $@
 exit $?
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -29,6 +29,8 @@
 
     analyze/1, repair/1,
 
+    checkpoint/2,
+
     fold_chain/4,
 
     reset_ledger/1, reset_ledger/2, reset_ledger/3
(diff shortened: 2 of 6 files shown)


madninja:
Maybe rename it to `checkpoint` and take an argument for what "service" to checkpoint. Only support `blockchain`

We use rocksdb in multiple places, and this may be useful to capture other snapshots like peerbook or ledger at some point

scripts/extensions/checkpoint
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+if [ -t 0 ] ; then
+    export CLIQUE_COLUMNS=`stty size 2>/dev/null | cut -d ' ' -f 2`
```
evanmcc: does this reset the console afterwards?

src/blockchain.erl
```diff
@@ -1271,6 +1273,17 @@ missing_block(#blockchain{db=DB, default=DefaultCF}) ->
             not_found
     end.
 
+checkpoint(undefined, _Directory) -> {error, no_blockchain};
+checkpoint(#blockchain{db=DB}, Directory) ->
+   case filelib:is_file(Directory) of
```
evanmcc: Ideally we'd take the blockchain lock here to make sure that we stay block-aligned while this happens.

src/blockchain.erl
```diff
@@ -1271,6 +1273,17 @@ missing_block(#blockchain{db=DB, default=DefaultCF}) ->
...
+      true ->
+         %% This file/directory exists - it must *not* exist
+         {error, directory_exists};
```
evanmcc: I think that this is kind of suboptimal, ui-wise.  my preference would be to have a canonical directory "data/backups/" or whatever, and make a unique subfolder for the checkout named for the block height (e.g. "data/backups/backup.231412".

src/cli/blockchain_cli_checkpoint.erl
```diff
@@ -0,0 +1,69 @@
...
+checkpoint_blockchain_cmd() ->
+    [
+     [["checkpoint", "blockchain", '*'], [], [], fun do_checkpoint/3]
+    ].
+
+do_checkpoint(["checkpoint", "blockchain", Directory], [], []) ->
+    Blockchain = blockchain_worker:blockchain(),
+    case blockchain:checkpoint(Blockchain, Directory) of
+        ok ->
+            [clique_status:text("ok")];
```
evanmcc: if you do go with the naming/ui suggestions above, this might be a good place to spit out the directory name so the user can find it.

scripts/extensions/checkpoint
Vagabond: This just queries the tty for its size, we've done it for ages in all the other cli helpers

PR #384, opened 2020-02-28
More state channel fixes

vihu:
This addresses two issues with the state_channel work:

* Encapsulating the `helium_packet` into its own module. This avoids having to include the proto in tests and other modules.
* More importantly, the `state_channel_close` txn wouldn't fire on expiration because the state_channel signature was empty on creation, this fixes that as well.


src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -14,7 +14,8 @@
     start_link/1,
     credits/1, nonce/1,
     request/1,
-    packet/1
+    packet/1,
+    state_channels/0
```
vihu: Mostly for my debugging, but doesn't hurt to have this in here

PR #387, opened 2020-03-04
Remove nested case statements

vihu:
The extreme deeply nested cases haunt me at night, this attempts to fix those via "railway" mode

### DIFF
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -214,6 +214,23 @@ icdf_select(PopulationList, Rnd) ->
     Sum = lists:sum([Weight || {_Node, Weight} <- PopulationList]),
     icdf_select(PopulationList, normalize_float(Rnd * Sum), normalize_float(Rnd * Sum)).
 
+%%--------------------------------------------------------------------
+%% @doc
+%% This allows one to do "railway" oriented programming, instead of having
+%% deeply nested case statements, one can, pass in a list of steps, each
+%% step is processed only if the previous step outputs an {ok, NewData}.
+%%
+%% The job of this function is to process a list of "steps", where each
+%% step is a tuple like so: {tag, fun fun_name/1} and some initial data.
+%%
+%% Each function in each step operates on the initial data, and if successful
+%% is supposed to return an ok tuple, along with newdata (if it was changed).
+%%
+%% Not that the arity of the step functions is always 1.
+%%
+%% Refer to the eunit below to get a better understanding.
+%% @end
+%%-------------------------------------------------------------------
 -spec railway([{term(), fun()}], term()) -> {ok, term()} | {error, term(), term(), term()}.
 railway([], Data) ->
     {ok, Data};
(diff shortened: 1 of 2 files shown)


src/blockchain_utils.erl
```diff
@@ -213,6 +214,19 @@ icdf_select(PopulationList, Rnd) ->
     Sum = lists:sum([Weight || {_Node, Weight} <- PopulationList]),
     icdf_select(PopulationList, normalize_float(Rnd * Sum), normalize_float(Rnd * Sum)).
 
+-spec railway([{term(), fun()}], term()) -> {ok, term()} | {error, term(), term(), term()}.
+railway([], Data) ->
```
madninja: Add some documentation for this function

evanmcc:
I don't like this stuff either, but this function seems hard to use and a lot of ceremony.  maybe just move it to something like the consensus group transaction?  see https://github.com/helium/blockchain-core/blob/master/src/transactions/v1/blockchain_txn_consensus_group_v1.erl#L116

src/blockchain_utils.erl
vihu: Added here: 7f51b08

vihu:
> I don't like this stuff either, but this function seems hard to use and a lot of ceremony. maybe just move it to something like the consensus group transaction? see https://github.com/helium/blockchain-core/blob/master/src/transactions/v1/blockchain_txn_consensus_group_v1.erl#L116

lol, it's not _that_ hard https://github.com/helium/blockchain-core/pull/387/files#diff-83068e8683a5f6cb07beba4d7e433628R278. I borrowed heavily from the "railway" oriented approach, where it would short circuit if any subsequent step blows up. But yeah, if you hate this, we don't need to get it in here :)

Vagabond:
This also seems related to https://github.com/erlang/eep/blob/master/eeps/eep-0049.md

src/blockchain_utils.erl
```diff
@@ -213,6 +214,36 @@ icdf_select(PopulationList, Rnd) ->
...
+%%
+%% Refer to the eunit below to get a better understanding.
+%% @end
+%%-------------------------------------------------------------------
+-spec railway([{term(), fun()}], term()) -> {ok, term()} | {error, term(), term(), term()}.
+railway([], Data) ->
+    {ok, Data};
+railway([{Tag, Fun} | Rest], Data) ->
+    case Fun(Data) of
+        {ok, NewData} ->
```
Vagabond: Would suggest an 'ok' return that passes `Data` on to the next function without changes (but indicates the validation succeeded).

src/blockchain_utils.erl
```diff
@@ -213,6 +214,36 @@ icdf_select(PopulationList, Rnd) ->
...
+            railway(Rest, NewData);
+        {error, Reason} ->
+            {error, Tag, Reason, Data};
```
Vagabond: Why do we return the Data in the error cases?

src/blockchain_utils.erl
vihu: Added here: cddd354

src/blockchain_utils.erl
vihu: Mostly just for inspection

vihu:
This was just an experiment, closing.

PR #388, opened 2020-03-05
Macpie/state channels

macpie:


src/transactions/v1/blockchain_txn_state_channel_close_v1.erl
```diff
@@ -81,19 +81,24 @@ is_valid(Txn, Chain) ->
...
+                    Owner = blockchain_state_channel_v1:owner(SC),
+                    case blockchain_ledger_v1:find_state_channel(ID, Owner, Ledger) of
+                        {error, _Reason} ->
+                            {error, state_channel_not_open};
+                        {ok, _} ->
+                            case Owner == Closer of
+                                true ->
+                                    ok;
+                                false ->
+                                    case blockchain_state_channel_v1:balance(Closer, SC) of
```
vihu: Can we add some comments around this whole stanza, just explaining briefly what exactly is happening please?

test/blockchain_state_channel_SUITE.erl
```diff
@@ -604,6 +609,162 @@ replay_test(Config) ->
...
+    % Step 13: Adding close txn to blockchain
+    receive
+        {txn, Txn2} ->
+            B2 = ct_rpc:call(RouterNode, test_utils, create_block, [ConsensusMembers, [Txn2]]),
+            _ = ct_rpc:call(RouterNode, blockchain_gossip_handler, add_block, [RouterSwarm, B2, RouterChain, Self])
+    after 10000 ->
+        ct:fail("txn timeout")
+    end,
+
+    % Step 14: Waiting for close txn to be mine
```
vihu: We don't actually check for `blockchain_txn_state_channel_close_v1` here do we?

macpie: no but the fact it is gone from ledger mean the txn got mined

PR #389, opened 2020-03-06
Fix a case where we don't clear the blockchain_worker sync pid

Vagabond:


src/blockchain_worker.erl
```diff
@@ -529,7 +529,7 @@ handle_info({'DOWN', SyncRef, process, _SyncPid, _Reason},
             {noreply, State#state{sync_pid = undefined, sync_timer = Ref}};
         _ ->
             %% we're deep in the past here, so just start the next sync
-            {noreply, start_sync(State)}
+            {noreply, start_sync(State#state{sync_pid = undefined})}
```
evanmcc: might be safer to fix this at line 586 in start_sync?  otherwise we need to remember that the pid should be set to undefined before each call to start_sync.  but I'm OK to with this either way.

src/blockchain_worker.erl
```diff
@@ -560,8 +561,10 @@ maybe_sync(#state{blockchain = Chain} = State) ->
     erlang:cancel_timer(State#state.sync_timer),
     %% last block add time is relative to the system clock so as long as the local
     %% clock mostly increments this will eventually be true on a stuck node
+    {ok, Height} = blockchain:height(Chain),
+    SyncCooldownTime = application:get_env(blockchain, sync_cooldown_time, 300),
     case erlang:system_time(seconds) - blockchain:last_block_add_time(Chain) of
-        X when X > 300 ->
+        X when X > SyncCooldownTime orelse Height == 1  ->
```
evanmcc: maybe `Height < 100`?

PR #397, opened 2020-03-13
Move the OUI upgrade and field to master

evanmcc:
This also gets rid of neighbors, and should help tame some of the apparent drift we've been seeing.

NB: this is not something that we can roll back without a code change.  Once it's out there, we've committed to it.


vihu:
Should we add a getter for `oui` while here?

PR #398, opened 2020-03-17
add monitoring to gossip group to re-establish handlers

evanmcc:
we don't monitor the gossip group, so when it fails, we don't re-establish the gossip handler, and then reject all incoming gossip connections, and all gossiped blocks, which breaks poc.


Vagabond:
This looks good, do the stream handler not suffer from this problem?

Vagabond:
Dialyzer is mad

PR #402, opened 2020-03-20
Find state channels via a prefix fold instead of a secondary index

Vagabond:


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1807,12 +1780,13 @@ cache_get(#ledger_v1{db=DB}=Ledger, CF, Key, Options) ->
...
     end.
 
--spec cache_delete(ledger(), rocksdb:cf_handle(), any()) -> ok.
+-spec cache_delete(ledger(), rocksdb:cf_handle(), binary()) -> ok.
 cache_delete(Ledger, CF, Key) ->
     Cache = context_cache(Ledger),
     true = ets:insert(Cache, {{CF, Key}, ?CACHE_TOMBSTONE}),
     ok.
 
+-spec cache_fold(ledger(), rocksdb:cf_handle(), fun(({binary(), binary()}, any()) -> any()), any()) -> any().
```
vihu: Can we add attributes to this spec please?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1604,35 +1604,24 @@ find_state_channel(ID, Owner, Ledger) ->
...
-                          {ok, SC} -> maps:put(ID, SC, Acc)
-                      end
-              end,
-              maps:new(),
-              LedgerSCIDs
-             )
-    end.
+    SCsCF = state_channels_cf(Ledger),
+    OwnerLength = byte_size(Owner),
+    cache_fold(Ledger, SCsCF,
```
vihu: It would be nice if we can add some comments explaining the intention here?

PR #403, opened 2020-03-20
ledger snapshots

evanmcc:
Our current approach to following the chain is to download all blocks and then apply them to the ledger.  This takes approximately forever on the current v1 hardware.  snapshots are meant to provide a more concise starting point for hotspots so that they can come online more quickly.  for full nodes, the route of downloading and validating all of the blocks will still be available.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -5,18 +5,19 @@
 %%%-------------------------------------------------------------------
 -module(blockchain).
 
--export(
-   [
-    new/3, integrate_genesis/2,
+-export([
+    new/4, integrate_genesis/2,
     genesis_hash/1 ,genesis_block/1,
     head_hash/1, head_block/1,
     sync_hash/1,
     height/1,
     sync_height/1,
-    ledger/1, ledger/2, ledger_at/2,
+    ledger/0, ledger/1, ledger/2, ledger_at/2,
     dir/1,
-    blocks/1, get_block/2,
+
+    blocks/1, get_block/2, save_block/2,
     has_block/2,
+
     add_blocks/2, add_block/2, add_block/3,
     delete_block/2,
     config/2,
@@ -38,17 +39,22 @@
     fold_chain/4,
 
     reset_ledger/1, reset_ledger/2, reset_ledger/3,
+    reset_ledger_to_snap/0,
+
+    replay_blocks/4,
 
     init_assumed_valid/2,
 
-    add_gateway_txn/4, assert_loc_txn/6
-   ]).
+    add_gateway_txn/4, assert_loc_txn/6,
+
+    add_snapshot/2, get_snapshot/2, find_last_snapshot/1
+]).
 
 -include("blockchain.hrl").
 -include("blockchain_vars.hrl").
 
 -ifdef(TEST).
--export([save_block/2, bootstrap_hexes/1, can_add_block/2, get_plausible_blocks/1]).
+-export([bootstrap_hexes/1, can_add_block/2, get_plausible_blocks/1]).
 %% export a macro so we can interpose block saving to test failure
 -define(save_block(Block, Chain), ?MODULE:save_block(Block, Chain)).
 -include_lib("eunit/include/eunit.hrl").
@@ -64,6 +70,7 @@
     heights :: rocksdb:cf_handle(),
     temp_blocks :: rocksdb:cf_handle(),
     plausible_blocks :: rocksdb:cf_handle(),
+    snapshots :: rocksdb:cf_handle(),
     ledger :: blockchain_ledger_v1:ledger()
 }).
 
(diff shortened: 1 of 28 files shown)


src/blockchain.erl
```diff
@@ -395,18 +406,14 @@ sync_height(Blockchain = #blockchain{db=DB, temp_blocks=TempBlocksCF, default=De
             end
     end.
 
-%%--------------------------------------------------------------------
-%% @doc
-%% @end
-%%--------------------------------------------------------------------
+-spec ledger() -> blockchain_ledger_v1:ledger().
```
Vagabond, line 409: hah, finally

src/blockchain.erl
```diff
@@ -610,6 +616,7 @@ add_block(Block, #blockchain{db=DB, blocks=BlocksCF, heights=HeightsCF, default=
                     {ok, NextBlock} ->
                         case blockchain_block:prev_hash(NextBlock) of
                             Hash ->
+                                {ok, Batch} = rocksdb:batch(),
```
Vagabond, line 619: Maybe we don't need a batch here then?

src/blockchain.erl
```diff
@@ -747,6 +754,47 @@ add_block_(Block, Blockchain, Syncing) ->
                               true -> unvalidated_absorb_and_commit;
                               _ -> absorb_and_commit
                           end,
+                    Ledger = blockchain:ledger(Blockchain),
+                    case blockchain_block_v1:snapshot_hash(Block) of
+                        <<>> ->
+                            ok;
+                        ConsensusHash ->
```
Vagabond, line 761: Can we break this into a function?

src/blockchain.erl
```diff
@@ -747,6 +754,47 @@ add_block_(Block, Blockchain, Syncing) ->
...
+                                         || N <- lists:seq(DHeight, LedgerHeight)],
+                                        case blockchain_ledger_snapshot_v1:snapshot(Ledger, Blocks) of
+                                            {ok, Snap} ->
+                                                case blockchain_ledger_snapshot_v1:hash(Snap) of
+                                                    ConsensusHash ->
+                                                        ok = blockchain:add_snapshot(Snap, Blockchain);
+                                                    OtherHash ->
+                                                        lager:info("bad snapshot hash: ~p good ~p",
+                                                                   [OtherHash, ConsensusHash]),
+                                                        blockchain_worker:async_reset(Height)
```
Vagabond, line 786: *casually wields giant hammer*

Maybe make the log message a bit more explicit that we're doing something drastic?

Vagabond: I'm not even sure this is good behavior to do automatically, maybe just halting is better?

src/blockchain.erl
```diff
@@ -1373,6 +1443,64 @@ missing_block(#blockchain{db=DB, default=DefaultCF}) ->
             not_found
     end.
 
+-spec add_snapshot(blockchain_ledger_snapshot:snapshot(), blockchain()) ->
+                          ok | {error, any()}.
+add_snapshot(Snapshot, #blockchain{db=DB, snapshots=SnapshotsCF}) ->
```
Vagabond, line 1448: Should we GC old snapshots, especially on non-full nodes we probably shouldn't try to catch em all?

src/blockchain.erl
```diff
@@ -1394,15 +1522,39 @@ clean(Dir) when is_list(Dir) ->
...
-            Ledger = blockchain_ledger_v1:new(Dir),
+        {ok, DB, [DefaultCF, BlocksCF, HeightsCF, TempBlocksCF, PlausibleBlocksCF, SnapshotCF]} ->
+            Ledger =
+                case Mode of
+                    blessed_snapshot ->
+                        L = blockchain_ledger_v1:new(Dir),
+                        case blockchain_ledger_v1:current_height(L) of
+                            {ok, 1} ->
+                                L;
+                            {ok, CHt} when CHt < 100000000 ->
```
Vagabond, line 1541: What?

src/cli/blockchain_cli_snapshot.erl
```diff
@@ -0,0 +1,123 @@
...
+snapshot_load(["snapshot", "load", Filename], [], []) ->
+    Ret = snapshot_load(Filename),
+    [clique_status:text(io_lib:format("~p", [Ret]))];
+snapshot_load(_, _, _) ->
+    usage.
+
+snapshot_load(Filename) ->
+    {ok, BinSnap} = file:read_file(Filename),
+    %% no reason to hang this forever?
+    %% ok = blockchain_lock:acquire(15000),
```
Vagabond, line 113: Why not timeout?

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -0,0 +1,337 @@
...
+
+serialize(Snapshot, BlocksP) ->
+    Snapshot1 = case BlocksP of
+                    blocks -> Snapshot;
+                    noblocks -> Snapshot#blockchain_snapshot_v1{blocks = []}
+                end,
+    Bin = term_to_binary(Snapshot1, [{compressed, 9}]),
+    BinSz = byte_size(Bin),
+
+    %% SHA = crypto:hash(sha256, Bin),
```
Vagabond, line 161: Why are the hashes commented out?

src/cli/blockchain_cli_snapshot.erl
evanmcc: I wrote this pre-rebase to something that had the timeout and forgot to switch back to it.

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
evanmcc: I switched from hashes internal to the framing to hashes external to the framing, more like txns.  I'll remove these leftovers.

src/blockchain.erl
evanmcc: OK.  I already have halting behavior elsewhere, but perhaps we need a better way to signal that something else is wrong?  Can we affect the light at all anymore?

src/blockchain.erl
evanmcc: Yes.  We need a GC for this and for blocks.  I figured we could miss it out for v1 (we'll probably want to set it to something like a month or two anyway).

src/blockchain.erl
evanmcc: Hahah ooops.  I did that so this would compile, and then forgot to do the right actual thing.

src/blockchain.erl
evanmcc: I'm confused.   This is old behavior and the block looks required to keep the various column families in sync?

src/blockchain.erl
evanmcc: which part?

src/blockchain_worker.erl
```diff
@@ -264,11 +290,12 @@ signed_metadata_fun() ->
 %% gen_server Function Definitions
 %% ------------------------------------------------------------------
 init(Args) ->
-    erlang:process_flag(trap_exit, true),
+    %% erlang:process_flag(trap_exit, true),
```
Vagabond: Uncomment or remove?

src/blockchain.erl
```diff
@@ -774,18 +822,24 @@ add_block_(Block, Blockchain, Syncing) ->
...
-                                  end;
-                           (_, Error) -> Error
-                          end, ok, lists:seq(LedgerHeight+1, BlockchainHeight))
+            replay_blocks(Blockchain, Syncing, LedgerHeight, BlockchainHeight)
     end.
 
+replay_blocks(Chain, Syncing, LedgerHeight, ChainHeight) ->
+    lists:foldl(
+      fun(H, ok) ->
+              {ok, B} = get_block(H, Chain),
```
vihu, line 831: `get_block` may give you an `{error, _}`

src/blockchain.erl
```diff
@@ -1373,6 +1443,64 @@ missing_block(#blockchain{db=DB, default=DefaultCF}) ->
...
+    case rocksdb:get(DB, SnapshotsCF, <<Height:64/integer-unsigned-big>>, []) of
+       {ok, Hash} ->
+           ?MODULE:get_snapshot(Hash, Blockchain);
+        not_found ->
+            {error, not_found};
+        Error ->
+            Error
+    end.
+
+find_last_snapshot(Blockchain) ->
```
vihu, line 1489: Maybe add a spec to this fun since it's exported?

src/blockchain_swarm.erl
```diff
@@ -90,7 +90,6 @@ gossip_peers() ->
 %% gen_server Function Definitions
 %% ------------------------------------------------------------------
 init(Args) ->
-    erlang:process_flag(trap_exit, true),
```
vihu, line 93: Why was this removed?

src/blockchain_worker.erl
```diff
@@ -70,11 +78,13 @@
...
          sync_pid :: undefined | pid(),
          sync_paused = false :: boolean(),
+         snapshot_info :: undefined | {binary(), integer()},
          gossip_ref = make_ref() :: reference(),
          absorb_info :: undefined | {pid(), reference()},
          absorb_retries = 3 :: pos_integer(),
          resync_info :: undefined | {pid(), reference()},
-         resync_retries = 3 :: pos_integer()
+         resync_retries = 3 :: pos_integer(),
+         mode = startup :: assumed_valid | snapshot | normal | reset
```
vihu, line 87: `startup` should also be included in the type spec?

src/blockchain_worker.erl
```diff
@@ -504,11 +591,14 @@ code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
 
 terminate(_Reason, #state{blockchain=undefined}) ->
+    ct:pal("blockchain terminate ~p", [_Reason]),
```
vihu, line 594: Change these and below to `lager:warning` or something before landing?

src/blockchain_worker.erl
```diff
@@ -517,12 +607,31 @@ terminate(_Reason, #state{blockchain=Chain}) ->
...
+    maybe_sync_blocks(State);
+maybe_sync(#state{mode = snapshot, blockchain = Chain, sync_pid = Pid} = State) ->
+    Ledger = blockchain:ledger(Chain),
+    case blockchain_ledger_v1:current_height(Ledger) of
+        %% still waiting for the background process to download and
+        %% install the ledger snapshot
+        {ok, 1} when Pid /= undefined ->
+            reset_sync_timer(State);
+        {ok, 1} ->
+            {Hash, Height} = State#state.snapshot_info,
```
vihu, line 620: `State#state.snapshot_info` may return `undefined`. Maybe pattern match it out in the fun head if so and handle accordingly?

vihu: Maybe better to have getter and setters for `snapshot_info` and other newly introduced state fields?

src/blockchain_worker.erl
```diff
@@ -674,8 +844,42 @@ get_assume_valid_height_and_hash() ->
     end.
 
 load_chain(SwarmTID, BaseDir, GenDir) ->
-    AssumedValidBlockHashAndHeight = get_assume_valid_height_and_hash(),
-    case blockchain:new(BaseDir, GenDir, AssumedValidBlockHashAndHeight) of
+    QuickSyncMode = application:get_env(blockchain, quick_sync_mode, assumed_valid),
+    QuickSyncData =
+        case QuickSyncMode of
+            assumed_valid ->
```
vihu, line 850: Split the cases below into separate functions maybe?

src/blockchain_swarm.erl
evanmcc: it was hiding some crash issues I was seeing during development, so I got rid of it.  I think that it's a hindrance to debugging, and should only be used on supervising processes.

src/blockchain_worker.erl
evanmcc: no, I'll clean that up.

src/blockchain.erl
evanmcc: making it so we halt.

src/blockchain_worker.erl
evanmcc: oops, fixed.

src/blockchain_worker.erl
evanmcc: I don't really like getters and setters for internal use.   

Rather than match it out, I double checked that everywhere we go into `snapshot` we populate the `snapshot_info`.  init wasn't doing it correctly, that's fixed.

src/blockchain.erl
evanmcc: since this code path won't get hit except on manual intervention, that seems OK?

src/blockchain.erl
Vagabond: Oh, I thought we were immediately committing the batch, sorry.

src/blockchain.erl
Vagabond: the handling if we see a snapshot hash? just so this function doesn't get too deep?

src/blockchain.erl
```diff
@@ -1373,6 +1443,65 @@ missing_block(#blockchain{db=DB, default=DefaultCF}) ->
...
+    case rocksdb:get(DB, SnapshotsCF, <<Height:64/integer-unsigned-big>>, []) of
+       {ok, Hash} ->
+           ?MODULE:get_snapshot(Hash, Blockchain);
+        not_found ->
+            {error, not_found};
+        Error ->
+            Error
+    end.
+
+-spec find_last_snapshot(blockchain()) -> undefined | tuple() | {error, any()}.
```
Vagabond, line 1489:
```suggestion
-spec find_last_snapshot(blockchain()) -> undefined | {non_neg_integer(), blockchain_block:hash(), binary()} | {error, any()}.
```

src/blockchain.erl
```diff
@@ -774,18 +789,63 @@ add_block_(Block, Blockchain, Syncing) ->
...
-                                          run_absorb_block_hooks(Syncing, H, Blockchain);
-                                      {error, Reason} ->
-                                          lager:error("Error absorbing transaction, Ignoring Hash: ~p, Reason: ~p", [blockchain_block:hash_block(Block), Reason])
-                                  end;
-                           (_, Error) -> Error
-                          end, ok, lists:seq(LedgerHeight+1, BlockchainHeight))
+            replay_blocks(Blockchain, Syncing, LedgerHeight, BlockchainHeight)
+    end.
+
+check_snapshot(ConsensusHash, MyAddress, Signers,
```
Vagabond, line 795: check is sort of a misnomer here? we're generating a local copy of a snapshot from the chain?

src/blockchain_worker.erl
```diff
@@ -313,6 +355,38 @@ handle_call({new_ledger, Dir}, _From, State) ->
     Ledger1 = blockchain_ledger_v1:new(Dir),
     {reply, {ok, Ledger1}, State};
 
+handle_call({install_snapshot, Hash, Snapshot}, _From,
+            #state{blockchain = Chain, mode = Mode, swarm = Swarm} = State) ->
+    %% I don't think that we want to auto-repair right now, do default
+    %% this to disabled.
+    Halt = application:get_env(blockchain, halt_on_reset, true),
+    case Mode == reset andalso Halt of
```
Vagabond, line 363: I don't have a good feel for what 'halt on reset' means here.

Also, do we ever set mode to be reset anywhere?

evanmcc: well, this code currently stops the worker, but I'm gonna disable it shortly.  At this point all it will do is log.  Eventually it should either auto-repair or stop sync and go idle.

PR #405, opened 2020-03-24
Fix a badmatch when validating HTLC redeems

Vagabond:


madninja:
oh that was it?!

Vagabond:
Yeah, simple variable name reuse

PR #409, opened 2020-03-27
adjust gossip/sync strategy

evanmcc:
we've been having some issues with hotpots falling out of sync due to some prior bugfixing.  there is a 5 minute cooldown on starting a new sync, added when we were having some clock skew issues that were causing spots to never manage to sync.  unfortunately, it means that if a spot misses timely gossip, which happens often, it can't sync up for five minutes after the last block.  often, when the spot does sync, it then misses gossip again and has to wait another five minutes to have another attempt to get back onto the gossip train.

this patch tries, when we're not obviously clock skewed, to use the block time rather than the last-added block time.  this should lead to quicker sync cycles and make it more likely that the spot gets back into sync and stays there.

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -570,7 +570,7 @@ maybe_sync(#state{blockchain = Chain} = State) ->
     erlang:cancel_timer(State#state.sync_timer),
     %% last block add time is relative to the system clock so as long as the local
     %% clock mostly increments this will eventually be true on a stuck node
-    SyncCooldownTime = application:get_env(blockchain, sync_cooldown_time, 75),
+    SyncCooldownTime = application:get_env(blockchain, sync_cooldown_time, 60),
     SkewedSyncCooldownTime = application:get_env(blockchain, skewed_sync_cooldown_time, 300),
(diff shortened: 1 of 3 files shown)


src/blockchain_worker.erl
```diff
@@ -570,10 +570,23 @@ maybe_sync(#state{blockchain = Chain} = State) ->
     erlang:cancel_timer(State#state.sync_timer),
     %% last block add time is relative to the system clock so as long as the local
     %% clock mostly increments this will eventually be true on a stuck node
-    {ok, Height} = blockchain:height(Chain),
-    SyncCooldownTime = application:get_env(blockchain, sync_cooldown_time, 300),
-    case erlang:system_time(seconds) - blockchain:last_block_add_time(Chain) of
-        X when X > SyncCooldownTime orelse Height == 1  ->
+    SyncCooldownTime = application:get_env(blockchain, sync_cooldown_time, 75),
+    SkewedSyncCooldownTime = application:get_env(blockchain, skewed_sync_cooldown_time, 300),
```
Vagabond: Propose we lower this to 60s

PR #410, opened 2020-03-31
Handle cases when block or ledger height disagree

Vagabond:


Vagabond:
Still need review on this one @evanmcc

PR #412, opened 2020-03-31
Store blocks that look plausible and gossip them on

Vagabond:


src/blockchain.erl
```diff
@@ -1618,6 +1643,89 @@ resync_fun(ChainHeight, LedgerHeight, Blockchain) ->
...
 
+%% check if this block looks plausible
+%% if we can validate f+1 of the signatures against our consensus group
+%% that means it's probably not wildly impossible
+is_block_plausible(Block, Chain) ->
+    Ledger = ledger(Chain),
+    case blockchain_ledger_v1:consensus_members(Ledger) of
+        {error, _Reason} ->
+            false;
+        {ok, ConsensusAddrs} ->
```
evanmcc, line 1654: this adds a weird behavior contour around elections.  perhaps we could add but not gossip blocks from a different. later election epoch?

Vagabond: This is why I relaxed the signature check to only be F+1, it shouls stay valid across a couple epochs in the future...

evanmcc: oh, missed that, sounds good.

PR #415, opened 2020-04-10
don't ignore sync blocks for txn processing

evanmcc:
while it's good to ignore sync blocks on the initial sync, we can't ignore them once we're synced.  the txn manager is lightweight enough that I think that we can just do the processing while we sync.

### DIFF
--- a/src/transactions/blockchain_txn_mgr.erl
+++ b/src/transactions/blockchain_txn_mgr.erl
@@ -161,13 +161,13 @@ handle_info({blockchain_event, {new_chain, NC}}, State) ->
     {noreply, NewState};
 
 handle_info({blockchain_event, {add_block, _BlockHash, _Sync, _Ledger} = Event}, State0=#state{chain = undefined}) ->
-    lager:info("received add block event whilst no chain and sync ~p.  Initializing chain and then handling block",[_Sync]),
+    lager:info("received add block event whilst no chain.  Initializing chain and then handling block"),
     NC = blockchain_worker:blockchain(),
     State = initialize_with_chain(State0, NC),
     handle_add_block_event(Event, State#state{chain = NC});
-handle_info({blockchain_event, {add_block, _BlockHash, Sync, _Ledger} = Event}, State) ->
-    lager:debug("received add block event, sync is ~p",[Sync]),
-     handle_add_block_event(Event, State);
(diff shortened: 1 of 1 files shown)


andymck:
@evanmcc I would be concerned this change will result in the txn mgr being hammered by running the validations over and over with each sync block.  The validations can be relatively intensive and as sync add_block events will come in in quick succession arent we in danger of just typing the mgr up in knots

madninja:
> @evanmcc I would be concerned this change will result in the txn mgr being hammered by running the validations over and over with each sync block. The validations can be relatively intensive and as sync add_block events will come in in quick succession arent we in danger of just typing the mgr up in knots ?

The problem we found was that the core would start a new sync a number of times to catch up as the rest of the chain ran ahead. That sync of a small number of blocks would get missed by the txn_mgr and we wouldn't get the callbacks. Does that sound plausible?

Vagabond:
Maybe we should track if the transaction manager has seen a 'non sync' block and then since at one point it was synced, it can ignore the sync flag on sync blocks because we have some reasonable assumption that it's close to the the head of the chain?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -215,11 +214,8 @@ handle_add_block_event({add_block, BlockHash, Sync, _Ledger}, State=#state{chain
             %% If so we will only keep existing acceptions/rejections for rolled over members
             {IsNewElection, NewCGMembers} = check_block_for_new_election(Block),
             %% reprocess all txns remaining in the cache
-            ok = process_cached_txns(Chain, BlockHeight, SubmitF, Sync, IsNewElection, NewCGMembers),
-            %% only update the current block height if its not a sync block
-            NewCurBlockHeight = maybe_update_block_height(CurBlockHeight, BlockHeight, Sync),
-            lager:debug("received block height: ~p,  updated state block height: ~p", [BlockHeight, NewCurBlockHeight]),
-            {noreply, State#state{cur_block_height = NewCurBlockHeight}};
+            ok = process_cached_txns(Chain, BlockHeight, SubmitF, IsNewElection, NewCGMembers),
+            {noreply, State#state{cur_block_height = BlockHeight}};
```
evanmcc, line 218: do we ever use block height for anything?  I keep track of it in other servers for only-once processing, but I think that we may have fixed the bug that caused that.

Vagabond:
Replaced by #420

PR #418, opened 2020-04-11
dont stop dialers so eagerly

andymck:
This addresses some issues with diallers being stopped whilst handling validations in the add_block event handling of txn mgr:

1.  If the validations cannot decide on a txn ( its neither valid or invalid ) such as will be case in a txn with an out of sequence nonce, the mgr would have stopped the diallers.  If any of the corresponding CG members had not yet sent their response to the mgr, then we would not subsequently receive those responses.  

2.  If the validations declared the txn valid the mgr would then check if a new election had taken place OR if the number of active diallers is less than (F/2+1 - Number of Acceptions ):

`
case IsNewElection orelse length(Dialers) < (SubmitF - length(Acceptions)) of
    true ->
        %% the txn is valid and a new election has occurred, so keep txn in cache and resubmit
        %% keep any existing acceptions/rejections from the previous round
        ok = blockchain_txn_mgr_sup:stop_dialers(Dialers),
        RecvBlockHeight0 = normalise_block_height(CurBlockHeight, RecvBlockHeight),
        lager:info("Resubmitting txn: ~p", [blockchain_txn:hash(Txn)]), {NewAcceptions,
        NewRejections} = purge_old_cg_members(Acceptions, Rejections, NewGroupMembers),
        NewDialers = submit_txn_to_cg(Chain, Txn, SubmitF, NewAcceptions, NewRejections),
        cache_txn(Txn, Callback, RecvBlockHeight0, NewAcceptions, NewRejections, NewDialers);
    false ->
        ok
end; 
`

As you can see if either condition equals true, all existing diallers are stopped.  I think this was a dump oversight...why would we stop all existing diallers here ?  We dont know if all members have responded yet and in closing the diallers we block any possibility of the mgr receiving those responses.

The flow will however submit to an additional member for each non reecive

I have removed the stopping of the diallers here.

Impact on resubmits
--------------------
In both scenarios above, by stopping the diallers, if any of the corresponding members had not sents its response to the mgr as yet we block the potential for those responses to be received.  

In scenario 1, this will result in out of sequences txns never  getting resubmitted ( until it passes validations ).  We dont resubmit as part of the add_block event handling nor can we do it upon a yet to be received reject response from the CG member ( as that response can no longer be received ).  The txn is now in mgr limbo until it passes validations.

In scenario 2, this one is more non deterministic as to whether a resubmit will take place or not but for a resubmit to occur we either have to be in a new election or the number of active diallers is less than ((F/2+1) - Number of Acceptions)).    In a regular flow where a txn passes validations but is not yet selected by the CG for a block, this will be infrequent but once we do hit it any existing dialler will be stopped and thus any potential for receiving a response from those members is blocked.  This is somewhat counted by the fact that we will submit to a new selection of members.  So the direct observable impact here is somewhat grey/vague but i do believe that stopping the diallers is unwarranted and will be having negative impact.

The impact of this PR will have most observable impact on txns being received/processed out of sequence.

### DIFF
--- a/src/transactions/blockchain_txn_mgr.erl
+++ b/src/transactions/blockchain_txn_mgr.erl
@@ -167,7 +167,7 @@ handle_info({blockchain_event, {add_block, _BlockHash, _Sync, _Ledger} = Event},
     handle_add_block_event(Event, State#state{chain = NC});
 handle_info({blockchain_event, {add_block, _BlockHash, Sync, _Ledger} = Event}, State) ->
     lager:debug("received add block event, sync is ~p",[Sync]),
-     handle_add_block_event(Event, State);
+    handle_add_block_event(Event, State);
 
 handle_info(_Msg, State) ->
     lager:warning("blockchain_txn_mgr got unknown info msg: ~p", [_Msg]),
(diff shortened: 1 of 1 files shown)


Vagabond:
Replaced by #420

PR #419, opened 2020-04-11
Track active transaction manager dialers more consistently

Vagabond:
The transaction manager would often lose track of the active dialers for
a transaction if a dial failed or an election happened. This is
potentially problematic in that we'd lose responses from the other
active dialers possibly leading to an orphaned transaction.

This change tracks the dialer pids associated with the member they're
dialing. When a single dialer fails, only that dialer is removed. When
an election occurs, only the dialers for deposed members are removed and
when dialing new members, the members with active dialers are excluded
from being selected as a member to dial.

### DIFF
--- a/src/transactions/blockchain_txn_mgr.erl
+++ b/src/transactions/blockchain_txn_mgr.erl
@@ -39,11 +39,15 @@
           chain :: undefined | blockchain:blockchain()
          }).
 
+
+-type dialer() :: {pid(), libp2p_crypto:pubkey_bin()}.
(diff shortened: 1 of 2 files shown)


andymck:
I actually starting doing this very thing yesterday but had to cut it out as would not have time to see it through.  I see you have a bunch of other commits in the adt/txn-mgr-integration branch.  prob best to review that when you are done with it as this looks stale now....

Vagabond:
Replaced by #420

PR #420, opened 2020-04-13
Various attempts to fix and improve the txn manager

Vagabond:
This is a draft because this is a grab-bag of changes I made while trying to find and fix bugs in the txn manager. Some of the changes don't make sense to merge, but to aid discussion I'm going to open this draft PR. I will attempt to explain the rationale for some of my changes, and flag ones that should probably not merge in comments on the code.

### DIFF
--- a/src/transactions/blockchain_txn_dialer.erl
+++ b/src/transactions/blockchain_txn_dialer.erl
@@ -32,7 +32,8 @@
 -record(state, {
           parent :: pid(),
           txn :: blockchain_txn:txn(),
-          member :: libp2p_crypto:pubkey_bin()
+          member :: libp2p_crypto:pubkey_bin(),
+          timeout = make_ref() :: reference()
          }).
 
 %% ------------------------------------------------------------------
@@ -50,12 +51,13 @@ dial(Pid) ->
 init(Args) ->
     lager:debug("blockchain_txn_dialer started with ~p", [Args]),
     [Parent, Txn, Member] = Args,
-    {ok, #state{parent=Parent, txn=Txn, member=Member}}.
+    Ref = erlang:send_after(30000, Parent, {timeout, {self(), Txn, Member}}),
+    {ok, #state{parent=Parent, txn=Txn, member=Member, timeout=Ref}}.
 
 handle_call(_, _, State) ->
     {reply, ok, State}.
 
-handle_cast(dial, State=#state{member=Member, txn=Txn, parent=Parent}) ->
+handle_cast(dial, State=#state{member=Member, txn=Txn, parent=Parent, timeout=Ref}) ->
     Swarm = blockchain_swarm:swarm(),
     P2PAddress = libp2p_crypto:pubkey_bin_to_p2p(Member),
     TxnHash = blockchain_txn:hash(Txn),
@@ -65,6 +67,7 @@ handle_cast(dial, State=#state{member=Member, txn=Txn, parent=Parent}) ->
                                          blockchain_txn_handler,
                                          [self(), blockchain_txn:hash(Txn)]) of
         {error, Reason} ->
+            erlang:cancel_timer(Ref),
             lager:error("libp2p_framed_stream dial failed. Reason: ~p, To: ~p, TxnHash: ~p",
                         [Reason, P2PAddress, TxnHash]),
             Parent ! {dial_failed, {self(), Txn, Member}},
@@ -73,6 +76,7 @@ handle_cast(dial, State=#state{member=Member, txn=Txn, parent=Parent}) ->
             DataToSend = blockchain_txn:serialize(Txn),
             case libp2p_framed_stream:send(Stream, DataToSend) of
                 {error, Reason} ->
+                    erlang:cancel_timer(Ref),
                     lager:error("libp2p_framed_stream send failed. Reason: ~p, To: ~p, TxnHash: ~p",
                                 [Reason, P2PAddress, TxnHash]),
                     Parent ! {send_failed, {self(), Txn, Member}},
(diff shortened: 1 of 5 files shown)


src/transactions/blockchain_txn_mgr.erl
```diff
@@ -274,25 +298,24 @@ signatory_rand_members(Chain, SubmitF, Acceptions, Rejections) ->
             %% so use a random consensus member
```
evanmcc, line 298: imo this case should just make the group, and move the processing below it, as it looks like somehow we don't shuffle on the signatory side (the common case).

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -422,7 +465,7 @@ delete_cached_txn(Txn) ->
 -spec cached_txn(blockchain_txn:txn())-> {ok, cached_txn_type()} | {error, txn_not_found}.
 cached_txn(Txn)->
     case ets:lookup(?TXN_CACHE, Txn) of
-        [Txn] -> {ok, Txn};
+        [Res] -> {ok, Res};
```
Vagabond, line 468: This was the core bug and it was breaking the removing of dialers. The only way the txn could clear was to appear in a block, a rejection or a failure would never allow the txn to be offered to other members.

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -110,50 +120,64 @@ handle_cast({submit, Txn, Callback}, State=#state{chain = undefined}) ->
...
 
-handle_info({dial_failed, {Dialer, Txn, Member}}, #state{chain = Chain, submit_f = SubmitF} = State) ->
+handle_info({dial_failed, {Dialer, Txn, Member}}, State) ->
     lager:debug("txn: ~s, dial_failed: ~p, Dialer: ~p", [blockchain_txn:print(Txn), Member, Dialer]),
-    ok = retry(Txn, Chain, SubmitF),
+    ok = retry(Txn, Dialer),
+    {noreply, State};
+
+handle_info({timeout, {Dialer, Txn, Member}}, State) ->
+    lager:debug("txn: ~s, timeout: ~p, Dialer: ~p", [blockchain_txn:print(Txn), Member, Dialer]),
```
evanmcc, line 163: Should we kill the dialer here?  we clean up the state in retry, but don't seem to do that.  what happens when we remove the dialer but then get a different message from it (e.g. the timeout and a success both go into the message queue for this process before we manage to process either)?  Are we idempotent both ways?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -307,34 +330,51 @@ process_cached_txns(Chain, CurBlockHeight, SubmitF, _Sync, IsNewElection, NewGro
     {ValidTransactions, InvalidTransactions} = blockchain_txn:validate(Txns, Chain),
     ok = lists:foreach(
         fun({Txn, {Callback, RecvBlockHeight, Acceptions, Rejections, Dialers}}) ->
-            case {lists:member(Txn, InvalidTransactions), lists:member(Txn, ValidTransactions)} of
-                {false, false} ->
+            IsValidStandalone = case blockchain_txn:is_valid(Txn, Chain) of
```
Vagabond, line 333: This relies on my changes to check nonces in is_valid and probably should not go in. I have a related branch of core that has a function to determine dependencies between transactions which we could probably use here instead.

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -274,25 +298,24 @@ signatory_rand_members(Chain, SubmitF, Acceptions, Rejections) ->
...
-retry(Txn, Chain, SubmitF) ->
+-spec retry(blockchain_txn:txn(), pid()) -> ok.
+retry(Txn, Dialer) ->
     case cached_txn(Txn) of
         {error, _} ->
             %% We no longer have this txn, do nothing
             ok;
-        {ok, {Txn, {Callback, RecvBlockHeight, Acceptions, Rejections, _Dialers}}} ->
-            %% resubmit the txn to CG
-            NewDialers = submit_txn_to_cg(Chain, Txn, SubmitF, Acceptions, Rejections),
```
Vagabond, line 294: I removed all the submit_txn_to_cg calls except for the one called from process_cached_txns to simplifiy things while debugging. Additionally process_cached_txns had a bunch of logic not replicated on the other call paths, so we should consider that before blindly restoring the other submit calls.

src/transactions/blockchain_txn_mgr.erl
andymck: fecking bassdrinker hiding in plain sight

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -189,14 +213,12 @@ code_change(_OldVsn, State, _Extra) ->
...
-    SubmitF = submit_f(length(Signatories)),
-    RejectF = reject_f(length(Signatories)),
+    {ok, N} = blockchain:config(?num_consensus_members, blockchain:ledger(Chain)),
+    SubmitF = submit_f(N),
+    RejectF = reject_f(N),
     %% process any cached txn from before we had a chain, none of these will have been submitted as yet
-    F = fun({Txn, {Callback, _RecvBlockHeight, _Acceptions, _Rejections, _Dialers}}) ->
-            NewDialers = submit_txn_to_cg(Chain, Txn, SubmitF, [], []),
-            ok = cache_txn(Txn, Callback, Height, [], [], NewDialers)
+    F = fun({Txn, {Callback, _RecvBlockHeight, Acceptions, Rejections, Dialers}}) ->
```
andymck, line 220: If we end up not submitting the txns from here in the final changes and only doing so from the add_block event then we can remove this fun and its execution completely

src/transactions/blockchain_txn_dialer.erl
```diff
@@ -50,12 +51,13 @@ dial(Pid) ->
 init(Args) ->
     lager:debug("blockchain_txn_dialer started with ~p", [Args]),
     [Parent, Txn, Member] = Args,
-    {ok, #state{parent=Parent, txn=Txn, member=Member}}.
+    Ref = erlang:send_after(30000, Parent, {timeout, {self(), Txn, Member}}),
```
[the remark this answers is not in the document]
Vagabond: This is if the dialer is stuck somehow (libp2p dials might hang?). Sending the timeout directly to the parent means that any weird libp2p failure (or a process crash) cannot interfere with the delivery of this message. The only way this message gets stopped is by the dialer cancelling the timer.

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -206,6 +228,7 @@ initialize_with_chain(State, Chain)->
 handle_add_block_event({add_block, BlockHash, Sync, _Ledger}, State=#state{chain = Chain,
                                                                            cur_block_height = CurBlockHeight})->
     #state{submit_f = SubmitF, chain = Chain} = State,
+    HasBeenSynced = Sync == false orelse State#state.has_been_synced,
```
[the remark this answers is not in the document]
Vagabond: if you've managed to sync up at least once, the rationale is you're probably close enough to the head that your transactions are reasonably valid

src/transactions/blockchain_txn_mgr.erl
Vagabond: Is this still relevant?

src/transactions/blockchain_txn_mgr.erl
Vagabond: Yes, we should stop the dialer here. Yes, there's probably a race condition here.

andymck: i have added the stop as part of my branch to use records

src/transactions/blockchain_txn_mgr.erl
andymck: I commented this fun out for now as it serves no purpose.  We should uncomment if we readd submitting of txns as part of the submit api handling

PR #421, opened 2020-04-14
Fix zero amount bug in transations

vihu:
- Adds var, `allow_zero_amount`, must be set to false to trigger new checks.
- Update tests for the same

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -253,3 +253,5 @@
 -define(max_payments, max_payments).
 %% Var to switch off legacy payment txn
 -define(deprecate_payment_v1, deprecate_payment_v1).
+%% Var to allow zero amount pyament_v2 txns, set to false
+-define(allow_zero_amount, allow_zero_amount).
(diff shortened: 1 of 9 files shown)


include/blockchain_vars.hrl
```diff
@@ -253,3 +253,5 @@
 -define(max_payments, max_payments).
 %% Var to switch off legacy payment txn
 -define(deprecate_payment_v1, deprecate_payment_v1).
+%% Var to allow zero amount pyament_v2 txns, set to false
```
Vagabond, line 256:
```suggestion
%% Var to allow zero amount payment_v2 txns, set to false
```

src/transactions/v2/blockchain_payment_v2.erl
```diff
@@ -25,7 +25,7 @@
 
 -spec new(Payee :: libp2p_crypto:pubkey_bin(),
           Amount :: non_neg_integer()) -> payment().
-new(Payee, Amount) when Amount > 0 ->
```
vihu, line 28: This is a client side check and pointless

PR #422, opened 2020-04-15
Properly sequence dependent transactions in the transaction manager

Vagabond:
When delivering a transaction to the consensus group, check if any other
transactions current in-progress are dependencies. If any are found,
submit the transaction only to members which have accepted all the
dependencies. This ensures that the dependent transactions appear in the
same buffer as their dependencies or they wait until all the
dependencies have cleared.

### DIFF
--- a/src/transactions/blockchain_txn.erl
+++ b/src/transactions/blockchain_txn.erl
@@ -66,7 +66,8 @@
     wrap_txn/1,
     unwrap_txn/1,
     is_valid/2,
-    validate_fields/1
+    validate_fields/1,
+    depends_on/2
 ]).
 
 -ifdef(TEST).
@@ -616,6 +617,8 @@ nonce(Txn) ->
             blockchain_txn_payment_v1:nonce(Txn);
         blockchain_txn_security_exchange_v1 ->
             blockchain_txn_security_exchange_v1:nonce(Txn);
+        blockchain_txn_token_burn_v1 ->
+            blockchain_txn_token_burn_v1:nonce(Txn);
         blockchain_txn_payment_v2 ->
             blockchain_txn_payment_v2:nonce(Txn);
         _ ->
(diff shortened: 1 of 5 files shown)


src/transactions/blockchain_txn_mgr.erl
```diff
@@ -372,9 +362,26 @@ process_cached_txns(Chain, CurBlockHeight, SubmitF, _Sync, IsNewElection, NewGro
...
+                                    cache_txn(Txn, Callback, RecvBlockHeight0, NewAcceptions, NewRejections, RemainingDialers ++ NewDialers);
+                                Dependencies ->
+                                    {ok, {_, {_, _, A0, _, _}}} = cached_txn(hd(Dependencies)),
+                                    ElegibleMembers = sets:to_list(lists:foldl(fun(E, Acc) ->
+                                                                                       {ok, {_, {_, _, A, _, _}}} = cached_txn(E),
+                                                                                       sets:intersection(Acc, sets:from_list(A))
+                                                                               end, sets:from_list(A0), tl(Dependencies))),
+                                    {_, DialedMembers} = lists:unzip(Dialers),
+                                    NewDialers = dial_members(lists:sublist(((ElegibleMembers -- NewAcceptions) -- NewRejections) -- DialedMembers, SubmitF - length(NewAcceptions) - length(Dialers)), Chain, Txn),
```
andymck, line 381: this is hurting my brain trying to work out what we end up with.  Maybe break it out over a couple of steps?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -359,7 +349,7 @@ process_cached_txns(Chain, CurBlockHeight, SubmitF, _Sync, IsNewElection, NewGro
                     ok = blockchain_txn_mgr_sup:stop_dialers(Dialers),
                     ok = invoke_callback(Callback, {error, invalid}),
                     delete_cached_txn(Txn);
-                {_, true, true} ->
+                {_, true} ->
                     case IsNewElection orelse length(Dialers) < (SubmitF - length(Acceptions)) of
```
andymck, line 353: not related to this PR directly but this orelse bothers me....i think it would be more logical to have the IsNewElection part of the parent case statement ( ie case {lists:member(Txn, InvalidTransactions), lists:member(Txn, ValidTransactions), IsNewElection} of )

Seems we mix two scenarios up but yet the handling of it implies its dealing with a new election irrespective of which condition is true.

I dont mind having a go at refactoring this....

src/transactions/blockchain_txn.erl
```diff
@@ -663,9 +666,75 @@ actor(Txn) ->
             <<>>
     end.
 
+depends_on(Txn, Txns) ->
```
andymck, line 669: can we get the specs added here

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -372,9 +362,26 @@ process_cached_txns(Chain, CurBlockHeight, SubmitF, _Sync, IsNewElection, NewGro
...
+                                    %% XXX we assume we have correct dependency resolution here, if you add a new transaction with 
+                                    %% dependencies and don't fix depends_on, your transaction will probably get rejected
+                                    NewDialers = submit_txn_to_cg(Chain, Txn, SubmitF, NewAcceptions, NewRejections, RemainingDialers),
+                                    lager:info("Resubmitting txn: ~p to ~b new dialers after election", [blockchain_txn:hash(Txn), length(NewDialers)]),
+                                    cache_txn(Txn, Callback, RecvBlockHeight0, NewAcceptions, NewRejections, RemainingDialers ++ NewDialers);
+                                Dependencies ->
+                                    {ok, {_, {_, _, A0, _, _}}} = cached_txn(hd(Dependencies)),
+                                    ElegibleMembers = sets:to_list(lists:foldl(fun(E, Acc) ->
+                                                                                       {ok, {_, {_, _, A, _, _}}} = cached_txn(E),
+                                                                                       sets:intersection(Acc, sets:from_list(A))
```
andymck, line 378: Difficult to understand what intent is with ElegibleMembers here, especially over multiple blocks.  Some comments would  def help here.

But it looks to me like ElegibleMembers will be limited to atmost the set of CG members which have accepted the head dep txn.  What if there are multiple dep txns and each have a somewhat different set of accepted members ?  We will only ever be resubmitting the txn to a subset of those

src/transactions/v1/blockchain_txn_payment_v1.erl
```diff
@@ -158,16 +157,7 @@ is_valid(Txn, Chain) ->
                                                         {error, _}=Error1 ->
                                                             Error1;
                                                         ok ->
-                                                            case blockchain_ledger_v1:check_balance(Payer, Amount, Ledger) of
```
andymck, line 161: how come we dont need nonce checks now ? are we assuming with the depends_on check we wont get txns coming in out of order ?  how will this impact validations and absorb checks ?  anything we need to consider there?

Vagabond: We didn't have them before, I added them in the base branch before I had depends_on working, now we can remove them again.

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -372,9 +362,26 @@ process_cached_txns(Chain, CurBlockHeight, SubmitF, _Sync, IsNewElection, NewGro
...
+                                [] ->
+                                    %% XXX we assume we have correct dependency resolution here, if you add a new transaction with 
+                                    %% dependencies and don't fix depends_on, your transaction will probably get rejected
+                                    NewDialers = submit_txn_to_cg(Chain, Txn, SubmitF, NewAcceptions, NewRejections, RemainingDialers),
+                                    lager:info("Resubmitting txn: ~p to ~b new dialers after election", [blockchain_txn:hash(Txn), length(NewDialers)]),
+                                    cache_txn(Txn, Callback, RecvBlockHeight0, NewAcceptions, NewRejections, RemainingDialers ++ NewDialers);
+                                Dependencies ->
+                                    {ok, {_, {_, _, A0, _, _}}} = cached_txn(hd(Dependencies)),
+                                    ElegibleMembers = sets:to_list(lists:foldl(fun(E, Acc) ->
+                                                                                       {ok, {_, {_, _, A, _, _}}} = cached_txn(E),
```
evanmcc, line 377: can we make it so that the cache is using a module-only record?  I find that's usually a decent cleanup when we have this many fields.

src/transactions/blockchain_txn_mgr.erl
evanmcc: agree, this could be clarified a bit

src/transactions/blockchain_txn_mgr.erl
Vagabond: Yeah, this would be a good improvement

src/transactions/blockchain_txn.erl
```diff
@@ -663,9 +666,75 @@ actor(Txn) ->
...
+    end.
+
+
 %% ------------------------------------------------------------------
 %% EUNIT Tests
 %% ------------------------------------------------------------------
 -ifdef(TEST).
 
+
+depends_on_test() ->
```
evanmcc, line 716: this screams out for eqc

src/transactions/blockchain_txn_mgr.erl
Vagabond: If the member didn't have the first dep txn, it won't accept subsequent ones. This essentially forces us to submit only to members which have all the dep txns.

As transactions clear, the dep list shrinks which allows us more freedom in who to submit to, so eventually we'd be able to submit to any of the members once the dep list is empty.

src/transactions/blockchain_txn_mgr.erl
Vagabond: Right, the goal was to dial any elegible members who had not already accepted or rejected this txn and who we were not already dialing, up to the SubmitF threshold (less any active dialers).

PR #426, opened 2020-04-16
Consolidate state channel chain vars

vihu:


include/blockchain_vars.hrl
```diff
@@ -255,3 +255,22 @@
 -define(deprecate_payment_v1, deprecate_payment_v1).
 %% Set this var to false to disable zero amount txns (payment_v1, payment_v2, htlc_create)
 -define(allow_zero_amount, allow_zero_amount).
+
+
+%% ------------------------------------------------------------------
+%% State channel related vars
+%%
+%% Approximate blocks in a week, set to 10080
+-define(approx_blocks_in_week, approx_blocks_in_week).
```
Vagabond, line 264: Surely we can calculate this rom the other variables

include/blockchain_vars.hrl
```diff
@@ -255,3 +255,22 @@
...
+%% Max open state channels per router, set to 2
+-define(max_open_sc, max_open_sc).
+%% Max xor filter size, set to 1024*100
+-define(max_xor_filter_size, max_xor_filter_size).
+%% Max number of xor filters, set to 5
+-define(max_xor_filter_num, max_xor_filter_num).
+%% Max subnet size, 65536
+-define(max_subnet_size, max_subnet_size).
+%% Min subnet size, 8
+-define(min_subnet_size, min_subnet_size).
```
Vagabond, line 276: I think we also need one for the max number of subnets you can have

vihu: Was thinking we should just do `subnet_range` and `max_subnet_num`. And then we can potentially remove:

```
{request_subnet, SubnetSize} when SubnetSize < MinSubnetSize orelse SubnetSize > MaxSubnetSize ->
            {error, invalid_subnet_size};
```

PR #429, opened 2020-04-19
Optimistically re-gossip blocks if they look plausible

Vagabond:
This tries to simplify the block re-gossip mechanism and doesn't put it
behind absorbing the block. If the block passes the is_plausible checks
(height and signatures), immediately re-gossip it.

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -478,21 +478,8 @@ handle_cast(_Msg, State) ->
 
 handle_info(maybe_sync, State) ->
     {noreply, maybe_sync(State)};
-handle_info({blockchain_event, {add_block, Hash, Sync, _Ledger}}, #state{swarm=Swarm, blockchain=Chain} = State) ->
-    %% we added a new block to the chain, send it to all our peers
-    case blockchain:get_block(Hash, Chain) of
-        {ok, Block} when Sync == false ->
-            libp2p_group_gossip:send(
-              libp2p_swarm:gossip_group(Swarm),
-              ?GOSSIP_PROTOCOL,
-              blockchain_gossip_handler:gossip_data(Swarm, Block)
-             );
-        {ok, _} ->
-            %% don't gossip blocks we're syncing
-            ok;
-        {error, Reason} ->
-            lager:warning("unable to find new block ~p in blockchain: ~p", [Hash, Reason])
-    end,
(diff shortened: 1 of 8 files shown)


src/blockchain_worker.erl
```diff
@@ -478,21 +478,8 @@ handle_cast(_Msg, State) ->
 
 handle_info(maybe_sync, State) ->
     {noreply, maybe_sync(State)};
-handle_info({blockchain_event, {add_block, Hash, Sync, _Ledger}}, #state{swarm=Swarm, blockchain=Chain} = State) ->
```
evanmcc, line 481: if we're not using this anymore should we remove this clause and not register?

src/blockchain_worker.erl
```diff
@@ -478,21 +478,8 @@ handle_cast(_Msg, State) ->
...
-              blockchain_gossip_handler:gossip_data(Swarm, Block)
-             );
-        {ok, _} ->
-            %% don't gossip blocks we're syncing
-            ok;
-        {error, Reason} ->
-            lager:warning("unable to find new block ~p in blockchain: ~p", [Hash, Reason])
-    end,
+handle_info({blockchain_event, {add_block, _Hash, _Sync, _Ledger}}, State) ->
+    %% nothing to do here, block re-gossip is handled by the gossip handler
```
evanmcc, line 482: we claim it's handled in the handler

src/handlers/blockchain_gossip_handler.erl
```diff
@@ -30,40 +30,42 @@ init_gossip_data(WAT) ->
...
     lager:debug("Sender: ~p, MyAddress: ~p", [Sender, blockchain_swarm:pubkey_bin()]),
     case blockchain:add_block(Block, Chain) of
         ok ->
             lager:info("got gossipped block ~p", [blockchain_block:height(Block)]),
-            %% if we added it properly the add_block event will trigger gossip
             ok;
         plausible ->
             lager:warning("plausuble gossipped block doesn't fit with our chain, will start sync if not already active"),
             blockchain_worker:maybe_sync(),
-            libp2p_group_gossip:send(
```
evanmcc, line 62: but we remove that handling here?

src/blockchain_worker.erl
Vagabond: We use it for init_chain

PR #431, opened 2020-04-20
Use records for cached txn data in txn mgr

andymck:
Also includes a change where txn_list API returns a map with the Txn as the key and the cached data as the value

Associated miner PR: https://github.com/helium/miner/pull/372

### DIFF
--- a/src/transactions/blockchain_txn_mgr.erl
+++ b/src/transactions/blockchain_txn_mgr.erl
@@ -42,16 +42,18 @@
           has_been_synced= false :: boolean()
          }).
 
+-record(txn_data,
+        {
+            callback :: fun(),
+            recv_block_height=undefined :: undefined | integer(),
+            acceptions=[] :: [libp2p_crypto:pubkey_bin()],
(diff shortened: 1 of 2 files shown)


src/transactions/blockchain_txn_mgr.erl
```diff
@@ -201,7 +211,7 @@ handle_info(_Msg, State) ->
 terminate(_Reason, _State) ->
     lager:debug("terminating with reason ~p", [_Reason]),
     %% stop dialers of cached txns
-    [blockchain_txn_mgr_sup:stop_dialers(Dialers) || {_Txn, {_, _, _, _, Dialers}} <- cached_txns()],
+    catch [blockchain_txn_mgr_sup:stop_dialers(TxnData#txn_data.dialers) || {_Txn, TxnData} <- cached_txns()],
```
Vagabond, line 214: I think stop dialers catches now?

andymck: resolved

PR #433, opened 2020-04-20
Add depends_on for state_channel related txns

vihu:
Add depends_on verification for sc and related txns

### DIFF
--- a/src/transactions/blockchain_txn.erl
+++ b/src/transactions/blockchain_txn.erl
@@ -232,8 +232,7 @@ validate([Txn | Tail] = Txns, Valid, Invalid, PType, PBuf, Chain) ->
                             maybe_log_duration(type(Txn), Start),
                             validate(Tail, [Txn|Valid], Invalid, PType, PBuf, Chain);
                         {error, _Reason} ->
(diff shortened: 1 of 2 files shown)


src/transactions/blockchain_txn.erl
```diff
@@ -722,10 +722,35 @@ depends_on(Txn, Txns) ->
...
-        %blockchain_txn_routing_v1 ->
-        %    Actor = actor(Txn),
-        %    Nonce = blockchain_txn_routing_v1:nonce(Txn),
+        blockchain_txn_state_channel_open_v1 ->
+            Actor = blockchain_txn_state_channel_open_v1:actor(Txn),
+            Nonce = blockchain_txn_state_channel_open_v1:nonce(Txn),
+            OUI = blockchain_txn_state_channel_open_v1:oui(Txn),
+            lists:filter(fun(E) ->
+                                 (type(E) == blockchain_txn_state_channel_open_v1 andalso actor(E) == Actor andalso nonce(E) < Nonce) orelse
+                                 (type(E) == blockchain_txn_oui_v1 andalso blockchain_txn_oui_v1:owner(E) == Actor) orelse
```
Vagabond, line 731: Owner is unlikely to be the same here, it would be more fruitful to check for the router address?

PR #436, opened 2020-04-26
Return fast if we already have the gossipped block

Vagabond:
Instead of acquiring the blockchain lock and doing other expensive
thinga, and potentially regossiping a block several time, short circuit
out of the block gossip handler if we already have the block stored in
the blockchain.

### DIFF
--- a/src/blockchain_lock.erl
+++ b/src/blockchain_lock.erl
@@ -5,24 +5,38 @@
 
 -behaviour(gen_server).
 
--export([acquire/0, release/0, force_release/0, check/0]).
+-export([acquire/0, acquire/1, release/0, force_release/0, check/0]).
 
 -export([start_link/0, init/1, handle_call/3, handle_cast/2, handle_info/2]).
 
 -define(holding_lock, holding_lock).
 -define(lock_ref, lock_ref).
 
+-spec acquire() -> ok | error.
 acquire() ->
+    acquire(infinity).
+
+-spec acquire(Timeout:: infinity | pos_integer()) -> ok | error.
+acquire(Timeout) ->
     case get(?holding_lock) of
         undefined ->
             Ref = make_ref(),
-            put(?lock_ref, Ref),
-            put(?holding_lock, 1),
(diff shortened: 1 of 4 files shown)


src/handlers/blockchain_gossip_handler.erl
```diff
@@ -39,19 +39,26 @@ handle_gossip_data(_StreamPid, Data, [Swarm, Blockchain]) ->
             undefined ->
                 lager:notice("gossip_handler unknown block: ~p", [Block]);
             _ ->
-                case blockchain:is_block_plausible(Block, Blockchain) of
+                case blockchain:has_block(Block, Blockchain) of
```
vihu, line 42: Do we need to wrap this in a try/catch? `blockchain:has_block` may return you an error as well...

Vagabond: Made the fallback clause _

vihu:
Changes here appear to break `plausible_block_SUITE`, just ran the same on core master and it appears to pass there @Vagabond

evanmcc:
I like this, but I wonder if we couldn't add the block height to the dial info, so the remote server would just close the connection rather than accepting the block, so save on bandwidth.  I suppose concurrency means that we need to do both.

src/blockchain_lock.erl
```diff
@@ -5,24 +5,31 @@
...
+
+-spec acquire(Timeout:: infinity | pos_integer()) -> ok.
+acquire(Timeout) ->
     case get(?holding_lock) of
         undefined ->
             Ref = make_ref(),
             put(?lock_ref, Ref),
             put(?holding_lock, 1),
-            gen_server:call(?MODULE, {acquire, Ref}, infinity);
+            gen_server:call(?MODULE, {acquire, Ref}, Timeout);
```
evanmcc, line 26: I think that a safer implementation of this would:
1. catch the timeout
2. cleanup the pdict
3. send a release to the lock
4. return an error tuple to the caller

PR #438, opened 2020-04-28
Dont crash if txn deps don't resolve

vihu:
Noticed the txn_mgr crashing when the txn dependencies don't resolve, we should handle that.

```2020-04-28 15:44:37.109 [info] <0.879.0>@blockchain_txn_mgr:process_cached_txns:361 Invalidated txn: <<213,21,176,194,145,130,26,51,196,57,250,216,149,53,118,99,59,76,101,94,239,59,236,190,189,88,92,5,66,205,85,147>>
2020-04-28 15:44:37.109 [warning] <0.352.0>@blockchain_event:terminate:108 terminating remove_handler
2020-04-28 15:44:37.109 [error] <0.879.0>@blockchain_txn_mgr:check_for_deps_and_resubmit:419 gen_server blockchain_txn_mgr terminated with reason: no match of right hand value {error,txn_not_found} in blockchain_txn_mgr:check_for_deps_and_resubmit/5 line 419
2020-04-28 15:44:37.109 [error] <0.879.0>@blockchain_txn_mgr:check_for_deps_and_resubmit:419 CRASH REPORT Process blockchain_txn_mgr with 0 neighbours crashed with reason: no match of right hand value {error,txn_not_found} in blockchain_txn_mgr:check_for_deps_and_resubmit/5 line 419
```

### DIFF
--- a/src/transactions/blockchain_txn_mgr.erl
+++ b/src/transactions/blockchain_txn_mgr.erl
@@ -406,7 +406,7 @@ check_for_deps_and_resubmit(Txn, Txns, Chain, SubmitF, #txn_data{ acceptions = A
                                                                   dialers = Dialers} = TxnData)->
     %% check if this transaction has any dependencies
     %% figure out what, if anything, this transaction depends on
-    case lists:dropwhile(fun(E) -> cached_txn(E) == {error, txn_not_found} end, blockchain_txn:depends_on(Txn, Txns)) of
+    case lists:filter(fun(E) -> cached_txn(E) /= {error, txn_not_found} end, blockchain_txn:depends_on(Txn, Txns)) of
         [] ->
             %% NOTE: we assume we have correct dependency resolution here, if you add a new transaction with
             %% dependencies and don't fix depends_on, your transaction will probably get rejected
--- a/src/transactions/blockchain_txn_mgr.erl
+++ b/src/transactions/blockchain_txn_mgr.erl
@@ -406,7 +406,7 @@ check_for_deps_and_resubmit(Txn, Txns, Chain, SubmitF, #txn_data{ acceptions = A
                                                                   dialers = Dialers} = TxnData)->
     %% check if this transaction has any dependencies
     %% figure out what, if anything, this transaction depends on
-    case blockchain_txn:depends_on(Txn, Txns) of
+    case lists:dropwhile(fun(E) -> cached_txn(E) == {error, txn_not_found} end, blockchain_txn:depends_on(Txn, Txns)) of
         [] ->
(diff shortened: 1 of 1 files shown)


src/transactions/blockchain_txn_mgr.erl
```diff
@@ -416,20 +416,28 @@ check_for_deps_and_resubmit(Txn, Txns, Chain, SubmitF, #txn_data{ acceptions = A
...
-            %% those which we are already dialling
-            ElegibleMembers1 = ((ElegibleMembers -- Acceptions) -- Rejections) -- ExistingDialers,
-            %% determine max number of new diallers we need to start and then use this to get our target list to dial
-            MaxNewDiallersCount = SubmitF - length(Acceptions) - length(Dialers),
-            NewDialers = dial_members(lists:sublist(ElegibleMembers1, MaxNewDiallersCount), Chain, Txn),
-            lager:debug("txn ~p depends on ~p other txns, can dial ~p members and dialed ~p", [blockchain_txn:hash(Txn), length(Dependencies), length(ElegibleMembers), length(NewDialers)]),
-            cache_txn(Txn, TxnData#txn_data{dialers =  Dialers ++ NewDialers})
+            case cached_txn(hd(Dependencies)) of
+                {error, txn_not_found} ->
+                    ok;
```
Vagabond, line 421: What does returning ok mean here?

vihu: I don't think it does anything if it couldn't lookup the txn in the ets no? So either it was submitted and accepted but we didn't clear the cache or it got rejected. Either way I would imagine that it shouldn't blow up at the very least.

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -406,7 +406,7 @@ check_for_deps_and_resubmit(Txn, Txns, Chain, SubmitF, #txn_data{ acceptions = A
                                                                   dialers = Dialers} = TxnData)->
     %% check if this transaction has any dependencies
     %% figure out what, if anything, this transaction depends on
-    case blockchain_txn:depends_on(Txn, Txns) of
+    case lists:dropwhile(fun(E) -> cached_txn(E) == {error, txn_not_found} end, blockchain_txn:depends_on(Txn, Txns)) of
```
Vagabond, line 409: thinking about this, I don't think the dep txns are ordered here so maybe a lists:filter is better

vihu: Hmm, I'm running the router test which appears to invoke this pretty consistently. Will update shortly.

vihu: Yup, the `lists:filter` change does appear to work as intended afaict

PR #439, opened 2020-04-29
Define a hook function to run every time we absorb a block to the ledger

Vagabond:
The various places we absorbed to the ledger were getting fragmented, so
this hook implements the most complete set of things to do when
absorbing to the chain.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -748,27 +748,7 @@ add_block_(Block, Blockchain, Syncing) ->
                             lager:error("Error absorbing transaction, Ignoring Hash: ~p, Reason: ~p", [blockchain_block:hash_block(Block), Reason]),
                             Error;
                         ok ->
-                            try
-                                ok = blockchain_ledger_v1:maybe_gc_pocs(Blockchain)
-                            catch C:E ->
-                                      lager:info("poc gc failed with ~p:~p", [C,E]),
-                                      ok
-                            end,
-                            Ledger = blockchain:ledger(Blockchain),
-                            case blockchain_ledger_v1:refresh_gateway_witnesses(Hash, Ledger) of
-                                {error, Reason0}=Error0 ->
-                                    lager:error("Error refreshing witnesses, Reason: ~p", [Reason0]),
-                                    Error0;
-                                ok ->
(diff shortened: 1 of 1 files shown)


src/blockchain.erl
```diff
@@ -1125,6 +1102,9 @@ reset_ledger(Height,
                                       ok = blockchain_txn:absorb_and_commit(Block, CAcc,  fun() -> ok end,
                                                                             blockchain_block:is_rescue_block(Block))
                               end,
+                              {ok, Hash} = blockchain_block:hash_block(Block),
+                              %% can only get here if the absorb suceeded
+                              run_absorb_block_hooks(true, Hash, CAcc),
```
evanmcc, line 1107: the current reset doesn't notify, but I suppose that it should, so we can use it on ETL.

src/blockchain.erl
```diff
@@ -1724,13 +1704,7 @@ resync_fun(ChainHeight, LedgerHeight, Blockchain) ->
...
                                               lager:info("absorbing block ~p", [blockchain_block:height(Block)]),
                                               ok = blockchain_txn:unvalidated_absorb_and_commit(Block, Blockchain, fun() -> ok end, blockchain_block:is_rescue_block(Block)),
-                                              case blockchain_ledger_v1:new_snapshot(blockchain:ledger(Blockchain)) of
-                                                  {error, Reason}=Error ->
-                                                      lager:error("Error creating snapshot, Reason: ~p", [Reason]),
-                                                      Error;
-                                                  {ok, NewLedger} ->
-                                                      ok = blockchain_worker:notify({add_block, Hash, true, NewLedger})
-                                              end
+                                              run_absorb_block_hooks(true, Hashm Blockchain)
```
evanmcc, line 1707: typo, this won't compile

evanmcc: also I guess we lose the error here but maybe we don't care?

Vagabond: sorry, brain was not fully engaged

src/blockchain.erl
Vagabond: yeah, I think we should just always fire the hook whenever we add a block to the ledger.

PR #441, opened 2020-04-30
Add option to disable blockchain loading on start

Vagabond:
Some apps may wish to perform their own init before allowing the
blockchain to load and begin any block processing. The no_autoload
option allows this to be done. The app must itself later call
blockchain_worker:load/1 to tell the blockchain worker to load the
chain.

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -27,6 +27,8 @@
 
     new_ledger/1,
 
+    load/2,
+
     maybe_sync/0,
     sync/0,
     cancel_sync/0,
@@ -53,7 +55,7 @@
 
 -record(state,
         {
-         blockchain :: {no_genesis, blockchain:blockchain()} | blockchain:blockchain(),
+         blockchain :: undefined | {no_genesis, blockchain:blockchain()} | blockchain:blockchain(),
          swarm :: undefined | pid(),
          swarm_tid :: undefined | ets:tab(),
          sync_timer = make_ref() :: reference(),
@@ -118,6 +120,10 @@ sync_paused() ->
 
 new_ledger(Dir) ->
     gen_server:call(?SERVER, {new_ledger, Dir}, infinity).
+
+
+load(BaseDir, GenDir) ->
+    gen_server:cast(?SERVER, {load, BaseDir, GenDir}).
 
 %%--------------------------------------------------------------------
 %% @doc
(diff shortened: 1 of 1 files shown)


src/blockchain_worker.erl
```diff
@@ -238,44 +243,26 @@ init(Args) ->
...
-                %% blockchain:new will take care of any repairs needed, possibly asynchronously
-                %%
-                %% do ledger upgrade
-                {ok, GossipRef} = add_handlers(Swarm, Chain),
-                self() ! maybe_sync,
-                {ok, GenesisHash} = blockchain:genesis_hash(Chain),
-                ok = blockchain_txn_mgr:set_chain(Chain),
-                true = libp2p_swarm:network_id(Swarm, GenesisHash),
-                {Chain, GossipRef}
+        case application:get_env(blockchain, no_autoload, false) of
```
Vagabond, line 247: Should not be double negative

src/blockchain_worker.erl
```diff
@@ -577,3 +568,41 @@ send_txn(Txn) ->
...
+        BlockHashAndHeight ->
+            case application:get_env(blockchain, honor_assumed_valid, false) of
+                true ->
+                    BlockHashAndHeight;
+                _ ->
+                    undefined
+            end
+    end.
+
+load_chain(Swarm, Args) ->
```
Vagabond, line 588: Better args here would be good

madninja:
Beyond the two comments you left for yourself it does seem to be working as part of https://github.com/helium/blockchain-etl/pull/45

src/blockchain_worker.erl
```diff
@@ -119,6 +121,9 @@ sync_paused() ->
 new_ledger(Dir) ->
     gen_server:call(?SERVER, {new_ledger, Dir}, infinity).
 
+
+load(Args) ->
+    gen_server:call(?SERVER, {load, Args}, infinity).
```
Vagabond, line 126: Should be a cast so we have less chance of deadlocking

src/blockchain_worker.erl
```diff
@@ -577,3 +571,37 @@ send_txn(Txn) ->
...
+            case application:get_env(blockchain, honor_assumed_valid, false) of
+                true ->
+                    BlockHashAndHeight;
+                _ ->
+                    undefined
+            end
+    end.
+
+load_chain(Swarm, BaseDir, GenDir) ->
+    AssumedValidBlockHashAndHeight = get_assume_valid_height_and_hash(),
```
madninja, line 592: Since we get assumed valid related config information from the app env could we not do the same for base_dir and update_dir?. I know they're not set up the same way, but the blockchain_sup does have all the information already.

Vagabond: It's trickier because for a bunch of the tests we run several blockchains on the same node, so putting these in the appenv would break that.

PR #442, opened 2020-05-05
remove sync tid lookups

andymck:
This replaces various lookups to the swarm ETS table that require a sync call to be made to the swarm server to retrieve the TID and thus eliminates the potential of associated gen_server call timeouts ( which we have hit on a semi frequent basis ).

In the associated libp2p PR, the swarm ETS table is now a named tabled and is named after the blockchain supplied swarm name... the TID is now the same as the swarm name.  

The result is that blockchain core now has no need to retrieve the TID, any function calls that were previously supplying the SwarmPID and which would have resulted in a corresponding sync call to the swarm server to retrieve the associated TID have been modified to instead be supplied with the TID/Swarm Name.

NOTE: the sync flow is still present, its just being circumvented.  I believe I have got all calls using the old sync flow but could well have missed a few.  Any missed ones can easily be replaced as we encounter them.

Associated PRs:

https://github.com/helium/miner/pull/384
https://github.com/helium/erlang-libp2p/pull/276


src/blockchain_swarm.erl
```diff
@@ -44,45 +45,54 @@
...
+    ?SWARM_NAME.
+
 %%--------------------------------------------------------------------
 %% @doc
 %% @end
 %%--------------------------------------------------------------------
 -spec pubkey_bin() -> libp2p_crypto:pubkey_bin().
 pubkey_bin() ->
-    libp2p_swarm:pubkey_bin(whereis(?SWARM_NAME)).
+    libp2p_swarm:pubkey_bin(?SWARM_NAME).
```
evanmcc, line 62: another thing we could do here (perhaps not for this PR) is add the actual bin as a persistent term.

src/blockchain_worker.erl
```diff
@@ -268,7 +268,7 @@ init(Args) ->
     ok = blockchain_event:add_handler(self()),
     lager:info("~p init with ~p", [?SERVER, Args]),
     Swarm = blockchain_swarm:swarm(),
```
evanmcc, line 270: is anything actually using the `Swarm` here? can we remove this call?

src/blockchain_worker.erl
```diff
@@ -357,9 +357,9 @@ handle_cast({integrate_genesis_block, GenesisBlock}, #state{blockchain={no_genes
             lager:info("blockchain started with ~p, consensus ~p", [lager:pr(Blockchain, blockchain), ConsensusAddrs]),
             {ok, GenesisHash} = blockchain:genesis_hash(Blockchain),
             ok = notify({integrate_genesis_block, GenesisHash}),
-            {ok, GossipRef} = add_handlers(Swarm, Blockchain),
+            {ok, GossipRef} = add_handlers(SwarmTid, Blockchain),
             ok = blockchain_txn_mgr:set_chain(Blockchain),
-            true = libp2p_swarm:network_id(Swarm, GenesisHash),
+            true = libp2p_swarm:network_id(blockchain_swarm:tid(), GenesisHash),
```
evanmcc, line 362: why refetch the TID here?

src/blockchain_worker.erl
evanmcc: We set it in the state, but I mean: is anything using it now, can we remove it from the state?

src/blockchain_swarm.erl
```diff
@@ -44,45 +45,54 @@
 start_link(Args) ->
     gen_server:start_link({local, ?SERVER}, ?SERVER, Args, []).
 
+%%--------------------------------------------------------------------
+%% @doc
+%% @end
```
evanmcc, line 50: please don't insert blank comments.

src/blockchain_swarm.erl
```diff
@@ -44,45 +45,54 @@
...
+%%--------------------------------------------------------------------
+-spec tid() -> ets:tab().
+tid() ->
+    ?SWARM_NAME.
```
evanmcc, line 54: if the TID is static, why continue to cache it in the state?  Do we need a function even?  just replace it everywhere with the macro?

src/blockchain_worker.erl
andymck: there are a few cases where the swarm PID is being used, so yeah still required in state

src/blockchain_worker.erl
andymck: oversight, i had  a couple of iterations on this and missed that one. Fixing...

src/blockchain_swarm.erl
andymck: I had it like that originally but i feel having a single access point for this works better.  For instance miner needs to access this which means having to pull in the macro there...whereas miner already calls into blockchain_swarm:swarm(), so the new tid() function is a more like for like replacement and saves sharing header libs/macros

src/blockchain_swarm.erl
andymck: fixed

andymck:
@evanmcc thanks for the review, responded and relevant changes here: https://github.com/helium/blockchain-core/pull/442/commits/0f55cb263bb8456670dfddc6185c4f9c4aeb80e3

PR #443, opened 2020-05-05
Only gossip a block if it's really new to us

Vagabond:
blockchain:add_block didn't use to distinguish a block that it already
had from a block that was new, but validly stored. This add an `exists`
return to add_block so the gossip handler can distinguish between an
existing block and a new one, and only re-gossip the latter.

### DIFF
--- a/src/handlers/blockchain_gossip_handler.erl
+++ b/src/handlers/blockchain_gossip_handler.erl
@@ -17,7 +17,7 @@
 -export([
          init_gossip_data/1,
          handle_gossip_data/3,
-         add_block/3,
+         add_block/4,
          gossip_data/2
         ]).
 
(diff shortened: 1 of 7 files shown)


src/handlers/blockchain_gossip_handler.erl
```diff
@@ -67,17 +62,31 @@ handle_gossip_data(_StreamPid, Data, [Swarm, Blockchain]) ->
...
 
-add_block(Block, Chain, Sender) ->
+add_block(Block, Chain, Sender, Swarm) ->
     lager:debug("Sender: ~p, MyAddress: ~p", [Sender, blockchain_swarm:pubkey_bin()]),
     %% try to acquire the lock with a timeout, will crash this process if we can't get the lock
     ok = blockchain_lock:acquire(5000),
     case blockchain:add_block(Block, Chain) of
         ok ->
             lager:info("got gossipped block ~p", [blockchain_block:height(Block)]),
+            %% pass it along
```
madninja, line 72: maybe declare a `PasstItAlong = fun() -> .. .end`?

PR #445, opened 2020-05-11
allow the parent app to override the metadata fun

evanmcc:
this is so the miner can add lora status to the blockchain-provided metadata.


andymck:
lgtm

PR #447, opened 2020-05-12
Create blank dc entry if it's not found

vihu:
Test PR is in router: https://github.com/helium/routerv3/pull/65


vihu:
> Can you add a simple unit test that tests this creates a blank entry?

Added 6378623

PR #448, opened 2020-05-12
Implement price oracle message handlers

mrallen1:
This PR implements price oracle message handlers. In conjunction with https://github.com/helium/proto/pull/32

Some inspiration: https://community-development.makerdao.com/makerdao-mcd-faqs/faqs/oracles

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -293,3 +293,21 @@
 
 %% how often we attempt to take a snapshot of the ledger
 -define(snapshot_interval, snapshot_interval).
+%% ------------------------------------------------------------------
+
+%% ------------------------------------------------------------------
+%% Price oracle variables
+%%
+%% Oracle public keys - encoded like so...
+%% <<Len1:8/unsigned-integer, Key1/binary, Len2:8/unsigned-integer, Key2/binary, ...>>
+-define(price_oracle_public_keys, price_oracle_public_keys).
+%% How many blocks between price recalculations
+-define(price_oracle_refresh_interval, price_oracle_refresh_interval).
+%% How much delta between the current blockchain height and the transaction is allowed
+-define(price_oracle_height_delta, price_oracle_height_delta).
+%% How many seconds to delay scanning for prices.
+-define(price_oracle_price_scan_delay, price_oracle_price_scan_delay).
+%% How many seconds to stop scanning for oracle prices.
+%% (Will also affect what prices get dropped from the cached list of prices.)
+-define(price_oracle_price_scan_max, price_oracle_price_scan_max).
+%% ------------------------------------------------------------------
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -495,13 +495,13 @@ fold_blocks(Chain0, DelayedHeight, DelayedLedger, Height) ->
                               ok = blockchain_ledger_v1:maybe_gc_pocs(Chain1, Ledger0),
                               ok = blockchain_ledger_v1:maybe_gc_scs(Ledger0),
                               ok = blockchain_ledger_v1:refresh_gateway_witnesses(Hash, Ledger0),
+                              ok = blockchain_ledger_v1:maybe_recalc_price(Chain1, Ledger0),
 
                               %% take an intermediate snapshot here to
                               %% make things faster in the future
                               Ledger1 = ?MODULE:ledger(Chain1),
                               Ctxt = blockchain_ledger_v1:get_context(Ledger1),
                               blockchain_ledger_v1:context_snapshot(Ctxt, Ledger1),
-
                               {ok, Chain1};
                           {error, Reason} ->
                               {error, {block_absorb_failed, H, Reason}}
@@ -1068,7 +1068,10 @@ build_hash_chain_(StopHash, CF, Blockchain = #blockchain{db=DB}, [ParentHash|Tai
             end
     end.
 
-
+-spec fold_chain(fun((Blk :: blockchain_block:block(), AccIn :: any()) -> NewAcc :: any()),
+                     Acc0 :: any(),
+                    Block :: blockchain_block:block(),
+                    Chain :: blockchain()) -> AccOut :: any().
 %% @doc fold blocks in the chain `Chain' backwards from `Block' until a hole in the chain, the genesis block or the function returns `return'.
 fold_chain(Fun, Acc0, Block, Chain) ->
     case Fun(Block, Acc0) of
@@ -2045,6 +2048,8 @@ run_absorb_block_hooks(Syncing, Hash, Blockchain) ->
 
         ok = blockchain_ledger_v1:maybe_gc_scs(Ledger1),
 
+        ok = blockchain_ledger_v1:maybe_recalc_price(Blockchain, Ledger1),
+
         case blockchain_ledger_v1:refresh_gateway_witnesses(Hash, Ledger1) of
             {error, Reason0}=Error0 ->
                 lager:error("Error refreshing witnesses, Reason: ~p", [Reason0]),
@@ -2060,8 +2065,8 @@ run_absorb_block_hooks(Syncing, Hash, Blockchain) ->
                         ok = blockchain_worker:notify({add_block, Hash, Syncing, NewLedger})
                 end
         end
-    catch _:_ ->
-            lager:warning("hooks failed"),
+    catch What:Why:Stack ->
+            lager:warning("hooks failed ~p ~p ~s", [What, Why, lager:pr_stacktrace(Stack, {What, Why})]),
             blockchain_ledger_v1:delete_context(Ledger1)
     end.
 
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -24,7 +24,8 @@
     map_to_bitvector/1,
     bitvector_to_map/2,
     get_pubkeybin_sigfun/1,
-    approx_blocks_in_week/1
+    approx_blocks_in_week/1,
+    vars_keys_to_list/1
 ]).
 
 -ifdef(TEST).
@@ -288,6 +289,17 @@ approx_blocks_in_week(Ledger) ->
             10000
     end.
 
+-spec vars_keys_to_list( Base64String :: binary() ) -> [ binary() ].
+%% @doc Price oracle public keys are encoded like this
+%% <code>
+%% <<KeyLen1/integer, Key1/binary, KeyLen2/integer, Key2/binary, ...>>
+%% </code>
+%% This function takes the encoded string and deserializes
+%% it into a list of binary keys
+%% @end
+vars_keys_to_list(Data) when is_binary(Data) ->
+    [ Key || << Len:8/unsigned-integer, Key:Len/binary >> <=Data ].
+
 %% ------------------------------------------------------------------
 %% EUNIT Tests
 %% ------------------------------------------------------------------
@@ -340,5 +352,16 @@ bitvector_roundtrip_test() ->
     ?assertEqual(M3, bitvector_to_map(64, map_to_bitvector(M3))),
     ?assertEqual(M4, bitvector_to_map(122, map_to_bitvector(M4))),
     ok.
+
+oracle_keys_test() ->
+    #{ public := RawEccPK } = libp2p_crypto:generate_keys(ecc_compact),
+    #{ public := RawEdPK } = libp2p_crypto:generate_keys(ed25519),
+    EccPK = libp2p_crypto:pubkey_to_bin(RawEccPK),
+    EdPK = libp2p_crypto:pubkey_to_bin(RawEdPK),
+    TestStr = base64:encode(<< <<(byte_size(Key)):8/integer, Key/binary>> || Key <- [EccPK, EdPK] >>),
+    Results = vars_keys_to_list(TestStr),
+    ?assertEqual([EccPK, EdPK], Results),
+    Results1 = [ libp2p_crypto:bin_to_pubkey(K) || K <- Results ],
+    ?assertEqual([RawEccPK, RawEdPK], Results1).
 
 -endif.
--- a/src/ledger/v1/blockchain_ledger_oracle_price_entry.erl
+++ b/src/ledger/v1/blockchain_ledger_oracle_price_entry.erl
@@ -1,0 +1,79 @@
+-module(blockchain_ledger_oracle_price_entry).
+
+-export([
+    new/4,
+    timestamp/1,
+    price/1,
+    block_height/1,
+    public_key/1,
+    timestamp/2,
+    price/2,
+    block_height/2,
+    public_key/2
+]).
+
+-record(oracle_price_entry, {
+    timestamp = 0 :: non_neg_integer(),
+    block_height = 0 :: non_neg_integer(),
+    public_key = undefined :: libp2p_crypto:pubkey_bin() | undefined,
+    price = 0 :: non_neg_integer()
+}).
+
+-type oracle_price_entry() :: #oracle_price_entry{}.
+-export_type([oracle_price_entry/0]).
+
+-spec new( Timestamp :: non_neg_integer(),
+           BlockHeight :: non_neg_integer(),
+           PublicKey :: libp2p_crypto:pubkey_bin(),
+           Price :: non_neg_integer() ) -> oracle_price_entry().
+%% @doc Create a new price entry
+new(Timestamp, BlockHeight, PublicKey, Price) ->
+    #oracle_price_entry{
+       timestamp = Timestamp,
+       block_height = BlockHeight,
+       public_key = PublicKey,
+       price = Price
+    }.
+
+-spec timestamp( Entry :: oracle_price_entry() ) -> non_neg_integer().
+%% @doc Return the timestamp from the given price entry.
+timestamp(#oracle_price_entry{ timestamp = T }) -> T.
+
+-spec price( Entry :: oracle_price_entry() ) -> non_neg_integer().
+%% @doc Return the price from the given price entry.
+price(#oracle_price_entry{ price = P }) -> P.
+
+-spec block_height( Entry :: oracle_price_entry() ) -> non_neg_integer().
+%% @doc Return the block_height from the given price entry.
+block_height(#oracle_price_entry{ block_height = H }) -> H.
+
+-spec public_key( Entry :: oracle_price_entry() ) -> libp2p_crypto:pubkey_bin() | undefined.
+%% @doc Return the public_key from the given price entry.
+public_key(#oracle_price_entry{ public_key = K }) -> K.
+
+-spec timestamp( Entry :: oracle_price_entry(),
+                 Value :: non_neg_integer() ) -> Updated :: oracle_price_entry().
+%% @doc Update the timestamp value for the given price entry and return the update
+timestamp(Entry = #oracle_price_entry{}, Value) when is_integer(Value)
+                                                andalso Value >= 0 ->
+    Entry#oracle_price_entry{ timestamp = Value }.
+
+-spec block_height( Entry :: oracle_price_entry(),
+                    Value :: non_neg_integer() ) -> Updated :: oracle_price_entry().
+%% @doc Update the block height value for the given price entry and return the update
+block_height(Entry = #oracle_price_entry{}, Value) when is_integer(Value)
+                                                andalso Value >= 0 ->
+    Entry#oracle_price_entry{ block_height = Value }.
+
+-spec public_key( Entry :: oracle_price_entry(),
+                  Value :: libp2p_crypto:pubkey_bin() ) -> Updated :: oracle_price_entry().
+%% @doc Update the public key value for the given price entry and return the update
+public_key(Entry = #oracle_price_entry{}, Value) when is_binary(Value) ->
+    Entry#oracle_price_entry{ public_key = Value }.
+
+-spec price( Entry :: oracle_price_entry(),
+             Value :: non_neg_integer() ) -> Updated :: oracle_price_entry().
+%% @doc Update the price value for the given price entry and return the update
+price(Entry = #oracle_price_entry{}, Value) when is_integer(Value)
+                                            andalso Value >= 0 ->
+    Entry#oracle_price_entry{ price = Value }.
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -1,4 +1,4 @@
-%%%-------------------------------------------------------------------
+
 %% @doc
 %% == Blockchain Ledger ==
 %% @end
@@ -95,6 +95,11 @@
     raw_fingerprint/2, %% does not use cache
     cf_fold/4,
 
+    maybe_recalc_price/2,
+    add_oracle_price/2,
+    current_oracle_price/1,
+    current_oracle_price_list/1,
+
     apply_raw_changes/2,
 
     set_hexes/2, get_hexes/1,
@@ -132,12 +137,19 @@
     snapshot_threshold_txns/1,
     load_threshold_txns/2,
 
+    load_oracle_price/2,
+    load_oracle_price_list/2,
+
     clean/1, close/1,
     compact/1
 ]).
 
 -include("blockchain.hrl").
 -include("blockchain_vars.hrl").
+
+-ifdef(TEST).
+-export([median/1]).
+-endif.
 
 -record(ledger_v1, {
     dir :: file:filename_all(),
@@ -173,6 +185,8 @@
 -define(MASTER_KEY, <<"master_key">>).
 -define(VARS_NONCE, <<"vars_nonce">>).
 -define(BURN_RATE, <<"token_burn_exchange_rate">>).
+-define(CURRENT_ORACLE_PRICE, <<"current_oracle_price">>). %% stores the current calculated price
+-define(ORACLE_PRICES, <<"oracle_prices">>). %% stores a rolling window of prices
 -define(hex_list, <<"$hex_list">>).
 -define(hex_prefix, "$hex_").
 
@@ -180,6 +194,7 @@
 
 -define(BITS_23, 8388607). %% biggest unsigned number in 23 bits
 -define(BITS_25, 33554431). %% biggest unsigned number in 25 bits
+-define(DEFAULT_ORACLE_PRICE, 0).
 
 -type ledger() :: #ledger_v1{}.
 -type sub_ledger() :: #sub_ledger_v1{}.
@@ -1416,6 +1431,115 @@ maybe_gc_scs(Ledger) ->
 
 %%--------------------------------------------------------------------
 %% @doc
+%% Maybe recalculate the median of the oracle prices that are valid
+%% in a sliding window.
+%%
+%% Prices are calculated every so many blocks, frequency controlled by
+%% the chain_var called `price_oracle_refresh_interval'
+%%
+%% Prices are considered valid if they are:
+%% <ul>
+%%      <li>Older than `price_oracle_scan_delay' seconds, and,</li>
+%%      <li>No more than `price_oracle_scan_max' seconds old</li>
+%% </ul>
+%%
+%% More recent prices from the same oracle replace older prices.
+%%
+%% Additionally there must be valid prices from a majority of the
+%% oracles. Example: If there are 9 oracles, there must be valid
+%% prices from <u>at least</u> 5 different oracles.
+%%
+%% If there are less prices than a majority, then the last calculated
+%% price is reused.
+%% @end
+%%--------------------------------------------------------------------
+-spec maybe_recalc_price( Blockchain :: blockchain:blockchain(),
+                          Ledger :: ledger() ) -> ok.
+maybe_recalc_price(Blockchain, Ledger) ->
+    case blockchain:config(?price_oracle_refresh_interval, Ledger) of
+        {error, not_found} -> ok;
+        {ok, I} -> do_maybe_recalc_price(I, Blockchain, Ledger)
+    end.
+
+do_maybe_recalc_price(Interval, Blockchain, Ledger) ->
+    DefaultCF = default_cf(Ledger),
+    {ok, CurrentHeight} = current_height(Ledger),
+    {ok, LastPrice} = current_oracle_price(Ledger),
+
+    case CurrentHeight rem Interval == 0 of
+        false -> ok;
+        true ->
+            {ok, Block} = blockchain:get_block(CurrentHeight, Blockchain),
+            BlockT = blockchain_block:time(Block),
+            {NewPrice, NewPriceList} = recalc_price(LastPrice, BlockT, DefaultCF, Ledger),
+            cache_put(Ledger, DefaultCF, ?ORACLE_PRICES, term_to_binary(NewPriceList)),
+            cache_put(Ledger, DefaultCF, ?CURRENT_ORACLE_PRICE, term_to_binary(NewPrice))
+    end.
+
+recalc_price(LastPrice, BlockT, _DefaultCF, Ledger) ->
+    {ok, DelaySecs} = blockchain:config(?price_oracle_price_scan_delay, Ledger),
+    {ok, MaxSecs} = blockchain:config(?price_oracle_price_scan_max, Ledger),
+    StartScan = BlockT - DelaySecs, % typically 1 hour (in seconds)
+    EndScan = BlockT - MaxSecs, % typically 1 day + 1 hour (in seconds)
+    {ok, Prices} = current_oracle_price_list(Ledger),
+    NewPriceList = trim_price_list(EndScan, Prices),
+    {ok, RawOracleKeys} = blockchain:config(?price_oracle_public_keys, Ledger),
+    Maximum = length(blockchain_utils:vars_keys_to_list(RawOracleKeys)),
+    Minimum = (Maximum div 2) + 1,
+
+    ValidPrices = lists:foldl(
+                    fun(E, Acc) ->
+                            select_prices_by_time(StartScan, EndScan, E, Acc)
+                    end, #{},
+                    %% guarantee that prices are sorted in timestamp order
+                    %% so that newer prices will replace older prices
+                    lists:sort(fun sort_price_entry_time/2, NewPriceList)),
+
+    NumPrices = maps:size(ValidPrices),
+
+    case NumPrices >= Minimum andalso NumPrices =< Maximum of
+        true -> {median(maps:values(ValidPrices)), NewPriceList};
+        false -> {LastPrice, NewPriceList}
+    end.
+
+sort_price_entry_time(A, B) ->
+    ATime = blockchain_ledger_oracle_price_entry:timestamp(A),
+    BTime = blockchain_ledger_oracle_price_entry:timestamp(B),
+
+    ATime =< BTime.
+
+select_prices_by_time(Start, End, Entry, Acc) ->
+    T = blockchain_ledger_oracle_price_entry:timestamp(Entry),
+    PK = blockchain_ledger_oracle_price_entry:public_key( Entry),
+    P = blockchain_ledger_oracle_price_entry:price(Entry),
+
+    if
+        T >= Start andalso T =< End -> Acc#{ PK => P };
+        T > End -> Acc;
+        true -> Acc
+    end.
+
+median(L) ->
+    Sorted = lists:sort(L),
+    Len = length(L),
+    case Len rem 2 of
+        0 -> lists:nth(Len div 2, Sorted);
+        1 ->
+            Div = Len div 2,
+            Div1 = Div+1,
+            (lists:nth(Div, Sorted) + lists:nth(Div1, Sorted)) div 2 %% no floats
+    end.
+
+trim_price_list(LastTime, PriceEntries) ->
+    lists:dropwhile(fun(Entry) ->
+                            case blockchain_ledger_oracle_price_entry:timestamp(Entry) of
+                                T when T > LastTime -> true;
+                                _ -> false
+                            end
+                    end, PriceEntries).
+
+%%--------------------------------------------------------------------
+%% @doc
 %% @end
 %%--------------------------------------------------------------------
 -spec find_entry(libp2p_crypto:pubkey_bin(), ledger()) -> {ok, blockchain_ledger_entry_v1:entry()}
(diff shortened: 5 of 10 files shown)


src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -782,6 +786,9 @@ validate_var(?sc_grace_blocks, Value) ->
 validate_var(?dc_payload_size, Value) ->
     validate_int(Value, "dc_payload_size", 1, 32, false);
 
+validate_var(?price_oracle_public_keys, Value) ->
+    validate_public_key_format(Value);
```
Vagabond, line 790: Would call this public_keys_format if there's a list?

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,164 @@
...
+fee(_Txn) ->
+    0.
+
+%%--------------------------------------------------------------------
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec sign(txn_price_oracle(), libp2p_crypto:sig_fun()) -> txn_price_oracle().
+sign(Txn, SigFun) ->
+    EncodedTxn = blockchain_txn_price_oracle_v1_pb:encode_msg(Txn),
```
Vagabond, line 93: Zero the signature field before signing

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,198 @@
...
+
+%%--------------------------------------------------------------------
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec new(OraclePublicKey :: binary(), Price :: non_neg_integer(),
+          BlockHeight :: non_neg_integer()) -> txn_price_oracle().
+new(OraclePK, Price, BlockHeight) ->
+    #blockchain_txn_price_oracle_v1_pb{
+       oracle_public_key=OraclePK,
```
vihu, line 45: I'm most definitely nitpicking but we already know we are `blockchain_txn_price_oracle_v1` so just name this a `pubkey` or whatever, no need to have oracle here too no?

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,198 @@
...
+    OraclePK = ?MODULE:oracle_public_key(Txn),
+    BlockHeight = ?MODULE:block_height(Txn),
+    BaseTxn = Txn#blockchain_txn_price_oracle_v1_pb{signature = <<>>},
+    EncodedTxn = blockchain_txn_price_oracle_v1_pb:encode_msg(BaseTxn),
+    case blockchain_txn:validate_fields([{{oracle_public_key, OraclePK},
+                                          {member, ?price_oracle_public_keys}}, % XXX: Implement
+                                         {{price, Price}, {integer, 1, 100}}]) of
+        ok ->
+            %% maybe these tests should be reversed...
+            %% or maybe we should fold over a list of closures...
```
vihu, line 130: or maybe railway_mode :trollface:

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,198 @@
...
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec hash(txn_price_oracle()) -> blockchain_txn:hash().
+hash(Txn) ->
+    BaseTxn = Txn#blockchain_txn_price_oracle_v1_pb{signature = <<>>},
+    EncodedTxn = blockchain_txn_price_oracle_v1_pb:encode_msg(BaseTxn),
+    crypto:hash(sha256, EncodedTxn).
+
+%%--------------------------------------------------------------------
```
vihu, line 61: Just get rid of these empty docs, they are noise. Or add some docs :slightly_smiling_face:, but please no more empty docs. I don't know why we do this all over the place.

src/blockchain_utils.erl
```diff
@@ -288,6 +289,23 @@ approx_blocks_in_week(Ledger) ->
...
+-spec vars_oracle_keys_to_list( Base64String :: binary() ) -> [ binary() ].
+%% @doc Price oracle public keys are encoded like this
+%% <code>
+%% base64:encode(<<KeyLen1/integer, Key1/binary, KeyLen2/integer, Key2/binary, ...>>)
+%% </code>
+%% This function takes the encoded string and deserializes
+%% it into a list of binary keys
+%% @end
+vars_oracle_keys_to_list(B64Data) when is_binary(B64Data) ->
+    <<Len/integer, Rest/binary>> = base64:decode(B64Data),
```
Vagabond, line 301:
```suggestion
    <<Len:8/unsigned-integer, Rest/binary>> = base64:decode(B64Data),
```

Vagabond: You can use list and binary compehensions here to make this easier, I think:

```
4> Encoded = << <<(byte_size(X)):8/integer, X/binary>> || X <- [<<"hello">>, <<"world">>, <<"how">>, <<"are">>, <<"things">>] >>.
<<5,104,101,108,108,111,5,119,111,114,108,100,3,104,111,
  119,3,97,114,101,6,116,104,105,110,103,115>>
5> [ Value || <<Len:8/integer-unsigned, Value:Len/binary>> <= Encoded ].
[<<"hello">>,<<"world">>,<<"how">>,<<"are">>,<<"things">>]
```

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,238 @@
...
+-spec absorb(txn_price_oracle(), blockchain:blockchain()) -> ok | {error, any()}.
+absorb(Txn, Chain) ->
+    Ledger = blockchain:ledger(Chain),
+    Price = ?MODULE:price(Txn),
+    OraclePK = ?MODULE:oracle_public_key(Txn),
+
+    blockchain_ledger_v1:add_oracle_price(OraclePK, Price, Ledger),
+
+    %% potentially recalculate price
+    Ht = blockchain_ledget_v1:current_height(Ledger),
```
Vagabond, line 162:
```suggestion
    Ht = blockchain_ledger_v1:current_height(Ledger),
```

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,238 @@
...
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec absorb(txn_price_oracle(), blockchain:blockchain()) -> ok | {error, any()}.
+absorb(Txn, Chain) ->
+    Ledger = blockchain:ledger(Chain),
+    Price = ?MODULE:price(Txn),
+    OraclePK = ?MODULE:oracle_public_key(Txn),
+
+    blockchain_ledger_v1:add_oracle_price(OraclePK, Price, Ledger),
```
Vagabond, line 159: So there can only be one price per reporter?

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,238 @@
...
+
+    %% potentially recalculate price
+    Ht = blockchain_ledget_v1:current_height(Ledger),
+    case blockchain:get_block(Ht) of
+        {ok, Block} -> recalc_price(Block, Chain);
```
Vagabond, line 164: The computed value here seems to be dropped?

src/blockchain_utils.erl
```diff
@@ -288,6 +289,23 @@ approx_blocks_in_week(Ledger) ->
             10000
     end.
 
+-spec vars_oracle_keys_to_list( Base64String :: binary() ) -> [ binary() ].
```
Vagabond, line 292: Would suggest naming this more generically so we can use it for other lists of packed keys in the future.

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
Vagabond: And we never consult it?

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,238 @@
+%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain Transaction Price Oracle ==
+%% @end
+%%%-------------------------------------------------------------------
+-module(blockchain_txn_price_oracle_v1).
+
+-behavior(blockchain_txn).
```
vihu, line 8: We should also start adding `-behavior(blockchain_json)` for new txns from now on.

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,238 @@
...
+
+-include("blockchain_utils.hrl").
+-include("blockchain_vars.hrl").
+-include_lib("helium_proto/include/blockchain_txn_price_oracle_v1_pb.hrl").
+
+%% Controls how old the price is allowed to be.
+%%
+%% If the current block height differs from the recorded block height
+%% in the message, reject the price.
+-define(MAX_HEIGHT_DIFF, 20).
```
vihu, line 18: Maybe this needs to be a chain variable? Something we can tweak if need be later down the line?

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,238 @@
...
+    case blockchain_txn:validate_fields([{{oracle_public_key, OraclePK},
+                                          {member, ?price_oracle_public_keys}}, % XXX: Implement
+                                         {{price, Price}, {integer, 1, 100}}]) of
+        ok ->
+            case libp2p_crypto:verify(EncodedTxn, Signature, OraclePK) of
+                false ->
+                    {error, bad_signature};
+                true ->
+                    case validate_block_height(BlockHeight,
+                                               blockchain_ledger_v1:current_height(Ledger)) of
```
vihu, line 138: This is incorrect `blockchain_ledger_v1:current_height(Ledger)` gives you either an `{ok, Height}` or `{error, Reason}`.

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
vihu:
```suggestion
    {ok, Ht} = blockchain_ledger_v1:current_height(Ledger),
```

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,238 @@
...
+%% ------------------------------------------------------------------
+%% Private functions
+%% ------------------------------------------------------------------
+validate_block_height(MsgHeight, Current) when (Current - MsgHeight) < ?MAX_HEIGHT_DIFF ->
+    true;
+validate_block_height(_MsgHeight, _Current) -> false.
+
+recalc_price(Start, Chain) ->
+    CurrentTime = erlang:system_time(seconds),
+    HrAgo = CurrentTime - 3600,
```
vihu, line 190: Maybe add some comments around what these numbers stand for?

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,238 @@
...
+    EligibleBlocks = blockchain:fold_chain(fun(Block, Acc) ->
+                                                   BlockT = blockchain_block:time(Block),
+                                                   if
+                                                       BlockT > HrAgo andalso BlockT < DayAgo -> [ Block | Acc ];
+                                                       BlockT > DayAgo -> return;
+                                                       true -> Acc
+                                                   end
+                                           end, [], Start, Chain),
+
+    median(lists:sort(lists:flatten(
```
vihu, line 202: Might be better to separate this out, there appears to be a lot happening with this sorted flattened folded list :trollface:

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
mrallen1: I guess not. I'll take it out.

src/blockchain_utils.erl
mrallen1: I changed it to `vars_keys_to_list`

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
mrallen1: Could also just release a new value in the source code. But I guess a chain var would be more secure.

Vagabond: Variables in the source code can't change because they'd break syncing.

include/blockchain_price_entry.hrl
```diff
@@ -0,0 +1,7 @@
+-record(oracle_price_entry, {
```
Vagabond, line 1: How does this relate to the proto record?

src/blockchain.erl
```diff
@@ -483,6 +483,7 @@ fold_blocks(Chain0, DelayedHeight, DelayedLedger, Height) ->
                               ok = blockchain_ledger_v1:maybe_gc_pocs(Chain1, Ledger0),
                               ok = blockchain_ledger_v1:maybe_gc_scs(Ledger0),
                               ok = blockchain_ledger_v1:refresh_gateway_witnesses(Hash, Ledger0),
+                              ok = blockchain_ledget_v1:maybe_recalc_price(Ledger0),
```
Vagabond, line 486:
```suggestion
                              ok = blockchain_ledger_v1:maybe_recalc_price(Ledger0),
```

include/blockchain_price_entry.hrl
mrallen1: It's basically a convenience to label the fields of the tuple I want to store in the ledger

include/blockchain_vars.hrl
```diff
@@ -293,3 +293,17 @@
...
+
+%% ------------------------------------------------------------------
+%% Price oracle variables
+%%
+%% Oracle public keys
+%% array of B64 encoded public keys
+%% base64:encode(<<Len1:8/unsigned-integer, Key1/binary, Len2:8/unsigned-integer, Key2/binary, ...>>)
+-define(price_oracle_public_keys, price_oracle_public_keys).
+%% How many blocks between price recalculations
+-define(price_oracle_refresh_interval, price_oracle_refresh_oracle).
```
vihu, line 306:
```suggestion
-define(price_oracle_refresh_interval, price_oracle_refresh_interval).
```

src/blockchain_utils.erl
```diff
@@ -341,4 +353,17 @@ bitvector_roundtrip_test() ->
...
     ok.
 
+oracle_keys_test() ->
+    #{ public := RawEccPK } = libp2p_crypto:generate_keys(ecc_compact),
+    #{ public := RawEdPK } = libp2p_crypto:generate_keys(ed25519),
+    EccPK = libp2p_crypto:pubkey_to_bin(RawEccPK),
+    EdPK = libp2p_crypto:pubkey_to_bin(RawEdPK),
+    LenEcc = byte_size(EccPK),
+    LenEd = byte_size(EdPK),
+    TestStr = base64:encode(<<LenEcc/integer, EccPK/binary, LenEd/integer, EdPK/binary>>),
```
Vagabond, line 363:
```suggestion
    TestStr = base64:encode(<< <<(byte_size(Key)):8/integer, Key/binary>> || Key <- [EccPK, EdPK] >>),
```

Vagabond: Or have a proper function for doing this and use that

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1383,6 +1390,92 @@ maybe_gc_scs(Ledger) ->
...
+%% Maybe recalculate the median of the oracle prices that are valid
+%% in a sliding window over 24 hours.
+%%
+%% Prices are calculated every so many blocks, frequency controlled by
+%% the chain_var called `price_oracle_refresh_interval'
+%%
+%% Prices are considered valid if they are:
+%% <ul>
+%%      <li>Older than one hour, and,</li>
+%%      <li>No more than 24 hours old</li>
```
Vagabond, line 1404: 24 or 25?

include/blockchain_price_entry.hrl
vihu: This makes me feel uncomfortable

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1383,6 +1390,92 @@ maybe_gc_scs(Ledger) ->
...
+
+    case CurrentHeight - LastHeight of
+        X when X < Interval -> ok;
+        X when X >= Interval ->
+            NewPrice = recalc_price(LastPrice, DefaultCF, Ledger),
+            cache_put(Ledger, DefaultCF, ?CURRENT_ORACLE_PRICE, {CurrentHeight, NewPrice})
+    end.
+
+recalc_price(LastPrice, DefaultCF, Ledger) ->
+    CurrentTime = erlang:system_time(seconds), %% XXX: probably should be time in head block...
```
Vagabond, line 1439: Yes, it should be

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1977,6 +2070,12 @@ allocate_subnet(Size, _Itr, {error, invalid_iterator}, {LastBase, LastSize}) ->
             {ok, <<NewBase:25/integer-unsigned-big, Mask:23/integer-unsigned-big>>}
     end.
 
+-spec add_oracle_price(oracle_price_entry(), ledger()) -> ok.
+add_oracle_price(PriceEntry, Ledger) ->
+    DefaultCF = default_cf(Ledger),
+    {ok, Prices} = cache_get(Ledger, DefaultCF, ?ORACLE_PRICES, []),
+    cache_put(Ledger, DefaultCF, ?ORACLE_PRICES, [ PriceEntry | Prices ]).
```
Vagabond, line 2077: We need to trim prices that are too old, right?

mrallen1: yeah thats right

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,249 @@
...
+-spec sign(txn_price_oracle(), libp2p_crypto:sig_fun()) -> txn_price_oracle().
+sign(Txn, SigFun) ->
+    Zeroed = Txn#blockchain_txn_price_oracle_v1_pb{signature = <<>>},
+    EncodedTxn = blockchain_txn_price_oracle_v1_pb:encode_msg(Zeroed),
+    Txn#blockchain_txn_price_oracle_v1_pb{signature=SigFun(EncodedTxn)}.
+
+%%--------------------------------------------------------------------
+%% @doc
+%% Validate that this txn has a valid oracle public key, that the
+%% price is an integer and not negative, and that the signature is
```
Vagabond, line 135: can you comment more on the block height checks?

src/ledger/v1/blockchain_ledger_v1.erl
mrallen1: I made it 25 so we look back through a full day, excluding the current hour

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
mrallen1: I did so but not sure it's what you're looking for

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,254 @@
...
+       fee => fee(Txn),
+       public_key => public_key(Txn),
+       price => price(Txn),
+       block_height => block_height(Txn)
+     }.
+
+%% ------------------------------------------------------------------
+%% Private functions
+%% ------------------------------------------------------------------
+validate_block_height(MsgHeight, Current, MaxHeight) when (Current - MsgHeight) =< MaxHeight ->
```
Vagabond, line 222: I think we should also check if there's already a vote from this oracle at or before this height.

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,254 @@
...
+-spec absorb(txn_price_oracle(), blockchain:blockchain()) -> ok | {error, any()}.
+absorb(Txn, Chain) ->
+    Ledger = blockchain:ledger(Chain),
+    {ok, LedgerHeight} = blockchain_ledger_v1:current_height(Ledger),
+    Blk = blockchain:get_block(LedgerHeight),
+    Time = blockchain_block:time(Blk),
+
+    Entry = #oracle_price_entry{
+               price = ?MODULE:price(Txn),
+               public_key = ?MODULE:public_key(Txn),
```
Vagabond, line 191: We should store the block height from the txn here.

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
```diff
@@ -0,0 +1,254 @@
...
+                                         signature=Sig}) ->
+    io_lib:format("type=price_oracle oracle_signature=~p, price=~p, block_height=~p, signature=~p",
+                  [OraclePK, Price, BH, Sig]).
+
+-spec to_json(txn_price_oracle(), blockchain_json:opts()) -> blockchain_json:json_object().
+to_json(Txn, _Opts) ->
+    #{ type => <<"price_oracle_v1">>,
+       hash => ?BIN_TO_B64(hash(Txn)),
+       fee => fee(Txn),
+       public_key => public_key(Txn),
```
madninja, line 214: This needs `?BIN_TO_B58(public_key(Txn))`

src/transactions/v1/blockchain_txn_price_oracle_v1.erl
mrallen1: OK, good idea

include/blockchain_price_entry.hrl
mrallen1: Why does it make you feel uncomfortable @vihu

Vagabond: We try not to leak records across module boundaries, as a rule. This would be nicer to have as a module that provides accessors to the record rather than leaking the record definition around.

vihu:
@mrallen1 I believe this needs a rebase

include/blockchain_price_entry.hrl
mrallen1: Should I change it then?

mrallen1:
@vihu Yeah it does and I'll do a rebase+squash after review

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1977,6 +2079,41 @@ allocate_subnet(Size, _Itr, {error, invalid_iterator}, {LastBase, LastSize}) ->
...
+    end.
+
+-spec current_oracle_price_list(ledger()) -> {ok, [ oracle_price_entry() ]} | {error, any()}.
+current_oracle_price_list(Ledger) ->
+    DefaultCF = default_cf(Ledger),
+    case cache_get(Ledger, DefaultCF, ?ORACLE_PRICES, []) of
+        {ok, BinPrices} ->
+            {ok, binary_to_term(BinPrices)};
+        not_found ->
+            {ok, []};
```
vihu, line 2112: I was thinking if it would make more sense to have this be something like an `{error, not_found}` here. Ignore if that will get overly complicated or is required by design to have an empty list as a default.

vihu:
A general comment on the CTs. I see `ExtraVars` being created for every test case. Is it possible to have them be created at `init_per_testcase` level? If they need overwriting we can do in the test case itself I'd imagine?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -95,6 +95,12 @@
     raw_fingerprint/2, %% does not use cache
     cf_fold/4,
 
+    maybe_recalc_price/2,
+    add_oracle_price/2,
+    current_oracle_price/1,
+    current_oracle_price_list/1,
+    median/1, %% for test purposes only
```
vihu, line 102: Maybe put this inside an `ifdef TEST` then?

test/blockchain_price_oracle_SUITE.erl
```diff
@@ -0,0 +1,398 @@
...
+    _Blocks1 = [
+               begin
+                {ok, Block} = test_utils:create_block(ConsensusMembers, []),
+                blockchain:add_block(Block, Chain),
+                Block
+               end || _ <- lists:seq(1, BlocksN) ],
+
+    %% this block was already added 25 blocks ago, let's try to submit it
+    %% again.
+    ?assertMatch(exists, blockchain:add_block(PriceBlock, Chain)),
```
vihu, line 354: I don't truly understand this check to be honest. I would have expected that the `replay_txn` test would mark a resubmitted txn as invalid. Adding a previously added block would always give you `exists` no? I don't think we're actually check replay txn here...

src/ledger/v1/blockchain_ledger_v1.erl
mrallen1: I think it's better to return an empty list than an error, tbh

test/blockchain_price_oracle_SUITE.erl
mrallen1: OK, fair

include/blockchain_price_entry.hrl
Vagabond: yes.

rebar.lock
```diff
@@ -52,11 +52,11 @@
  {<<"intercept">>,{pkg,<<"intercept">>,<<"1.0.0">>},2},
  {<<"lager">>,
   {git,"https://github.com/erlang-lager/lager.git",
-       {ref,"d88b3663b9e6d5313207b30869deb160ffd503ec"}},
+       {ref,"f26a8ea0f4c2ed6506a54e0b9890edd8148b9500"}},
   0},
  {<<"libp2p">>,
   {git,"https://github.com/helium/erlang-libp2p.git",
-       {ref,"abff2d73d0d783a1bca7e852fde07db393e8919d"}},
```
Vagabond, line 59: Why all these bumps?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1383,6 +1394,98 @@ maybe_gc_scs(Ledger) ->
...
+%% @doc
+%% Maybe recalculate the median of the oracle prices that are valid
+%% in a sliding window over 24 hours.
+%%
+%% Prices are calculated every so many blocks, frequency controlled by
+%% the chain_var called `price_oracle_refresh_interval'
+%%
+%% Prices are considered valid if they are:
+%% <ul>
+%%      <li>Older than one hour, and,</li>
```
Vagabond, line 1407: Probably better to frame it in terms of chain vars

include/blockchain_price_entry.hrl
vihu: Yeah, make it something like a `blockchain_ledger_oracle_price_entry.erl` perhaps? With whatever getter/setter funs required within that module.

src/ledger/v1/blockchain_ledger_v1.erl
mrallen1: Yes, good idea.

rebar.lock
mrallen1: sorry i didn't notice that! must've been when i upgraded something using `rebar3 upgrade`

include/blockchain_price_entry.hrl
mrallen1: ok, this work is done.

src/ledger/v1/blockchain_ledger_oracle_price_entry.erl
```diff
@@ -0,0 +1,79 @@
...
+    timestamp/2,
+    price/2,
+    block_height/2,
+    public_key/2
+]).
+
+-record(oracle_price_entry, {
+    timestamp = 0 :: non_neg_integer(),
+    block_height = 0 :: non_neg_integer(),
+    public_key = undefined :: binary() | undefined,
```
vihu, line 18: I'm just putting this comment here but it's more generic I suppose. Why is the public_key a "binary" an not a `libp2p_crypto:pubkey_bin`? We should be storing these keys in `libp2p_crypto:pubkey_bin` form on the ledger as well. Unless there is some constraint which I don't quite understand...

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1419,6 +1429,101 @@ maybe_gc_scs(Ledger) ->
...
+            BlockT = blockchain_block:time(Block),
+            {NewPrice, NewPriceList} = recalc_price(LastPrice, BlockT, DefaultCF, Ledger),
+            cache_put(Ledger, DefaultCF, ?ORACLE_PRICES, term_to_binary(NewPriceList)),
+            cache_put(Ledger, DefaultCF, ?CURRENT_ORACLE_PRICE, term_to_binary({CurrentHeight, NewPrice}))
+    end.
+
+recalc_price(LastPrice, BlockT, _DefaultCF, Ledger) ->
+    {ok, DelaySecs} = blockchain:config(?price_oracle_price_scan_delay, Ledger),
+    {ok, MaxSecs} = blockchain:config(?price_oracle_price_scan_max, Ledger),
+    StartScan = BlockT - DelaySecs, % typically 1 hour (in seconds)
```
vihu, line 1477: Still suspicious about this...Does it do what you expect it to do? `BlockT` is in ms and `DelaySecs` presumably is seconds. Might be good to have an eunit just for these funs and check whether they are working as intended?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1419,6 +1429,101 @@ maybe_gc_scs(Ledger) ->
...
+    PK = blockchain_ledger_oracle_price_entry:public_key( Entry),
+    P = blockchain_ledger_oracle_price_entry:price(Entry),
+
+    if
+        T >= Start andalso T =< End -> Acc#{ PK => P };
+        T > End -> Acc;
+        true -> Acc
+    end.
+
+median(L) ->
```
vihu, line 1508: Maybe put this fun in the utils module?

src/ledger/v1/blockchain_ledger_v1.erl
mrallen1: BlockT's unit is epoch seconds

src/ledger/v1/blockchain_ledger_oracle_price_entry.erl
mrallen1: OK I will update the units here

src/ledger/v1/blockchain_ledger_v1.erl
mrallen1: Maybe if we need to calculate medians from somewhere else too, we could move it. I think it's fine where it is for now.

src/ledger/v1/blockchain_ledger_v1.erl
mrallen1: @Vagabond can you confirm please? But every block time I've looked at has been epoch seconds.

Vagabond: block times are in seconds

include/blockchain_vars.hrl
```diff
@@ -293,3 +293,22 @@
 
 %% how often we attempt to take a snapshot of the ledger
 -define(snapshot_interval, snapshot_interval).
+%% ------------------------------------------------------------------
+
+%% ------------------------------------------------------------------
+%% Price oracle variables
+%%
+%% Oracle public keys
+%% array of B64 encoded public keys
```
Vagabond, line 302: Not base64, right?

src/blockchain.erl
evanmcc: this seems still not fixed?

src/blockchain.erl
```diff
@@ -2031,6 +2034,8 @@ run_absorb_block_hooks(Syncing, Hash, Blockchain) ->
 
         ok = blockchain_ledger_v1:maybe_gc_scs(Ledger1),
 
+        ok = blockchain_ledger_v1:maybe_recalc_price(Blockchain, Ledger1),
```
evanmcc, line 2037: ideally these would always be in the same order to prevent any subtle interactions.  I have a PR that consolidates them, so OK to ignore this reordering:  however, there should be a call to `maybe_recalc_price` in `blockchain_txn:absorb_delayed`.  search for the other maybe functions there.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1419,6 +1429,101 @@ maybe_gc_scs(Ledger) ->
...
+%%--------------------------------------------------------------------
+-spec maybe_recalc_price( Blockchain :: blockchain:blockchain(),
+                          Ledger :: ledger() ) -> ok.
+maybe_recalc_price(Blockchain, Ledger) ->
+    {ok, Interval} = blockchain:config(?price_oracle_refresh_interval, Ledger),
+    DefaultCF = default_cf(Ledger),
+    {ok, CurrentHeight} = current_height(Ledger),
+    {ok, {LastHeight, LastPrice}} = current_oracle_price(Ledger),
+
+    case CurrentHeight - LastHeight of
```
Vagabond, line 1464: Why not case CurrentHeight mod Interval == 0 of...

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1419,6 +1429,101 @@ maybe_gc_scs(Ledger) ->
...
+%% price is reused.
+%% @end
+%%--------------------------------------------------------------------
+-spec maybe_recalc_price( Blockchain :: blockchain:blockchain(),
+                          Ledger :: ledger() ) -> ok.
+maybe_recalc_price(Blockchain, Ledger) ->
+    {ok, Interval} = blockchain:config(?price_oracle_refresh_interval, Ledger),
+    DefaultCF = default_cf(Ledger),
+    {ok, CurrentHeight} = current_height(Ledger),
+    {ok, {LastHeight, LastPrice}} = current_oracle_price(Ledger),
```
evanmcc, line 1462: I feel that it's simpler to do mod intervals and not store the last height, but this is probably ok.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1419,6 +1429,101 @@ maybe_gc_scs(Ledger) ->
...
+
+    case CurrentHeight - LastHeight of
+        X when X < Interval -> ok;
+        X when X >= Interval ->
```
evanmcc, line 1466: specifically it's this that's making me worried, about somehow running this code twice.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1419,6 +1429,101 @@ maybe_gc_scs(Ledger) ->
...
+    Len = length(L),
+    case Len rem 2 of
+        0 -> lists:nth(Len div 2, Sorted);
+        1 ->
+            Div = Len div 2,
+            Div1 = Div+1,
+            (lists:nth(Div, Sorted) + lists:nth(Div1, Sorted)) div 2 %% no floats
+    end.
+
+trim_price_list(MaxSecs, PriceEntries) ->
```
evanmcc, line 1519: nit: naming here is odd, would prefer
```suggestion
trim_price_list(LastTime, PriceEntries) ->
```
or something like that...

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -824,6 +837,26 @@ validate_var(?snapshot_version, Value) ->
...
 
+validate_var(?price_oracle_public_keys, Value) ->
+    validate_oracle_public_keys_format(Value);
+
+validate_var(?price_oracle_price_scan_delay, Value) ->
+    %% Allowed: 0 seconds to 86400 seconds (1 day)
+    validate_int(Value, "price_oracle_price_scan_delay", 0, 86400, false);
+
+validate_var(?price_oracle_price_scan_max, Value) ->
+    %% Allowed: 0 seconds to 604800 seconds (7 days)
```
Vagabond, line 848: 0 seems like it might be bad?

test/blockchain_price_oracle_SUITE.erl
Vagabond: Yea, this is not doing what you think it is, you need to make a new block with the old transactions in it

Vagabond:
I think, barring the remaining comments, this looks pretty good.

include/blockchain_vars.hrl
mrallen1: Yes? I thought the chain_var *would* be Base64 encoded? Did I misunderstand something here?

Vagabond: it doesn't need to be and consumes unneeded space by storing it as such. The ledger is binary-clean IIRC. I also don't think we're storing it base64'd as of now.

src/transactions/v1/blockchain_txn_vars_v1.erl
mrallen1: It may be desirable to allow 0 in the case we want to disable this for some reason?

include/blockchain_vars.hrl
madninja: so these need to be pubkey_bin()'s? If they have to be in text form, public keys are _always_ b58 (not b64) encoded

madninja: I take it back.. you're describing a plain binary encoding.. pubkey_bin() types basically

src/transactions/v1/blockchain_txn_vars_v1.erl
Vagabond: ah, fair, I suppose.

include/blockchain_vars.hrl
Vagabond: yes

test/blockchain_price_oracle_SUITE.erl
Vagabond: I cleaned this up

include/blockchain_vars.hrl
Vagabond: I've removed the unnecessary base64, all the tests still pass

Vagabond:
We need to confirm Evan's two concerns are addressed, but other than that :+1: from me

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2013,6 +2127,45 @@ allocate_subnet(Size, _Itr, {error, invalid_iterator}, {LastBase, LastSize}) ->
...
+    DefaultCF = default_cf(Ledger),
+    Prices = case cache_get(Ledger, DefaultCF, ?ORACLE_PRICES, []) of
+                 {ok, BinPrices} ->
+                     binary_to_term(BinPrices);
+                 not_found ->
+                     []
+             end,
+    cache_put(Ledger, DefaultCF, ?ORACLE_PRICES, term_to_binary([ PriceEntry | Prices ])).
+
+-spec current_oracle_price(ledger()) -> {ok, {non_neg_integer(), non_neg_integer()}} | {error, any()}.
```
evanmcc, line 2143: looks like this spec may be stale

mrallen1: good catch

PR #451, opened 2020-05-15
Add state_channel cli commands

vihu:
Exposes two state channel commands: `sc list` and `sc active`

### DIFF
--- a/src/cli/blockchain_cli_sc.erl
+++ b/src/cli/blockchain_cli_sc.erl
@@ -1,0 +1,149 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain CLI SC ==
+%% @end
+%%%-------------------------------------------------------------------
+-module(blockchain_cli_sc).
+
+-behavior(clique_handler).
+
+-export([register_cli/0]).
+
+-include("blockchain.hrl").
+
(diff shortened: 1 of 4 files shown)


src/cli/blockchain_cli_sc.erl
```diff
@@ -0,0 +1,149 @@
...
+                        {summaries, io_lib:format("~p", [Summaries])}
+                       ] | Acc]
+              end, [], SCs).
+
+format_sc_summaries(Summaries) ->
+    lists:foldl(fun(Summary, Acc) ->
+                        {ok, ClientName} = erl_angry_purple_tiger:animal_name(libp2p_crypto:bin_to_b58(blockchain_state_channel_summary_v1:client_pubkeybin(Summary))),
+                        NumDCs = blockchain_state_channel_summary_v1:num_dcs(Summary),
+                        NumPackets = blockchain_state_channel_summary_v1:num_packets(Summary),
+                        [ {receiver, io_lib:format("~p", [ClientName])},
```
Vagabond, line 139: don't use io_lib format here at all, just return the values, you're printing the term in format_sc_list

vihu: Thanks. Updated in dbd61a4

PR #453, opened 2020-05-18
Add to_json support for transactions and blocks

madninja:
This leaves out all the `signature` like fields from the resulting json. When we need this we can add an option to the to_json opts to enable various additional fields


vihu:
This seems to be missing jsons for `blockchain_txn_update_gateway_oui_v1` and `blockchain_txn_bundle_v1`. I know we don't have either of them on chain but maybe have it in for completeness?

madninja:
> This seems to be missing jsons for `blockchain_txn_update_gateway_oui_v1` and `blockchain_txn_bundle_v1`. I know we don't have either of them on chain but maybe have it in for completeness?

Fixed in 46dbfb0b

PR #460, opened 2020-05-21
Add blockchain_follower

madninja:


src/blockchain_follower.erl
```diff
@@ -0,0 +1,138 @@
...
+    {noreply, State}.
+
+handle_info({blockchain_event, From, Event}, State=#state{}) ->
+    %% Received a sync blockchain_event. Dispatch as if async and then
+    %% acknowledge once it returns
+    Result = handle_info({blockchain_event, Event}, State),
+    blockchain_event:acknowledge(From),
+    Result;
+
+handle_info({blockchain_event, {new_chain, Chain}}, State=#state{follower_mod=FollowerMod}) ->
```
vihu, line 75: Has the `new_chain` event trumped over the `integrated_genesis_block` event? That seems to be missing from here...

madninja: I don't think so (cc @Vagabond ) but block 1 is loaded as part of the catch-up in the next add_block

Vagabond: Do we even need `integrated_genesis_block` anymore?

madninja:
Got a verbal 👍 from @Vagabond

PR #462, opened 2020-05-22
centralize gc rescue handling

evanmcc:
currently when we add a gc or something that needs to be deterministically applied to the ledger, we have to add it in several places.  this PR moves all that handling into one place, inside the block absorb, so it's more transparent to other parts of the code.

it also moves rescue handling into the absorb, since there's no reason for anything outside of block absorption to know about it now that we have `blockchain_block:is_rescue/1`.


Vagabond:
Do we still want to do this?

evanmcc:
I feel like we do.  I'll attempt to rebase sometime next week.

PR #465, opened 2020-05-26
optional decompression in fingerprints

evanmcc:
compression has done a number on the validity of fingerprints.  until snapshots are enabled and we can switch to that for ledger convergence information, we may need the stuff in here to diagnose what's going on when there is drift.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -484,7 +484,9 @@ raw_fingerprint(#ledger_v1{mode = Mode} = Ledger, Extended) ->
         Filter = ?BC_UPGRADE_NAMES,
         DefaultVals = cache_fold(
                         Ledger, DefaultCF,
-                        %% these are nonsense, ignore them
+                        %% if any of these are in the CF, it's a
+                        %% result of an old, fixed bug, they're safe
+                        %% to ignore.
                         fun({<<"$block_", _/binary>>, _}, Acc) ->
                                 Acc;
                            ({K, _} = X, Acc) ->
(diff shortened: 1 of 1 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -484,22 +484,48 @@ raw_fingerprint(#ledger_v1{mode = Mode} = Ledger, Extended) ->
         Filter = ?BC_UPGRADE_NAMES,
         DefaultVals = cache_fold(
                         Ledger, DefaultCF,
-                        fun({K, _} = X, Acc) ->
+                        %% these are nonsense, ignore them
```
Vagabond, line 487: Can you elaborate on why they're nonsense?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -484,22 +484,48 @@ raw_fingerprint(#ledger_v1{mode = Mode} = Ledger, Extended) ->
...
+                             {EntriesCF, blockchain_ledger_entry_v1},
+                             {DCEntriesCF, blockchain_ledger_data_credits_entry_v1},
+                             {HTLCsCF, blockchain_ledger_htlc_v1},
+                             {PoCsCF, t2b},
+                             {SecuritiesCF, blockchain_ledger_security_entry_v1},
+                             {RoutingCF, blockchain_ledger_routing_v1},
+                             {SCsCF, blockchain_ledger_state_channel_v1},
+                             {SubnetsCF, undefined}
+                            ]];
+                _ ->
```
Vagabond, line 519: Shouldn't we use the decompressing fun here too?

Vagabond:
I feel like we should just always decompress, otherwise we'll never ger rid of fingerprint drift

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: they should be deleted when they were no longer useful; but instead were left due to a bug.  that bug is fixed, but the data remain.  if you did a full sync using post snapshot code they would not be there.

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: I dunno, it seems really expensive.  I don't know how often we call this, if it's affordable or not.

Vagabond: But since we have mixed compression now, we can never get a sane fingerprint otherwise?

Vagabond: Another option would be to run a proper "compressing" upgrade hook?

PR #469, opened 2020-05-28
Remove most eqc noise

vihu:
I have commented out the file writes and `io:formats` here, preferring not to remove them entirely for local usage. Also fixed a bad function call.

### DIFF
--- a/eqc/path_v2_eqc.erl
+++ b/eqc/path_v2_eqc.erl
@@ -40,18 +40,17 @@ prop_path_check() ->
 
                 PathLength = length(Path),
 
-                B58Path = #{libp2p_crypto:bin_to_b58(TargetPubkeyBin) => [[libp2p_crypto:bin_to_b58(P) || P <- Path]]},
+                %% B58Path = #{libp2p_crypto:bin_to_b58(TargetPubkeyBin) => [[libp2p_crypto:bin_to_b58(P) || P <- Path]]},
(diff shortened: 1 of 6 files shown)


eqc/path_v2_eqc.erl
```diff
@@ -40,18 +40,17 @@ prop_path_check() ->
...
-
-                case length(Path) > 1 of
-                    true ->
-                        ok = file:write_file("/tmp/paths_js", io_lib:fwrite("~p.\n", [B58Path]), [append]),
-                        ok = file:write_file("/tmp/paths_target", io_lib:fwrite("~p: ~p.\n", [name(TargetPubkeyBin), HumanPath]), [append]);
-                    false ->
-                        ok = file:write_file("/tmp/paths_beacon", io_lib:fwrite("~p: ~p.\n", [name(TargetPubkeyBin), HumanPath]), [append])
-                end,
-
+                %% io:format("Time: ~p\t Path: ~p~n", [erlang:convert_time_unit(Time, microsecond, millisecond), HumanPath]),
```
evanmcc, line 45: we might be able to move local and have it deleted on success instead of just commenting it out, but no one probably has the time right now

vihu: honestly, I just wanted something quick and dirty to alleviate the /tmp on kite concerns

PR #471, opened 2020-05-28
speed up starting

evanmcc:
currently ledger_at only adds a snapshot when we ask for a particular height.  This can lead to certain ledger states being computed over and over and over again, since there is no snapshot at that specific height.  It also means that parallel validations can't share very much work.

this PR changes the behavior to snapshot at every visited height, and to seek backwards first, so that if we have some intermediate state computed, we can do the least work possible.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -473,7 +473,7 @@ fold_blocks(Chain0, DelayedHeight, DelayedLedger, Height) ->
         lists:foldl(
           fun(_, Acc) when is_tuple(Acc) ->
                   Acc;
-             (H, none) when H == DelayedHeight ->
+             (H, none) when H == (DelayedHeight+1) ->
                   {DelayedHeight, DelayedLedger};
              (H, none) ->
                   case blockchain_ledger_v1:has_snapshot(Height, DelayedLedger) of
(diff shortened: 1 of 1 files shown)


src/blockchain.erl
```diff
@@ -479,6 +496,13 @@ fold_blocks(Chain0, DelayedHeight, DelayedLedger, Height) ->
                               ok = blockchain_ledger_v1:maybe_gc_pocs(Chain1, Ledger0),
```
evanmcc, line 496: we could possibly skip more work here by checking if someone has raced us to this shapshot, but it adds a lot of complexity.

Vagabond: pre-calculating on load seems to prevent races

src/blockchain.erl
```diff
@@ -468,6 +468,23 @@ ledger_at(Height, Chain0, ForceRecalc) ->
...
 fold_blocks(Chain0, DelayedHeight, DelayedLedger, Height) ->
+    %% to minimize work, check backwards for snapshots
+    {HighestSnapHeight, HighestLedger} =
+        lists:foldl(
+          fun(_, Acc) when is_tuple(Acc) ->
+                  Acc;
+             (H, none) when H == (DelayedHeight+1) ->
+                  {DelayedHeight, DelayedLedger};
+             (H, none) ->
+                  case blockchain_ledger_v1:has_snapshot(Height, DelayedLedger) of
```
Vagabond, line 479: This should be H here, not Height?

evanmcc: yes, copy-paste error.

PR #474, opened 2020-05-30
Improve sync timer management, logging and stream timeouts

Vagabond:
I think snapshot syncs are having stream timeouts (they close normal even though the other side is sending us the snapshot). This tries to improve that.

### DIFF
--- a/src/handlers/blockchain_snapshot_handler.erl
+++ b/src/handlers/blockchain_snapshot_handler.erl
@@ -46,16 +46,18 @@ server(Connection, Path, _TID, Args) ->
 %% ------------------------------------------------------------------
 %% libp2p_framed_stream Function Definitions
 %% ------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/handlers/blockchain_snapshot_handler.erl
```diff
@@ -46,16 +46,18 @@ server(Connection, Path, _TID, Args) ->
 %% ------------------------------------------------------------------
 %% libp2p_framed_stream Function Definitions
 %% ------------------------------------------------------------------
-init(client, _Conn, [Hash, Height, Chain]) ->
+init(client, Conn, [Hash, Height, Chain]) ->
```
evanmcc, line 49: I think that it'd be good to increase this timeout (perhaps by not as much) for the sync/ffwd handler too.

Vagabond: I think we should do that in a followup pr if this works.

PR #475, opened 2020-05-30
Don't crash if we don't have a genesis block on startup

Vagabond:
Just small fixes.

I think there's a bug where we go onto the snapshot path even if honor_quick_sync is false, as well.


Vagabond:
oh, so we have `{honor_quick_sync, boolean()}` *and* `{sync_mode, full | quick}`, we should probably not have so much redundancy.

PR #486, opened 2020-06-03
rough draft of better snapshot upgrade code

evanmcc:


Vagabond:
This appears to do the thing if combined with proto#35

Vagabond:
is there more this needs?

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -167,7 +212,7 @@ serialize(Snapshot) ->
 serialize(Snapshot, BlocksP) ->
     Snapshot1 = case BlocksP of
                     blocks -> Snapshot;
-                    noblocks -> Snapshot#blockchain_snapshot_v1{blocks = []}
```
Vagabond, line 170: Should we mediate the shapshot version based on the oracle chain var?

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -69,8 +114,8 @@
 
          blocks :: [blockchain_block:block()],
 
-         oracle_price :: non_neg_integer(),
```
Vagabond, line 72: Is there a more extensible way to add more fields that will be more compatible in future?

PR #491, opened 2020-06-09
Fix price oracle calculations

Vagabond:


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1514,7 +1514,7 @@ select_prices_by_time(Start, End, Entry, Acc) ->
     P = blockchain_ledger_oracle_price_entry:price(Entry),
 
     if
-        T >= Start andalso T =< End -> Acc#{ PK => P };
+        T >= End andalso T =< Start -> Acc#{ PK => P };
```
vihu, line 1517: :trollface: nice catch

PR #492, opened 2020-06-09
txn fees

andymck:
Adds support for transaction fees and staking fees.
Linked PRs:

https://github.com/helium/proto/pull/39


include/blockchain_txn_fees.hrl
```diff
@@ -0,0 +1,46 @@
+
+-define(DC_PRICE, 0.00001).
+-define(BONES_PER_HNT, 100000000).
+
+%% Fees below are quoted in USD
+-define(STAKING_FEES, [
```
Vagabond, line 6: These should probably be chain variables?

src/transactions/v1/blockchain_txn_oui_v1.erl
```diff
@@ -205,8 +205,19 @@ absorb(Txn, Chain) ->
                 {false, LedgerOUI} ->
                     {error, {invalid_oui, OUI, LedgerOUI}};
                 true ->
-                    case blockchain_ledger_v1:debit_fee(ActualPayer, Fee + StakingFee, Ledger) of
+                    case blockchain_ledger_v1:debit_fee(ActualPayer, TxnFee + StakingFee, Ledger) of
```
Vagabond, line 208: Would it be better to do the implicit burn in the debit_fee function so we can keep that cleanly isolated?

src/transactions/blockchain_txn.erl
```diff
@@ -581,11 +581,11 @@ dc_to_hnt(DCAmount, Chain)->
     Ledger = blockchain:ledger(Chain),
     case blockchain_ledger_v1:current_oracle_price(Ledger) of
         {ok, 0} ->
-            0;
+            {ok, 0};
         {ok, OracleHNTPrice} ->
             DCInUSD = DCAmount * ?DC_PRICE,
             %% need to put USD amount into 1/100_000_000th cents, same as oracle price
-            trunc((DCInUSD * 100000000 / OracleHNTPrice) * ?BONES_PER_HNT)
+            {ok, trunc((DCInUSD * 100000000 / OracleHNTPrice) * ?BONES_PER_HNT)}
```
mrallen1: probably want to use `div` here right?

src/transactions/v1/blockchain_txn_oui_v1.erl
andymck: yeah good idea , done

src/transactions/blockchain_txn.erl
andymck: well DCInUSD at this point is likely going to be a float

src/transactions/v1/blockchain_txn_create_htlc_v1.erl
```diff
@@ -237,32 +250,28 @@ is_valid(Txn, Chain) ->
...
         {error, _}=Error ->
             Error;
         {ok, _Entry} ->
-            case blockchain_ledger_v1:debit_fee(Payer, Fee, Ledger) of
+            AreFeesEnabled = blockchain_ledger_v1:txn_fees_active(Ledger),
+            case blockchain_ledger_v1:debit_fee(Payer,TxnFee, Ledger, AreFeesEnabled) of
                 {error, _Reason}=Error ->
                     Error;
                 ok ->
-                    case blockchain_ledger_v1:debit_account(Payer, Amount, Nonce, Ledger) of
```
Vagabond, line 252: Removing this debit account means that the HTLC essentially double spends tokens because they don't move into the HTLC, they're copied.

src/transactions/v1/blockchain_txn_state_channel_open_v1.erl
```diff
@@ -106,22 +124,23 @@ is_valid(Txn, Chain) ->
...
     ExpireWithin = ?MODULE:expire_within(Txn),
     Nonce = ?MODULE:nonce(Txn),
-    case blockchain_ledger_v1:debit_dc(Owner, Nonce, Ledger) of
-        {error, _}=Error ->
-            Error;
-        ok ->
-            blockchain_ledger_v1:add_state_channel(ID, Owner, ExpireWithin, Nonce, Ledger)
+    TxnFee = ?MODULE:fee(Txn),
+    %% TODO - confirm 'Owner' is the account which pays the fee
+    case blockchain_ledger_v1:debit_fee(Owner, TxnFee, Ledger, AreFeesEnabled) of
```
Vagabond, line 141: Debit DC is not the same as debit fee, we need to do both

src/transactions/v1/blockchain_txn_create_htlc_v1.erl
andymck: well spotted, i must have fat fingered this whilst copying & pasting

src/transactions/blockchain_txn.erl
Vagabond: I think this turned out to be unnecessary and got removed?

src/blockchain.erl
```diff
@@ -1646,7 +1646,7 @@ load_genesis(Dir) ->
                       PayerB58::string() | undefined,
                       Fee::pos_integer(),
                       StakingFee::non_neg_integer()) -> {ok, binary()}.
-add_gateway_txn(OwnerB58, PayerB58, Fee, StakingFee) ->
+add_gateway_txn(OwnerB58, PayerB58, _Fee, _StakingFee) ->
```
Vagabond, line 1649: Should this just be removed?

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -161,7 +160,7 @@ snapshot(Ledger0, Blocks) ->
                leading_hash = <<>>,
 
                current_height = CurrHeight,
-               transaction_fee = TransactionFee,
+               transaction_fee =  0, %% NOTE transaction fee is no longer supported, defaulting to zero or safe to remove??
```
Vagabond, line 163: Default it to 0, we can remove it in the next snapshot version.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1429,6 +1418,130 @@ maybe_gc_scs(Ledger) ->
...
+%%--------------------------------------------------------------------
+-spec dc_to_hnt(non_neg_integer(), ledger()) -> {ok, non_neg_integer()}.
+dc_to_hnt(DCAmount, Ledger)->
+    case ?MODULE:current_oracle_price(Ledger) of
+        {ok, 0} ->
+            {ok, 0};
+        {ok, OracleHNTPrice} ->
+            DCInUSD = DCAmount * ?DC_PRICE,
+            %% need to put USD amount into 1/100_000_000th cents, same as oracle price
+            {ok, trunc((DCInUSD * 100000000 / OracleHNTPrice) * ?BONES_PER_HNT)}
```
Vagabond, line 1527: We should try to use only integer division here with `div` if possible to avoid floating point demons

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1670,36 +1803,43 @@ debit_dc(Address, Nonce, Ledger) ->
...
+                    {error, {insufficient_dc_balance, Amount, Balance}}
             end
     end.
 
 -spec debit_fee(Address :: libp2p_crypto:pubkey_bin(), Fee :: non_neg_integer(), Ledger :: ledger()) -> ok | {error, any()}.
-debit_fee(_Address, 0,_Ledger) ->
+debit_fee(_Address, Fee,_Ledger) ->
+    debit_fee(_Address, Fee,_Ledger, false).
+-spec debit_fee(Address :: libp2p_crypto:pubkey_bin(), Fee :: non_neg_integer(), Ledger :: ledger(), UseImplicitBurn :: boolean()) -> ok | {error, any()}.
+debit_fee(_Address, 0,_Ledger, _UseImplicitBurn) ->
```
Vagabond, line 1821: Under what circumstances do we want to set UseImplicitBurn to false?

src/transactions/v1/blockchain_txn_assert_location_v1.erl
```diff
@@ -363,16 +401,17 @@ absorb(Txn, Chain) ->
...
-
-    case blockchain_ledger_v1:debit_fee(ActualPayer, Fee + StakingFee, Ledger) of
+    case blockchain_ledger_v1:debit_fee(ActualPayer, Fee + StakingFee, Ledger, AreFeesEnabled) of
         {error, _Reason}=Error ->
             Error;
         ok ->
             blockchain_ledger_v1:add_gateway_location(Gateway, Location, Nonce, Ledger)
     end,
 
+    %% TODO - the flow here allows the hex to be updated even if the gateway location is not updated
```
Vagabond, line 412: Yes, this seems wrong, we should hoist the following logic into the ok clause above

src/transactions/v1/blockchain_txn_oui_v1.erl
```diff
@@ -220,13 +216,42 @@ absorb(Txn, Chain) ->
...
+-spec calculate_staking_fee(txn_oui(), blockchain:blockchain()) -> non_neg_integer().
+calculate_staking_fee(Txn, Chain) ->
+    Ledger = blockchain:ledger(Chain),
+    calculate_staking_fee(Txn, Ledger, blockchain_ledger_v1:txn_fees_active(Ledger)).
+
+-spec calculate_staking_fee(txn_oui(), blockchain_ledger_v1:ledger(), boolean()) -> non_neg_integer().
+calculate_staking_fee(_Txn, _Ledger, false) ->
+    ?LEGACY_STAKING_FEE;
+calculate_staking_fee(Txn, Ledger, true) ->
+    %% get total price of txn in USD and divide by DC price
```
Vagabond, line 249: Should we just make all fees denominated in DC for simplicity?

src/transactions/v1/blockchain_txn_routing_v1.erl
```diff
@@ -49,46 +51,50 @@
...
-request_subnet(OUI, Owner, SubnetSize, Fee, Nonce) ->
+-spec request_subnet(non_neg_integer(), libp2p_crypto:pubkey_bin(), pos_integer(), non_neg_integer()) -> txn_routing().
+request_subnet(OUI, Owner, SubnetSize, Nonce) ->
     #blockchain_txn_routing_v1_pb{
        oui=OUI,
        owner=Owner,
        update={request_subnet, SubnetSize},
-       fee=Fee,
+       fee=?LEGACY_TXN_FEE,
+       staking_fee=0,  %% routing did not implement staking fee until txn fees, so has to default to zero
```
Vagabond, line 97: We havent done any of these on chain yet, so this should be ok

src/transactions/v1/blockchain_txn_state_channel_open_v1.erl
```diff
@@ -106,22 +124,31 @@ is_valid(Txn, Chain) ->
...
     Ledger = blockchain:ledger(Chain),
+    AreFeesEnabled = blockchain_ledger_v1:txn_fees_active(Ledger),
     ID = ?MODULE:id(Txn),
     Owner = ?MODULE:owner(Txn),
     ExpireWithin = ?MODULE:expire_within(Txn),
     Nonce = ?MODULE:nonce(Txn),
-    case blockchain_ledger_v1:debit_dc(Owner, Nonce, Ledger) of
-        {error, _}=Error ->
+    TxnFee = ?MODULE:fee(Txn),
+    %% TODO - confirm 'Owner' is the account which pays the fee
```
Vagabond, line 140: Confirmed

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -865,6 +865,47 @@ validate_var(?price_oracle_height_delta, Value) ->
...
+    case Value of
+        true -> ok;
+        false -> ok;
+        _ -> throw({error, {invalid_txn_fees, Value}})
+    end;
+
+validate_var(?staking_keys, Value) ->
+    case Value of
+        [] -> throw({error, {invalid_staking_keys, Value}});
+        X when is_list(X) -> ok;
```
Vagabond, line 879: I think you'll want to look at how the oracle keys are stored.

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -865,6 +865,47 @@ validate_var(?price_oracle_height_delta, Value) ->
...
+        _ -> throw({error, {invalid_staking_keys, Value}})
+    end;
+
+%% TODO finalize max price for each below
```
Vagabond, line 883: We should document what denomination these are in here.

Vagabond:
We should be checking the fee is at *least* the expected fee, but it can be over it (to allow for a fee market later).

include/blockchain_vars.hrl
```diff
@@ -311,3 +311,16 @@
...
 %% ------------------------------------------------------------------
+
+
+%% ------------------------------------------------------------------
+%% transaction fee vars
+-define(txn_fees, txn_fees).
+-define(staking_keys, staking_keys).
+-define(staking_fee_txn_oui_v1, staking_fee_txn_oui_v1).
+-define(staking_fee_txn_oui_v1_per_address, staking_fee_txn_oui_v1_per_address).
+-define(staking_fee_txn_routing_v1, staking_fee_txn_routing_v1).
```
Vagabond, line 322: I think this is the same as the per_address?

src/transactions/v1/blockchain_txn_oui_v1.erl
mrallen1: I think that's a really good idea

Vagabond: check out #499

include/blockchain_vars.hrl
andymck: Maybe?  As its a sep txn I had  a specific support for it

src/blockchain.erl
andymck: yeah I had been removing any specific passing of fee and staking fees.  Looks like i missed this one so should be fine for it to be removed

src/ledger/v1/blockchain_ledger_v1.erl
andymck: UseImplicitBurn would be set to false if the chain var is not present or active.

andymck: that way we make no implicit burn attempt

include/blockchain_vars.hrl
Vagabond: I think since you're buying the same resource it can use the same fee, and it has been changed on #499

src/transactions/v1/blockchain_txn_vars_v1.erl
andymck: k, i see it will update along same lines

src/blockchain.erl
andymck: done

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
andymck: removed 'NOTES', all default to zero

src/transactions/v1/blockchain_txn_assert_location_v1.erl
andymck: done

src/transactions/v1/blockchain_txn_vars_v1.erl
andymck: Done

src/ledger/v1/blockchain_ledger_v1.erl
Vagabond: I guess that makes sense, it reads more like an option than a mode we're in though.

include/blockchain_txn_fees.hrl
```diff
@@ -0,0 +1,10 @@
+
+-define(DC_TO_USD, 0.00001).
+-define(USD_TO_DC, 100000).
+-define(BONES_PER_HNT, 100000000).
+-define(ORACLE_PRICE_SCALING_FACTOR, 100000000).
+-define(LEGACY_STAKING_FEE, 1).
+-define(LEGACY_TXN_FEE, 0).
+
+%% Macros
+-define(fee(Txn), trunc(byte_size(blockchain_txn:serialize(Txn)) / 24)).
```
Vagabond, line 10: The code for the state channels looks like this:

```erlang
  calculate_dc_amount(Ledger, PayloadSize) ->
      case blockchain_ledger_v1:config(?dc_payload_size, Ledger) of
          {ok, DCPayloadSize} ->
              case PayloadSize =< DCPayloadSize of
                  true ->
                      1;
                  false ->
                      erlang:ceil(PayloadSize/DCPayloadSize)
              end;
          _ ->
              {error, dc_payload_size_not_set}
      end.
```

I don't know how important that is that they're the same though?

src/ledger/v1/blockchain_ledger_v1.erl
andymck: Well in my head it is an option as its value will be txn specific, for instance whilst we are syncing blocks its value can alternate depending on whether blocks come in out of order

andymck: I can rename UseImplicitBurn, that name prob does make it sound like a mode if you read it out of context from the caller

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1721,6 +1855,34 @@ check_dc_balance(Address, Amount, Ledger) ->
...
 
+
+-spec check_dc_or_hnt_balance(Address :: libp2p_crypto:pubkey_bin(), Amount :: non_neg_integer(), Ledger :: ledger(), boolean()) -> ok | {error, any()}.
+check_dc_or_hnt_balance(_Address, 0, _Ledger, _IsFeesEnabled) ->
+    ok;
+check_dc_or_hnt_balance(Address, Amount, Ledger, IsFeesEnabled) ->
+    case ?MODULE:find_dc_entry(Address, Ledger) of
+        {error, dc_entry_not_found} ->
+            {ok, AmountInHNT} = ?MODULE:dc_to_hnt(Amount, Ledger),
+            ct:pal("~p Dc -> ~p HNT", [Amount, AmountInHNT]),
```
Vagabond, line 1866:
```suggestion

```

include/blockchain_txn_fees.hrl
andymck: @Vagabond sorry but i am not following the context of the above ?

Vagabond: So this is how we calculate how many DCs a packet payload is, I'm pointing out that your calculation is different (specifically the use of ceil()) and uses a hardcoded 24 bytes instead of the `?dc_payload_size` from the ledger.

Vagabond: You could make this:

```
-define(fee(Txn, Ledger), blockchain_state_channel_utils:calculate_dc_amount(Ledger, byte_size(blockchain_txn:serialize(Txn)))).
```

Vagabond: And if we had the ledger, we could apply the multiplier here as well.

andymck: k, got it now.  as the fee macro is shared and the use case above is specific to SCs, i have used calculate_dc_amount/2 directly in the calculate_fee function of state channels open.

src/blockchain.erl
```diff
@@ -1668,7 +1665,7 @@ add_gateway_txn(OwnerB58, PayerB58, Fee, StakingFee) ->
                      StakingFee::pos_integer(),
                      Fee::pos_integer()
                     ) -> {ok, binary()}.
-assert_loc_txn(H3String, OwnerB58, PayerB58, Nonce, StakingFee, Fee) ->
+assert_loc_txn(H3String, OwnerB58, PayerB58, Nonce, _StakingFee, _Fee) ->
```
vihu, line 1668: Remove `_StakingFee, _Fee` if they are unused. Also would have to update the usage of this fun in `blockchain_cli_txn.erl`

Vagabond, line 1668: Redundant parameters?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1617,6 +1721,27 @@ debit_account(Address, Amount, Nonce, Ledger) ->
...
+                true ->
+                    Entry1 = blockchain_ledger_entry_v1:new(
+                        blockchain_ledger_entry_v1:nonce(Entry),
+                        (Balance - Fee)
+                    ),
+                    Bin = blockchain_ledger_entry_v1:serialize(Entry1),
+                    EntriesCF = entries_cf(Ledger),
+                    cache_put(Ledger, EntriesCF, Address, Bin);
+                false ->
+                    {error, {insufficient_balance, Fee, Balance}}
```
vihu, line 1741: Make this error atom different than the one in debit_account. Something like: `insufficient_balance_for_fee` or whatever

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1446,6 +1434,122 @@ maybe_gc_scs(Ledger) ->
...
 
+%%--------------------------------------------------------------------
+%% @doc  get staking server keys from chain var
+%% @end
+%%--------------------------------------------------------------------
+-spec staking_keys(Ledger :: ledger()) -> not_found | [libp2p_crypto:pubkey_bin()].
+staking_keys(Ledger)->
+    case blockchain:config(?staking_keys, Ledger) of
+        {error, not_found} -> not_found;
+        {ok, V} -> V
```
Vagabond, line 1445: I think you're still missing a deserialize here?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1446,6 +1434,122 @@ maybe_gc_scs(Ledger) ->
...
+%%--------------------------------------------------------------------
+%% @doc
+%% converts DC to HNT bones
+%% @end
+%%--------------------------------------------------------------------
+-spec dc_to_hnt(non_neg_integer(), ledger() | pos_integer()) -> {ok, non_neg_integer()}.
+dc_to_hnt(DCAmount, OracleHNTPrice) when is_integer(OracleHNTPrice) ->
+    DCInUSD = DCAmount * ?DC_TO_USD,
+    %% need to put USD amount into 1/100_000_000th cents, same as oracle price
+    {ok, trunc((DCInUSD * 100000000 / OracleHNTPrice) * ?BONES_PER_HNT)};
```
Vagabond, line 1527: Should this be ceil, trunc or round?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1667,10 +1792,9 @@ credit_dc(Address, Amount, Ledger) ->
 
 -spec debit_dc(Address :: libp2p_crypto:pubkey_bin(),
                Nonce :: non_neg_integer(),
+               Amount :: non_neg_integer(),
                Ledger :: ledger()) -> ok | {error, any()}.
-debit_dc(Address, Nonce, Ledger) ->
-    {ok, Fee} = ?MODULE:transaction_fee(Ledger),
-
+debit_dc(Address, Nonce, Amount, Ledger) ->
```
vihu, line 1797: Sorry but I think this was wrong before this work, having an empty dc entry when `dc_entry_not_found` hits is okay but the second error clause should short circuit any further logic entirely imo.

include/blockchain_vars.hrl
```diff
@@ -311,3 +311,15 @@
 %% (Will also affect what prices get dropped from the cached list of prices.)
 -define(price_oracle_price_scan_max, price_oracle_price_scan_max).
 %% ------------------------------------------------------------------
+
+
+%% ------------------------------------------------------------------
+%% transaction fee vars, denominated in DC
```
vihu, line 317: Add some single line comments to indicate what we expect these vars to initially be

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1446,6 +1434,122 @@ maybe_gc_scs(Ledger) ->
...
+        {error, not_found} -> false;
+        {ok, V} -> V
+    end.
+
+%%--------------------------------------------------------------------
+%% @doc  get staking fee chain var value for OUI
+%% or return default
+%% @end
+%%--------------------------------------------------------------------
+-spec staking_fee_txn_oui_v1(Ledger :: ledger()) -> non_neg_integer().
```
vihu, line 1464:
```suggestion
-spec staking_fee_txn_oui_v1(Ledger :: ledger()) -> pos_integer().
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1446,6 +1434,122 @@ maybe_gc_scs(Ledger) ->
...
+        {error, not_found} -> 0;
+        {ok, V} -> V
+    end.
+
+%%--------------------------------------------------------------------
+%% @doc  get staking fee chain var value for add gateway
+%% or return default
+%% @end
+%%--------------------------------------------------------------------
+-spec staking_fee_txn_add_gateway_v1(Ledger :: ledger()) -> non_neg_integer().
```
vihu, line 1488:
```suggestion
-spec staking_fee_txn_add_gateway_v1(Ledger :: ledger()) -> pos_integer().
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1446,6 +1434,122 @@ maybe_gc_scs(Ledger) ->
...
+        {error, not_found} -> 1;
+        {ok, V} -> V
+    end.
+
+%%--------------------------------------------------------------------
+%% @doc  get payment txn's txn fee multiplier
+%% or return default
+%% @end
+%%--------------------------------------------------------------------
+-spec payment_txn_fee_multiplier(Ledger :: ledger()) -> non_neg_integer().
```
vihu, line 1500:
```suggestion
-spec payment_txn_fee_multiplier(Ledger :: ledger()) -> pos_integer().
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1446,6 +1434,122 @@ maybe_gc_scs(Ledger) ->
...
+    case blockchain:config(?payment_txn_fee_multiplier, Ledger) of
+        {error, not_found} -> 1;
+        {ok, V} -> V
+    end.
+%%--------------------------------------------------------------------
+%% @doc  get staking fee chain var value for add gateway
+%% or return default
+%% @end
+%%--------------------------------------------------------------------
+-spec staking_fee_txn_assert_location_v1(Ledger :: ledger()) -> non_neg_integer().
```
vihu, line 1511:
```suggestion
-spec staking_fee_txn_assert_location_v1(Ledger :: ledger()) -> pos_integer().
```

src/ledger/v1/blockchain_ledger_v1.erl
vihu, line 1527: > `100000000 / OracleHNTPrice`
Isn't that 10^8 ?ORACLE_PRICE_SCALING_FACTOR?

vihu: Also, can we add simple eunits for this?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1446,6 +1434,122 @@ maybe_gc_scs(Ledger) ->
...
+            dc_to_hnt(DCAmount, OracleHNTPrice)
+    end.
+
+%%--------------------------------------------------------------------
+%% @doc
+%% converts HNT bones to DC
+%% @end
+%%--------------------------------------------------------------------
+-spec hnt_to_dc(non_neg_integer(), ledger() | pos_integer()) -> {ok, non_neg_integer()}.
+hnt_to_dc(HNTAmount, OracleHNTPrice) when is_integer(OracleHNTPrice) ->
```
vihu, line 1542: An eunit for this would be useful as well

src/transactions/v1/blockchain_txn_add_gateway_v1.erl
```diff
@@ -9,30 +9,31 @@
...
     sign_request/2,
     sign_payer/2,
     is_valid_gateway/1,
     is_valid_owner/1,
     is_valid_payer/1,
+    is_valid_staking_key/2,
     is_valid/2,
     absorb/2,
-    calculate_staking_fee/1,
+    calculate_fee/2, calculate_fee/3, calculate_staking_fee/2, calculate_staking_fee/3,
```
Vagabond, line 36: Do we need to export the calculate fee/3 functions?

src/transactions/v1/blockchain_txn_add_gateway_v1.erl
```diff
@@ -144,6 +147,57 @@ staking_fee(Txn) ->
...
+                  %% no payer signature if there's no payer
+                  ?fee(Txn#blockchain_txn_add_gateway_v1_pb{fee=0, staking_fee=0,
+                                                            owner_signature = <<0:512>>,
+                                                            gateway_signature = <<0:512>>,
+                                                            payer_signature = <<>>});
+              _ ->
+                  ?fee(Txn#blockchain_txn_add_gateway_v1_pb{fee=0, staking_fee=0,
+                                                            owner_signature = <<0:512>>,
+                                                            gateway_signature = <<0:512>>,
+                                                            payer_signature = <<0:512>>}) %% TODO - without setting this to fixed length test is failing...need to double check
```
Vagabond, line 180: Signatures are always 64 bytes, so that's fine.

src/ledger/v1/blockchain_ledger_v1.erl
vihu: Actually, now that I notice it, there are a lot of places where we have the error tuple, "insufficient_balance". Now is the time to clean them up and have verbose error tuples.

src/blockchain.erl
andymck: done ( i am sure i did this already but lost if somewhere )

andymck: yip, fixed

src/ledger/v1/blockchain_ledger_v1.erl
Vagabond: There's an eunit at the end of the module?

vihu: oh my bad

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1738,6 +1872,33 @@ check_dc_balance(Address, Amount, Ledger) ->
             end
     end.
 
+
+-spec check_dc_or_hnt_balance(Address :: libp2p_crypto:pubkey_bin(), Amount :: non_neg_integer(), Ledger :: ledger(), boolean()) -> ok | {error, any()}.
+check_dc_or_hnt_balance(_Address, 0, _Ledger, _IsFeesEnabled) ->
```
vihu, line 1877: I don't get this function, we already have `check_balance` and `check_dc_balance`, if we do need this combo, why not just use those two? I see `check_balance` being used here but presumably you can make this function tinier, no?

src/transactions/v1/blockchain_txn_add_gateway_v1.erl
```diff
@@ -228,51 +282,56 @@ is_valid_payer(#blockchain_txn_add_gateway_v1_pb{payer=PubKeyBin,
...
-                    end
+        {_, _, _, false} ->
+            {error, payer_invalid_staking_key};
+        {true, true, true, true} ->
+            AreFeesEnabled = blockchain_ledger_v1:txn_fees_active(Ledger),
+            StakingFee = ?MODULE:staking_fee(Txn),
+            ExpectedStakingFee = ?MODULE:calculate_staking_fee(Txn, Chain),
+            TxnFee = ?MODULE:fee(Txn),
+            ExpectedTxnFee = ?MODULE:calculate_fee(Txn, Chain),
+            case {(ExpectedTxnFee =< TxnFee orelse not AreFeesEnabled), ExpectedStakingFee == StakingFee} of
```
vihu, line 318: Is this correct? Shouldn't the first short-circuit check be on `AreFeesEnabled`? So something like:
```
            case AreFeesEnabled of
                false ->
                    %% old behavior
                true ->
                    %% check the fees
            end

```

src/transactions/v1/blockchain_txn_assert_location_v1.erl
```diff
@@ -298,20 +303,23 @@ is_valid(Txn, Chain) ->
...
             end,
+            AreFeesEnabled = blockchain_ledger_v1:txn_fees_active(Ledger),
             StakingFee = ?MODULE:staking_fee(Txn),
-            ExpectedStakingFee = ?MODULE:calculate_staking_fee(Chain),
-            case ExpectedStakingFee == StakingFee of
-                false ->
+            ExpectedStakingFee = ?MODULE:calculate_staking_fee(Txn, Chain),
+            TxnFee = ?MODULE:fee(Txn),
+            ExpectedTxnFee = calculate_fee(Txn, Chain),
+            case {(ExpectedTxnFee =< TxnFee orelse not AreFeesEnabled), ExpectedStakingFee == StakingFee} of
```
vihu, line 316: Same question as add_gateway_v1 is_valid

src/ledger/v1/blockchain_ledger_v1.erl
andymck: lets go with ceil and get a few extra micro DCs ?....it worked out well for richard pryor in superman 3

src/ledger/v1/blockchain_ledger_v1.erl
andymck: @vihu not following you here, can you elaborate.  Also as a note, the only caller of this function atm is the state channel open txn and its passing in a zero amount atm.

vihu: My concern isn't with your changes fwiw.

I'm concerned about this piece:
```
    Entry = case ?MODULE:find_dc_entry(Address, Ledger) of
                {error, dc_entry_not_found} ->
                    %% Just create a blank entry if dc_entry_not_found
                    blockchain_ledger_data_credits_entry_v1:new(0, 0);
                {error, _}=Error ->
                    Error;
                {ok, Entry0} ->
                    Entry0
            end,

    case Nonce =:= blockchain_ledger_data_credits_entry_v1:nonce(Entry) + 1 of

```

If the second arm of the case clause is hit, there's no point to do any further logic I would imagine, no?

src/transactions/v1/blockchain_txn_add_gateway_v1.erl
andymck: I dont think we do anymore, i was using them in some tests previously but i think i refactored...will check

src/transactions/v1/blockchain_txn_add_gateway_v1.erl
andymck: TODO removed

src/ledger/v1/blockchain_ledger_v1.erl
andymck: This function is to basically check if the user has sufficient DC balance to cover the txn fees and if not then to automatically check if he can pay the fees instead from his HNT balance.

I think you are right, it can be made simpler, smaller by calling out to check_dc_balance rather than emulating it

test/blockchain_price_oracle_SUITE.erl
```diff
@@ -27,12 +32,91 @@ all() -> [
...
+    txn_fees_pay_with_dc,
+    txn_fees_pay_with_hnt
 ].
 
 %%--------------------------------------------------------------------
 %% TEST CASE SETUP
 %%--------------------------------------------------------------------
+init_per_testcase(TestCase, Config0)  when TestCase == txn_fees_pay_with_hnt;
+                                           TestCase == txn_fees_pay_with_dc ->
+    Config = blockchain_ct_utils:init_base_dir_config(?MODULE, TestCase, Config0),
```
Vagabond, line 45:
```suggestion
    Config = blockchain_ct_utils:init_base_dir_config(?MODULE, TestCase, Config),
```

src/transactions/v1/blockchain_txn_add_gateway_v1.erl
andymck: the case supports both old and new behaviour in a single statement, admittedly it is a lil hard to read but I do think its better than having multiple nested statements ( especially given some of this code is already knee deep in nested cases )

vihu: Fair enough, just wanted to make sure that it does do the right thing.

src/ledger/v1/blockchain_ledger_v1.erl
andymck: it may break a lot of shit

src/ledger/v1/blockchain_ledger_v1.erl
andymck: fixed

include/blockchain_vars.hrl
andymck: Done

src/ledger/v1/blockchain_ledger_v1.erl
andymck: fixed

src/ledger/v1/blockchain_ledger_v1.erl
andymck: fixed

src/transactions/v1/blockchain_txn_add_gateway_v1.erl
andymck: Removed

src/ledger/v1/blockchain_ledger_v1.erl
vihu: Really? I thought we'd just propagate the error through no? It's not necessary though, may be once the tests are in better shape here we can try it otherwise leave it be.

include/blockchain_txn_fees.hrl
Vagabond: Well, no, my point is that since we're doing basicially the same calculation as we do for state channel packets for transaction fees, we could call that state_channel_utils function in the fee macro and take advantage of the common code, and also make it use the `?dc_payload_size` rather than hardcode the number 24

src/ledger/v1/blockchain_ledger_v1.erl
Vagabond: Yes.

src/ledger/v1/blockchain_ledger_v1.erl
andymck: Actually i have another CH to return meaningful errors from txn mgr.  I'll handle this suggestion as part of that

src/ledger/v1/blockchain_ledger_v1.erl
andymck: fixed

src/ledger/v1/blockchain_ledger_v1.erl
andymck: hmm I tried to reduce the footprint here but tbh the code gets messier.  We either end up pulling the same ledger entry twice or alternatively breaking up the existing check_db_balance function to allow it to take a prefetched entry.  I think what is there now is the better option from a readability and flow perspective

include/blockchain_txn_fees.hrl
andymck: fixed

include/blockchain_vars.hrl
```diff
@@ -311,3 +311,23 @@
...
+%% the staking fee in DC for each OUI
+-define(staking_fee_txn_oui_v1, staking_fee_txn_oui_v1).
+%% the staking fee in DC for each OUI/routing address
+-define(staking_fee_txn_oui_v1_per_address, staking_fee_txn_oui_v1_per_address).
+%% the staking fee in DC for adding a gateway
+-define(staking_fee_txn_add_gateway_v1, staking_fee_txn_add_gateway_v1).
+%% the staking fee in DC for asserting a location
+-define(staking_fee_txn_assert_location_v1, staking_fee_txn_assert_location_v1).
+%% a mutliplier which will be applied to txn fee of all payment txns, in order to make their DC costs meaninful
+-define(payment_txn_fee_multiplier, payment_txn_fee_multiplier).
```
vihu, line 332: Suggestion to rename this to just `txn_fee_multiplier`

Vagabond: :+1:

andymck: done

src/transactions/v1/blockchain_txn_add_gateway_v1.erl
```diff
@@ -228,51 +281,56 @@ is_valid_payer(#blockchain_txn_add_gateway_v1_pb{payer=PubKeyBin,
     PubKey = libp2p_crypto:bin_to_pubkey(PubKeyBin),
     libp2p_crypto:verify(EncodedTxn, Signature, PubKey).
 
+-spec is_valid_staking_key(txn_add_gateway(), blockchain_ledger_v1:ledger())-> boolean().
+is_valid_staking_key(#blockchain_txn_add_gateway_v1_pb{payer=Payer}=_Txn, Ledger) ->
+    case blockchain_ledger_v1:staking_keys(Ledger) of
+        not_found -> true; %% chain var not active, so default to true
```
Vagabond, line 287: So we can delete this var later to deactivate this check?

Vagabond:
We need a test for the staking keys check for add gateways, right?

Vagabond:
> We need a test for the staking keys check for add gateways, right?

I will add this.

Vagabond:
> > We need a test for the staking keys check for add gateways, right?
> 
> I will add this.
Done

PR #496, opened 2020-06-12
WIP for establishing packet offer semantics

vihu:
This enables support for offer/buy semantics when a hotspot sends a packet to the router.
It follows a synchronous path like so: `packet -> offer -> purchase (yes/no) -> send packet (if yes)`

Proto: https://github.com/helium/proto/pull/38


src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -115,41 +135,138 @@ handle_info(_Msg, State) ->
...
+                      Stream :: pid(),
+                      State :: state()) -> state().
+handle_purchase(Purchase, Stream, #state{swarm=Swarm}=State) ->
+    PacketHash = blockchain_state_channel_purchase_v1:packet_hash(Purchase),
+    PurchaseSC = blockchain_state_channel_purchase_v1:sc(Purchase),
+    PurchaseSCID = blockchain_state_channel_v1:id(PurchaseSC),
+    case find_packet(PacketHash, State) of
+        undefined ->
+            %% Drop it, we don't know this packet
+            lager:warning("Dropping purchase, packet_hash: ~p", [PacketHash]),
```
macpie, line 177: maybe drop that line to a debug if this happens too often?

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -73,14 +77,33 @@ packet(Packet, HandlerPid) ->
...
                   end
           end),
     ok.
 
+-spec offer(blockchain_state_channel_offer_v1:offer(), pid()) -> ok.
+offer(Offer, HandlerPid) ->
+    spawn(fun() ->
+                  case blockchain_state_channel_offer_v1:validate(Offer) of
+                      {error, _Reason} ->
+                          lager:warning("offer failed to validate ~p ~p", [_Reason, Offer]);
```
macpie, line 93: We might get a lot of offers maybe we should drop that to  debug also?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -81,7 +81,9 @@
     get_oui_counter/1, set_oui_counter/2, increment_oui_counter/1,
     add_oui/5,
 
-    find_routing/2, find_routing_for_packet/2, find_router_ouis/2,
+    find_routing/2, find_routing_for_packet/2,
+    find_routing_via_eui/3, find_routing_via_devaddr/2,
```
Vagabond, line 85: Do these need to be exported?

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -115,41 +135,138 @@ handle_info(_Msg, State) ->
...
+                    add_sc(BannerSCID, BannerSC, State);
+                OurSC ->
+                    %% We already know about this state channel
+                    %% Validate this banner
+                    case validate_banner(BannerSC, OurSC) of
+                        ok ->
+                            lager:info("successful banner validation, inserting BannerSC, id: ~p", [BannerSCID]),
+                            add_sc(BannerSCID, BannerSC, State);
+                        {error, Reason} ->
+                            lager:error("invalid banner, reason: ~p", [Reason]),
```
Vagabond, line 161: Should still be stored in this case if validate banner can detect double spend attempts.

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -115,41 +135,138 @@ handle_info(_Msg, State) ->
...
+    %% Hold onto the packet, if a matching purchase comes in later
+    PacketHash = blockchain_helium_packet_v1:packet_hash(Packet),
+    add_packet(PacketHash, Packet, State0).
+
+-spec validate_banner(BannerSC :: blockchain_state_channel_v1:state_channel(),
+                      OurSC :: blockchain_state_channel_v1:state_channel()) -> ok | {error, any()}.
+validate_banner(BannerSC, OurSC) ->
+    BannerSCNonce = blockchain_state_channel_v1:nonce(BannerSC),
+    OurSCNonce = blockchain_state_channel_v1:nonce(OurSC),
+    case BannerSCNonce == OurSCNonce + 1 of
```
Vagabond, line 236: It doesn't need to be monotonic, just greater and all the balances should be the same or larger.

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -115,41 +135,138 @@ handle_info(_Msg, State) ->
...
     end.
 
-%% ------------------------------------------------------------------
-%% Internal functions
-%% ------------------------------------------------------------------
+-spec validate_purchase(PurchaseSCID :: binary(),
+                        PurchaseSC :: blockchain_state_channel_v1:state_channel(),
+                        State :: state()) -> ok | {error, any()}.
+validate_purchase(PurchaseSCID, PurchaseSC, State) ->
+    %% TODO: Improve purchase validation
```
Vagabond, line 247: So the nonce should be higher, it should be signed, all the balances should be the same or higher and *our* balance should have gone up by the right amount.

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -115,41 +135,138 @@ handle_info(_Msg, State) ->
...
+
+-spec find_packet(PacketHash :: binary(), State :: state()) -> blockchain_helium_packet_v1:packet() | undefined.
+find_packet(PacketHash, #state{packets=Packets}) ->
+    maps:get(PacketHash, Packets, undefined).
+
+-spec add_packet(PacketHash :: binary(),
+                 Packet :: blockchain_helium_packet_v1:packet(),
+                 State :: state()) -> state().
+add_packet(PacketHash, Packet, #state{packets=Packets}=State) ->
+    State#state{packets=maps:put(PacketHash, Packet, Packets)}.
```
Vagabond, line 264: My inclination is to make this a synchronous protocol and not allow pipelining of offers for simplicity here so we wouldn't need a map, just the "last" packet we offered.

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -169,6 +301,12 @@ find_stream(OUI, #state{streams=Streams}) ->
 add_stream(OUI, Stream, #state{streams=Streams}=State) ->
     State#state{streams=maps:put(OUI, Stream, Streams)}.
 
+-spec add_sc(SCID :: binary(),
+             SC :: blockchain_state_channel_v1:state_channel(),
+             State :: state()) -> state().
+add_sc(SCID, SC, #state{state_channels=SCs}=State) ->
```
Vagabond, line 307: I would suggest you allow for 'sibling' values, where you havd 2 conflicting versions of a state channel. Also these should be in the rocksdb.

src/ledger/v1/blockchain_ledger_v1.erl
vihu: No, I was using one of these in a previous commit, no longer required.

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -200,13 +338,13 @@ send_to_route(Packet, Route, Region, State=#state{swarm=Swarm}) ->
                                                     {ok, NewStream} ->
                                                         unlink(NewStream),
                                                         erlang:monitor(process, NewStream),
-                                                        ok = send_packet(Packet, Swarm, NewStream, Region),
```
Vagabond, line 203: If the hotspot and the router are in the same OUI, we should keep using send_packet here.

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -169,8 +198,37 @@ handle_cast({packet, SCPacket},
...
+
+            %% Since we updated active sc, send new banner
+            ok = send_banner(NewSC, HandlerPid),
+
+            NewState = case find_stream(ClientPubkeyBin, State) of
+                           HandlerPid ->
+                               State;
+                           _ ->
+                               %% We either have an undefined stream or a different one
+                               add_stream(ClientPubkeyBin, HandlerPid, State)
```
Vagabond, line 210: Remember this will invert the client/server role and make everyone sad, so probably do not do this

src/state_channel/blockchain_state_channels_client.erl
vihu: Updated in ec3e46a

src/state_channel/blockchain_state_channels_client.erl
vihu: Have reworked this function quite a bit, definitely needs another look

src/state_channel/blockchain_state_channels_client.erl
vihu: Pushed some stuff here: 98310c3 which changes the map to a list and does enqueue/deque of packets

src/state_channel/blockchain_state_channels_client.erl
vihu: Added some banner validation here: 9088144

vihu:
I'm closing this in favor of https://github.com/helium/blockchain-core/pull/515. I've basically squashed all the work here into a single commit there

PR #504, opened 2020-06-16
gateway limits

evanmcc:
limit how many gateways we're looking at when we might pull a lot, such as in targeting or path-building.

### DIFF
--- a/eqc/path_v4_eqc.erl
+++ b/eqc/path_v4_eqc.erl
@@ -16,11 +16,19 @@ prop_path_check() ->
     ?FORALL({Hash, PathLimit, ChallengerIndex},
             {gen_hash(), gen_path_limit(), gen_challenger_index()},
             begin
+                {ok, GWCache} = blockchain_gateway_cache:start_link(),
                 Ledger = ledger(poc_v8_vars()),
                 application:set_env(blockchain, disable_score_cache, true),
                 {ok, _Pid} = blockchain_score_cache:start_link(),
                 ActiveGateways = blockchain_ledger_v1:active_gateways(Ledger),
                 LedgerVars = ledger_vars(Ledger),
+
+                %% use this to artificially constrain path limit for testing
+                %% PathLimit = 5,
+
+                %% use this to make timings more realistic for pi
+                %% disks, set to 1 or 2
+                application:set_env(blockchain, find_gateway_sim_delay, 0),
 
                 %% Overwrite poc_path_limit for checking generated path limits
                 Vars = maps:put(poc_path_limit, PathLimit, LedgerVars),
@@ -29,7 +37,7 @@ prop_path_check() ->
                 {ChallengerPubkeyBin, _ChallengerLoc} = find_challenger(ChallengerIndex, ActiveGateways),
 
                 {ok, {TargetPubkeyBin, TargetRandState}} = blockchain_poc_target_v3:target(ChallengerPubkeyBin, Hash, Ledger, Vars),
-                {_Time, Path} = timer:tc(fun() ->
+                {Time, Path} = timer:tc(fun() ->
                                                 blockchain_poc_path_v4:build(TargetPubkeyBin,
                                                                              TargetRandState,
                                                                              Ledger,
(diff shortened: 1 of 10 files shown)


src/poc/blockchain_poc_path_v4.erl
```diff
@@ -564,3 +566,10 @@ assign_single_witness_prob(Witnesses) ->
                      end
              end,
              Witnesses).
+
+limit_witnesses(#{?poc_path_witness_limit := Limit}, RandState, Witnesses0) ->
```
vihu, line 570: I _think_ this might be coming in as an atom instead of with a `?`. Can you double check this please?

evanmcc: I don't know what you mean here.

vihu: eh, ignore, it's just a macro and the include is up top, so it's probably fine. And you already ran the eqcs which triggered it

eqc/path_v4_eqc.erl
```diff
@@ -16,27 +16,60 @@ prop_path_check() ->
...
                 {ok, {TargetPubkeyBin, TargetRandState}} = blockchain_poc_target_v3:target(ChallengerPubkeyBin, Hash, Ledger, Vars),
-                {_Time, Path} = timer:tc(fun() ->
+                {Time, Path} = timer:tc(fun() ->
                                                 blockchain_poc_path_v4:build(TargetPubkeyBin,
                                                                              TargetRandState,
                                                                              Ledger,
                                                                              block_time(),
                                                                              Vars)
                                         end),
+                application:set_env(blockchain, collect_stats, false),
```
Vagabond, line 47: why do you set the env var right before checking it?

src/ledger/v1/blockchain_ledger_gateway_v2.erl
```diff
@@ -46,7 +46,7 @@
           hist = erlang:error(no_histogram) :: #{integer() => integer()}, %% sampled rssi histogram
           first_time :: undefined | non_neg_integer(), %% first time a hotspot witnessed this one
           recent_time :: undefined | non_neg_integer(), %% most recent a hotspots witnessed this one
-          time = #{} :: #{integer() => integer()} %% TODO: add time of flight histogram
+          time = [] :: [{integer(), integer()}] %% TODO: add time of flight histogram
```
Vagabond, line 49: Didn't you say this was a bad idea?

evanmcc: that commit got pushed to this branch by accident, I'll rebase it out.

evanmcc: #513 contains the corrected version of this code, fwiw.

eqc/path_v4_eqc.erl
evanmcc: because otherwise the compiler complains.  I can make it more compact, I suppose.

src/poc/blockchain_poc_path_v4.erl
evanmcc: yeah, for correctness reasons, I really want all chain var references to use the macros so we can leverage that checking.

src/poc/blockchain_poc_path_v4.erl
```diff
@@ -111,7 +112,8 @@ next_hop(GatewayBin, Ledger, HeadBlockTime, Vars, RandState, Indices) ->
     case blockchain_ledger_gateway_v2:witnesses(Gateway) of
         W when map_size(W) == 0 ->
             {error, no_witness};
-        Witnesses ->
+        Witnesses0 ->
+            {RandState1, Witnesses} = limit_witnesses(Vars, RandState, Witnesses0),
```
Vagabond, line 116: Don't we need to guard this from being called behind a chain var check?

Vagabond: oh, I see the function clause now.

PR #505, opened 2020-06-17
re-enable APIs for creating new txns which support fee args

andymck:
Re added support for the txn 'new' apis which accept txn and staking fees ( which applicable ).  This is a temp measure until all apps and client which rely on blockchain core are updated to have the fee args removed

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -45,8 +45,7 @@
 
     init_assumed_valid/2,
 
-    add_gateway_txn/4, assert_loc_txn/6,   %% tmp apis
-    add_gateway_txn/2, assert_loc_txn/4,
+    add_gateway_txn/4, assert_loc_txn/6,
 
     add_snapshot/2, get_snapshot/2, find_last_snapshot/1,
     find_last_snapshots/2,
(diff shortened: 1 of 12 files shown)


src/blockchain.erl
```diff
@@ -1673,7 +1686,7 @@ assert_loc_txn(H3String, OwnerB58, PayerB58, Nonce) ->
             end,
     {ok, PubKey, SigFun, _ECDHFun} =  blockchain_swarm:keys(),
     PubKeyBin = libp2p_crypto:pubkey_to_bin(PubKey),
-    Txn = blockchain_txn_assert_location_v1:new(PubKeyBin, Owner, Payer, H3Index, Nonce),
+    Txn = blockchain_txn_assert_location_v1:new(PubKeyBin, Owner, Payer, H3Index, Nonce, StakingFee, Fee),
```
Vagabond, line 1689: Would it not be simpler to call fee/2 and staking_fee/2 on the Txn?

Vagabond:
I'd propose simply using the fee setter functions rather than adding back a ton of new transaction creation functions

andymck:
superseded by https://github.com/helium/blockchain-core/pull/507

PR #512, opened 2020-06-18
Macpie/timing

macpie:


src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -115,61 +165,33 @@ handle_info(_Msg, State) ->
...
+    Q = get_queued_packet(AddressOrOUI, State),
+    State#state{waiting=maps:put(AddressOrOUI, Q ++ [{Packet, Region}], Waiting)}.
 
--spec find_routing(Packet :: blockchain_helium_packet_v1:packet()) -> {ok, [blockchain_ledger_routing_v1:routing(), ...]} | {error, any()}.
+-spec delete_queued_packet(AddressOrOUI :: non_neg_integer() | string(), State :: state()) -> state().
+delete_queued_packet(AddressOrOUI, #state{waiting=Waiting}=State) ->
+    State#state{waiting=maps:remove(AddressOrOUI, Waiting)}.
+
+-spec find_routing(Packet :: blockchain_helium_packet_v1:packet()) -> {ok, [blockchain_ledger_routing_v1:routing()]} | {error, any()}.
 find_routing(Packet) ->
```
Vagabond, line 195: Make this function take the chain, so we can cache the chain upstream?

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -91,9 +91,43 @@ code_change(_OldVsn, State, _Extra) ->
 %% ------------------------------------------------------------------
 %% gen_server message handling
 %% ------------------------------------------------------------------
-handle_cast({packet, Packet, DefaultRouters, Region}, State) ->
-    NewState = handle_packet(Packet, DefaultRouters, Region, State),
-    {noreply, NewState};
+handle_cast({default_routers, Packet, DefaultRouters, Region}, #state{swarm=Swarm}=State0) ->
```
vihu, line 94: Can we reduce this duplication a bit? The only thing which changes is the call to `dial_*_and_send_packet` between these two casts afaict.

src/state_channel/blockchain_state_channels_client.erl
macpie: I move it up a little but it will be hard to cache upstream as the miner lora calls this

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -66,7 +59,14 @@ response(Resp) ->
 
 -spec packet(blockchain_helium_packet_v1:packet(), [string()], atom()) -> ok.
 packet(Packet, DefaultRouters, Region) ->
-    gen_server:cast(?SERVER, {packet, Packet, DefaultRouters, Region}).
+    case find_routing(Packet, blockchain_worker:blockchain()) of
```
Vagabond, line 62: Ideally we'd support padding in a chain to packet/4 and bypass the chain lookup, but we can do that later

vihu: I am thinking of doing that as part of https://github.com/helium/blockchain-core/pull/496

PR #514, opened 2020-06-19
make txn fee code paths more testable. add unit tests for each txn's fee

andymck:
Adds unit test for txn fee and staking fee logic

### DIFF
--- a/include/blockchain_txn_fees.hrl
+++ b/include/blockchain_txn_fees.hrl
@@ -1,4 +1,6 @@
 
+
+
 -define(DC_TO_USD, 0.00001).
 -define(USD_TO_DC, 100000).
 -define(BONES_PER_HNT, 100000000).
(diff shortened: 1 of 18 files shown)


include/blockchain_txn_fees.hrl
```diff
@@ -7,6 +9,23 @@
...
+    ?MODULE:calculate_fee(Txn, Ledger, DCPayloadSize, TxnFeeMultiplier, IsTxnFeesActive)).
+
+
+-define(calculate_fee(Txn, Ledger),
+    TxnFeeMultiplier = blockchain_ledger_v1:txn_fee_multiplier(Ledger),
+    blockchain_utils:calculate_dc_amount(Ledger,
+        byte_size(blockchain_txn:serialize(Txn))) * TxnFeeMultiplier).
+-define(calculate_fee(Txn, Ledger, DCPayloadSize, TxnFeeMultiplier),
+    blockchain_utils:calculate_dc_amount(Ledger,
+        byte_size(blockchain_txn:serialize(Txn)) * TxnFeeMultiplier, DCPayloadSize)).
```
Vagabond, line 26:
```suggestion
        byte_size(blockchain_txn:serialize(Txn)), DCPayloadSize) * TxnFeeMultiplier).
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -849,6 +851,17 @@ config(ConfigName, Ledger) ->
             Error
     end.
 
+config(ConfigName, Ledger, Default) ->
```
vihu, line 854: I would advise against adding this, especially if it's being used at only one place (that I could find). The `dc_payload_size` is already set on the chain (that's where this `config/3` is getting used right?), you should be able to just rely on that fact?

src/transactions/blockchain_txn.erl
```diff
@@ -811,6 +811,35 @@ depends_on(Txn, Txns) ->
 %% ------------------------------------------------------------------
 -ifdef(TEST).
 
+%% txn fee data
+-define(DC_PAYLOAD_SIZE, 24).
+-define(TXN_MULTIPLIER, 5000).
+-define(USD_TO_DC, 100000).
+
+%% staking fees in DC data ( should be using same values as defined in chain vars )
+-define(OUI_STAKING_FEE, 100 * ?USD_TO_DC).
```
vihu, line 820: Are the multipliers for the various staking fees not gonna change? @Vagabond?

vihu:
@andymck I thought we were gonna drive the fee eunits through some common config (json or whatever) so that the js, rust cli side can also utilize the same? Are we not doing that anymore?

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: I agree (if we insist on adding it, it should behave more like `application:get_env/3` and not return a tuple wrapped value when there's a default present).

include/blockchain_txn_fees.hrl
```diff
@@ -7,6 +9,23 @@
 -define(LEGACY_TXN_FEE, 0).
 
 %% Macros
+-define(calculate_fee_prep(Txn, Chain),
+    Ledger = blockchain:ledger(Chain),
+    DCPayloadSize = blockchain_ledger_v1:config(?dc_payload_size, Ledger, 1),
```
evanmcc, line 14: yeah, this won't work with the current code because it's still tuple wrapped when it's present on the chain.

src/transactions/blockchain_txn.erl
Vagabond: These are just test values for when we don't have a ledger

PR #515, opened 2020-06-20
Enable packet offer-purchase semantics

vihu:
This introduces support for offer/buy semantics when a hotspot sends a packet to the router.
It follows a synchronous path like so: `packet -> offer -> purchase (yes/no) -> send packet (if yes)`

Proto: https://github.com/helium/proto/pull/38


src/state_channel/v1/blockchain_state_channel_offer_v1.erl
```diff
@@ -0,0 +1,123 @@
...
+
+-spec sign(offer(), function()) -> offer().
+sign(Offer, SigFun) ->
+    EncodedReq = ?MODULE:encode(Offer#blockchain_state_channel_offer_v1_pb{signature= <<>>}),
+    Signature = SigFun(EncodedReq),
+    Offer#blockchain_state_channel_offer_v1_pb{signature=Signature}.
+
+-spec validate(offer()) -> true | {error, any()}.
+validate(Offer) ->
+    %% TODO: enhance
```
mrallen1, line 87: how? please flesh this comment out

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -19,13 +19,15 @@
...
+    summaries/1, summaries/2, update_summary_for/3,
 
     add_payload/3,
     get_summary/2,
     num_packets_for/2, num_dcs_for/2,
 
-    to_json/2
+    to_json/2,
+
+    causality/2
```
mrallen1, line 30: in line with comment above, consider renaming this

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -286,6 +328,34 @@ can_fit(ClientPubkeyBin, Summaries) ->
             true
     end.
 
+-spec check_causality(SCSummaries :: summaries(),
```
mrallen1, line 331: Maybe this should get moved into a utility module so it could be used during the close txn close_proposal code?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2116,55 +2117,66 @@ find_routing(OUI, Ledger) ->
...
-                                {ok, Route} ->
-                                    {ok, [Route]};
-                                Error ->
-                                    Error
-                            end
-                    end;
-                <<_:25/integer, Prefix:7/integer>> ->
-                    {error, {unknown_devaddr_prefix, Prefix}}
-            end
+        <<_:25/integer, Prefix:7/integer>> ->
```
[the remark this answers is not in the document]
vihu: Maybe? The match is on `integer` vs `integer--unsigned-little` in the binary. I didn't actually change anything here just split the two clauses out to separate functions.

src/state_channel/v1/blockchain_state_channel_v1.erl
vihu: I'll rename it to `compare_causality`

src/state_channel/v1/blockchain_state_channel_v1.erl
vihu: Hmm, I thought the exposed `causality` (which I'll rename to `compare_causality`) might be enough for that? If we do need to expose this we can do so. I'll take another look when I start integrating this with the v2 work you have

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -255,6 +388,146 @@ send_packet(Swarm, Stream, Packet, Region) ->
...
+            Error
+    end.
+
+-spec store_state_channel(SC :: blockchain_state_channel_v1:state_channel(),
+                          State :: state()) -> ok | {error, any()}.
+store_state_channel(SC, #state{db=DB, cf=CF}=State) ->
+    SCID = blockchain_state_channel_v1:id(SC),
+    case get_state_channels(SCID, State) of
+        {ok, SCs} ->
+            ToInsert = erlang:term_to_binary([SC | SCs]),
```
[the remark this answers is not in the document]
vihu: I *think* it's okay the way it is right now. We call `store_state_channel` in two places and _only_ when there is a causal conflict (and in this case we do want to have two copies of the channel stored under the same scid key). As long as a state channel is valid, we will always overwrite the state channel value stored in rocks.

src/ledger/v1/blockchain_ledger_v1.erl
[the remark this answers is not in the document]
Vagabond: Yes, it will match if the low 7 bits do not match the DevAddrPrefix from the app env.

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -159,7 +220,64 @@ handle_info(_Msg, State) ->
...
 %% ------------------------------------------------------------------
 
--spec find_stream(AddressOrOUI :: string() | non_neg_integer(), State :: state()) -> undefined | dialing |pid().
+-spec handle_banner(Banner :: blockchain_state_channel_banner_v1:banner(),
+                    Stream :: pid(),
+                    State :: state()) -> ok | {error, any()}.
+handle_banner(Banner, Stream, State) ->
+    case blockchain_state_channel_banner_v1:sc(Banner) of
+        undefined ->
+            ok;
```
Vagabond, line 229: Arguably this is only ok if you're in the same OUI as the router.

vihu: hmm, we do send a banner with an empty sc here though: https://github.com/helium/blockchain-core/pull/515/files#diff-9f8546e8769ceed583b17d13ed817e36R88

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -171,19 +289,34 @@ add_stream(AddressOrOUI, Stream, #state{streams=Streams}=State) ->
...
+remove_packet_from_waiting(AddressOrOUI, #state{waiting=Waiting}=State) ->
     State#state{waiting=maps:remove(AddressOrOUI, Waiting)}.
 
+-spec enqueue_packet(Packet :: blockchain_helium_packet_v1:packet(),
+                     State :: state()) -> state().
+enqueue_packet(Packet, #state{packets=Packets}=State) ->
+    State#state{packets=[Packet | Packets]}.
+
+-spec deque_packet(State :: state()) -> {undefined | blockchain_helium_packet_v1:packet(), state()}.
+deque_packet(#state{packets=Packets}=State) when length(Packets) > 0 ->
```
Vagabond, line 313: dequeue

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -159,7 +220,64 @@ handle_info(_Msg, State) ->
...
+            State;
+        {error, _} ->
+            ok = libp2p_framed_stream:close(Stream),
+            State;
+        ok ->
+            case deque_packet(State) of
+                {undefined, _} ->
+                    %% XXX: We somehow don't have this packet in our queue,
+                    %% what should we do? close the stream maybe?
+                    %% store the state channel?
```
Vagabond, line 267: certainly store the SC at least

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -208,12 +345,12 @@ dial_and_send_packet(Swarm, Address, Packet, Region) when is_list(Address) ->
                 Self ! {dial_fail, Address, _Reason};
             {ok, Stream} ->
                 unlink(Stream),
-                ok = send_packet(Swarm, Stream, Packet, Region),
-                Self ! {dial_success, Address, Stream}
+                ok = send_offer(Swarm, Stream, Packet, Region),
+                Self ! {dial_success, Address, Stream, Chain}
```
Vagabond, line 349: Doesn't the caller already have a handle to the chain?

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -239,8 +376,18 @@ dial_and_send_packet(Swarm, Route, Packet, Region) ->
             not_dialed ->
                 Self ! {dial_fail, OUI, failed};
             {dialed, Stream} ->
-                ok = send_packet(Swarm, Stream, Packet, Region),
-                Self ! {dial_success, OUI, Stream}
+                %% Check if hotspot and router oui match for this packet,
+                %% if so, send packet directly otherwise send offer
+                case is_hotspot_in_router_oui(Swarm, OUI, Chain) of
```
Vagabond, line 381: Use send_packet_or_offer here?

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -208,12 +345,12 @@ dial_and_send_packet(Swarm, Address, Packet, Region) when is_list(Address) ->
                 Self ! {dial_fail, Address, _Reason};
             {ok, Stream} ->
                 unlink(Stream),
-                ok = send_packet(Swarm, Stream, Packet, Region),
-                Self ! {dial_success, Address, Stream}
+                ok = send_offer(Swarm, Stream, Packet, Region),
```
Vagabond, line 348: Use send_packet_or_offer here?

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -255,6 +402,145 @@ send_packet(Swarm, Stream, Packet, Region) ->
...
+        true ->
+            send_packet(Swarm, Stream, Packet, Region)
+    end.
+
+%% ------------------------------------------------------------------
+%% State channel validation functions
+%% ------------------------------------------------------------------
+-spec is_valid_sc(SC :: blockchain_state_channel_v1:state_channel(),
+                  State :: state()) -> ok | {error, any()}.
+is_valid_sc(SC, State) ->
```
Vagabond, line 453: If we don't already, we should check for the 'amount' being exceeded by the payments here (i.e. the state channel being exhausted)

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -255,6 +402,145 @@ send_packet(Swarm, Stream, Packet, Region) ->
...
+                            ok;
+                        false ->
+                            {error, causal_conflict}
+                    end
+            end
+    end.
+
+-spec is_active_sc(SC :: blockchain_state_channel_v1:state_channel()) -> boolean().
+is_active_sc(SC) ->
+    %% TODO: cache chain in state?
```
Vagabond, line 474: Yes.

src/state_channel/blockchain_state_channels_db_owner.erl
```diff
@@ -40,16 +46,40 @@ start_link(Args) ->
...
 init(Args) ->
     lager:info("~p init with ~p", [?MODULE, Args]),
     erlang:process_flag(trap_exit, true),
     BaseDir = maps:get(base_dir, Args),
-    {ok, DB} = open_db(BaseDir),
-    {ok, #state{db=DB}}.
+    CFs = maps:get(cfs, Args, ["default", "sc_servers_cf", "sc_clients_cf"]),
+    case open_db(BaseDir, CFs) of
+        {error, _Reason} ->
+            %% Go into a tight loop
```
Vagabond, line 65: Why?

src/state_channel/blockchain_state_channels_client.erl
Vagabond: Which you should only accept if you're in the same OUI and don't care about offer/buy

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -14,8 +14,13 @@
...
     nonce/1,
     packet/2,
+    offer/2,
     state_channels/0,
-    active_sc_id/0
+    active_sc_id/0,
+    active_sc/0,
+
+    %% For testing
+    insert_fake_sc_skewed/2
```
Vagabond, line 23: you can ifdef TEST this export...

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -94,32 +111,25 @@ code_change(_OldVsn, State, _Extra) ->
...
-                    ok = send_packet(Swarm, Stream, Packet, Region),
-                    StateAcc
-            end
-        end,
-        State0,
-        RoutesOrAddresses
-    ),
-    {noreply, State1};
+handle_cast(_, #state{chain=undefined}=State) ->
+    erlang:send_after(500, self(), post_init),
```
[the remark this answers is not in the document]
vihu: Fair enough, I'll remove this and let it go to the catch-all `handle_cast` I suppose

src/state_channel/blockchain_state_channels_db_owner.erl
```diff
@@ -40,16 +46,40 @@ start_link(Args) ->
...
+            init(Args);
```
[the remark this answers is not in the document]
vihu: Yeah, got overly defensive there, removed

src/state_channel/blockchain_state_channels_db_owner.erl
vihu: removed

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -264,6 +267,45 @@ to_json(SC, _Opts) ->
...
+                               {done, conflict};
+                           true ->
+                               {done, equal}
+                       end;
+                   {N1, N2} when N1 < N2 ->
+                       %% SC2 claims to have a higher nonce than SC1
+                       %% Every hotspot in SC1 summary must have either the same/lower packet and dc count
+                       %% when compared to SC2 summary
+                       check_causality(?MODULE:summaries(SC1), SC2, caused);
+                   {N1, N2} when N1 > N2 ->
```
Vagabond, line 300: I'm confused why you don't just pass SC1 and SC2 to a function here and assert that SC2 should always be an effect of SC1?

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -140,37 +188,44 @@ handle_cast({packet, SCPacket},
...
+            ok = send_banner(SignedSC, HandlerPid),
+
+            {noreply, State#state{state_channels=maps:update(ActiveSCID, {SignedSC, Skewed1}, SCs)}}
     end;
+handle_cast({offer, SCOffer, _Pid}, #state{active_sc_id=undefined}=State) ->
+    lager:warning("Got offer: ~p when no sc is active", [SCOffer]),
+    {noreply, State};
+handle_cast({offer, SCOffer, HandlerPid},
+            #state{active_sc_id=ActiveSCID, state_channels=SCs, owner={_Owner, OwnerSigFun}, chain=Chain}=State) ->
+    %% TODO: handle offer
```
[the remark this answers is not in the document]
vihu: yeah

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -140,37 +188,44 @@ handle_cast({packet, SCPacket},
...
+    Ledger = blockchain:ledger(Chain),
+    Routing = blockchain_state_channel_offer_v1:routing(SCOffer),
+    Region = blockchain_state_channel_offer_v1:region(SCOffer),
+    Hotspot = blockchain_state_channel_offer_v1:hotspot(SCOffer),
+    PacketHash = blockchain_state_channel_offer_v1:packet_hash(SCOffer),
+    PayloadSize = blockchain_state_channel_offer_v1:payload_size(SCOffer),
+    {ActiveSC, _} = maps:get(ActiveSCID, SCs, undefined),
+    lager:info("Routing: ~p, Hotspot: ~p", [Routing, Hotspot]),
+
+    %% XXX: Accepting all offers for now
```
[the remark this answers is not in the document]
vihu: yes, I guess that logic goes to the router side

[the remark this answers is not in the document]
vihu: All we have is a test handler here

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -213,6 +266,11 @@ handle_info({blockchain_event, {add_block, BlockHash, _Syncing, _Ledger}}, #stat
                end,
 
     {noreply, NewState};
+handle_info({'DOWN', _Ref, process, Pid, _}, State=#state{streams=Streams}) ->
```
andymck, line 269: i dont see anything being monitored by the server and no linked processes either

src/state_channel/v1/blockchain_state_channel_banner_v1.erl
```diff
@@ -0,0 +1,59 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain State Channel Banner ==
```
andymck, line 3: for my understanding, what does 'banner' mean here in the context of SCs ?

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -129,22 +136,42 @@ handle_call(state, _From, State) ->
...
+handle_info({dial_success, Address, Stream}, #state{swarm=Swarm}=State0) ->
+    Packets = get_waiting_packet(Address, State0),
+    lager:debug("dial_success sending ~p packet offers", [erlang:length(Packets)]),
+    lists:foreach(
+        fun({Packet, Region}) ->
+            ok = send_offer(Swarm, Stream, Packet, Region)
+        end,
+        Packets
+    ),
+    erlang:monitor(process, Stream),
```
andymck, line 173: i see the streams are monitored and removed from the state when they go down, is there any need to redial ?

src/state_channel/blockchain_state_channels_server.erl
vihu: Nice catch, it turns out the streams are never actually even put in the state in the first place here. Derp. Fixing.

src/state_channel/blockchain_state_channels_client.erl
vihu: Maybe down the line we may redial, but for this pass we'd just drop the packet if we couldn't get the stream to open. Any new incoming packet will atleast try to dial initially, if it's in "dialing" state, it will enqueue the packet and try again. But after that the packet is dropped.

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -158,8 +185,116 @@ handle_info(_Msg, State) ->
...
 %% ------------------------------------------------------------------
+-spec handle_packet(Packet :: blockchain_helium_packet_v1:packet(),
+                    RoutesOrAddresses :: [string()] | [blockchain_ledger_routing_v1:routing()],
+                    Region :: atom(),
+                    State :: state()) -> state().
+handle_packet(Packet, RoutesOrAddresses, Region, #state{swarm=Swarm, chain=Chain}=State0) ->
+    lager:info("handle_packet ~p to ~p", [Packet, RoutesOrAddresses]),
+    State1 = lists:foldl(
+               fun(RouteOrAddress, StateAcc) ->
+                       StreamKey = case erlang:is_list(RouteOrAddress) of %% NOTE: This is actually a string check
```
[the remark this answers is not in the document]
vihu: I've added an `is_routing` function to `blockchain_ledger_routing_v1` just to have a bit more descriptive check here. Also it's either a string or a "routing" info, OUI (int) is what we lookup after figuring that out :)

src/state_channel/v1/blockchain_state_channel_banner_v1.erl
vihu: Added some doc: ab8abc8

mrallen1:
@vihu BK complains about dialyzer errors with this latest commit - can you fix those and then we can marge ?

vihu:
> @vihu BK complains about dialyzer errors with this latest commit - can you fix those and then we can marge ?

oh yeah, will fix shortly

vihu:
> @vihu BK complains about dialyzer errors with this latest commit - can you fix those and then we can marge ?

10bbbc9 should fix it

PR #516, opened 2020-06-21
flush memtables before snapshot to prevent drift in snapshots

evanmcc:


evanmcc:
this wasn't the right place or fix, closing

PR #517, opened 2020-06-22
cache mode switch

evanmcc:
This should allow us to disable rocks snapshots, which seem to be causing some sync issues.  I can't seem to get the cache generated at commit time to work correctly, so this might not be as efficient as it should be.


evanmcc:
pulled the context fold optimization out.  I don't think that we're going this way, and it would need a rebase in any case.

PR #521, opened 2020-06-23
release rocks snapshots manually

evanmcc:
I don't know if this has anything to do with our current snapshot troubles, but this is a (small) memory leak from what we can tell, and also probably isn't great for disk and memory utilization for long-running nodes.


evanmcc:
this isn't actually workable, and the concern that it was meant to address has been proven not to be an issue.

PR #522, opened 2020-06-23
witness retention limits

evanmcc:
This PR establishes a chain var controlled limit on how many witnesses are allowed to be stored per gateway.  The limit is enforced by random selection of a subset of the witnesses.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_gateway_v2.erl
+++ b/src/ledger/v1/blockchain_ledger_gateway_v2.erl
@@ -404,7 +404,7 @@ witnesses(Gateway) ->
     Gateway#gateway_v2.witnesses.
 
 -spec witnesses(#{libp2p_crypto:pubkey_bin() => gateway_witness()}, gateway()) ->
-                       #{libp2p_crypto:pubkey_bin() => gateway_witness()}.
+                       gateway().
 witnesses(Witnesses, Gateway) ->
     Gateway#gateway_v2{witnesses = Witnesses}.
 
--- a/src/ledger/v1/blockchain_ledger_gateway_v2.erl
+++ b/src/ledger/v1/blockchain_ledger_gateway_v2.erl
@@ -26,7 +26,7 @@
     has_witness/2,
     clear_witnesses/1,
     remove_witness/2,
-    witnesses/1,
+    witnesses/1, witnesses/2,
     witness_hist/1, witness_recent_time/1, witness_first_time/1,
     oui/1, oui/2
 ]).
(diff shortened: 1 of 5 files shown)


src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -724,6 +724,8 @@ validate_var(?poc_path_limit, Value) ->
     validate_int(Value, "poc_path_limit", 3, 10, false);
 validate_var(?poc_witness_consideration_limit, Value) ->
     validate_int(Value, "poc_witness_consideration_limit", 10, 100, false);
+validate_var(?poc_witness_retention_limit, Value) ->
+    validate_int(Value, "poc_witness_consideration_limit", 10, 100, false);
```
vihu, line 728:
```suggestion
    validate_int(Value, "poc_witness_retention_limit", 10, 100, false);
```

src/ledger/v1/blockchain_ledger_gateway_v2.erl
```diff
@@ -403,6 +403,11 @@ has_witness(#gateway_v2{witnesses=Witnesses}, WitnessAddr) ->
 witnesses(Gateway) ->
     Gateway#gateway_v2.witnesses.
 
+-spec witnesses(#{libp2p_crypto:pubkey_bin() => gateway_witness()}, gateway()) ->
```
vihu, line 406: Uhm, this spec seems incorrect? The return type is a `gateway`

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1161,7 +1161,17 @@ add_gateway_witnesses(GatewayAddress, WitnessInfo, Ledger) ->
                                               erlang:error({add_gateway_error, Reason})
                                       end
                               end, GW0, WitnessInfo),
-            update_gateway(GW1, GatewayAddress, Ledger)
+            GW2 = case config(?poc_witness_retention_limit, Ledger) of
+                      {ok, Limit} ->
+                          Witnesses = blockchain_ledger_gateway_v2:witnesses(GW1),
+                          {ok, Height} = current_height(Ledger),
+                          %% this doesn't need to be hard to predict, just random
+                          RandState = blockchain_utils:rand_state(integer_to_binary(Height)),
```
Vagabond, line 1169: Would it be more efficent to check if the gateway is over the limit first?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1161,7 +1161,17 @@ add_gateway_witnesses(GatewayAddress, WitnessInfo, Ledger) ->
...
+                          Witnesses1 = blockchain_utils:deterministic_subset(Limit, RandState, Witnesses),
```
vihu, line 1170: `deteministic_subset` expects a list, `Witnesses` is a map

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1161,7 +1161,17 @@ add_gateway_witnesses(GatewayAddress, WitnessInfo, Ledger) ->
...
+                          blockchain_ledger_gateway_v2:witnesses(Witnesses1, GW1);
```
vihu, line 1171: `witnesses/2` expects a map as first arg

Vagabond:
What if, instead, we just reduced the rate of challenging/pathing through hotspots with more than X witnesses?

evanmcc:
So, we had an offline conversation, and are a bit worried about losing this information.  I think that #504 will take care of this sufficiently for now, so I'm going to close it but not delete the branch for now, since we might need it in the future.

PR #523, opened 2020-06-25
DC overcommit and refund for state channels

mrallen1:
To encourage honest behavior by routers and hotspots, DCs are overcommitted at state channel open time and then potentially refunded during state channel garbage collection.

[CH 7796](https://app.clubhouse.io/hlm/story/7796/state-channels-v2)

### DIFF
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -394,7 +394,7 @@ import(Chain, SHA,
                                       Hash = blockchain_block:hash_block(Block),
                                       ok = blockchain_ledger_v1:maybe_gc_pocs(Chain, Ledger2),
 
-                                      ok = blockchain_ledger_v1:maybe_gc_scs(Ledger2),
+                                      ok = blockchain_ledger_v1:maybe_gc_scs(Chain, Ledger2),
 
                                       ok = blockchain_ledger_v1:refresh_gateway_witnesses(Hash, Ledger2);
                                   _ ->
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -394,7 +394,7 @@ import(Chain, SHA,
                                       Hash = blockchain_block:hash_block(Block),
                                       ok = blockchain_ledger_v1:maybe_gc_pocs(Chain, Ledger2),
 
-                                      ok = blockchain_ledger_v1:maybe_gc_scs(Chain, Ledger2),
+                                      ok = blockchain_ledger_v1:maybe_gc_scs(Chain),
 
                                       ok = blockchain_ledger_v1:refresh_gateway_witnesses(Hash, Ledger2);
                                   _ ->
--- a/src/ledger/v1/blockchain_ledger_state_channel_v2.erl
+++ b/src/ledger/v1/blockchain_ledger_state_channel_v2.erl
@@ -26,23 +26,23 @@
 -endif.
 
 -record(ledger_state_channel_v2, {
-    id :: binary(),
-    owner :: binary(),
+    id :: blockchain_state_channel_v1:id(),
+    owner :: libp2p_crypto:pubkey_bin(),
     expire_at_block :: pos_integer(),
     original :: non_neg_integer(),
     amount :: non_neg_integer(),
     nonce :: non_neg_integer(),
     closer :: libp2p_crypto:pubkey_bin(),
-    sc :: blockchain_state_channel_v1:state_channel(),
-    close_state :: closed | dispute
+    sc :: undefined | blockchain_state_channel_v1:state_channel(),
+    close_state :: closed | dispute | undefined
 }).
 
 -type state_channel_v2() :: #ledger_state_channel_v2{}.
 
 -export_type([state_channel_v2/0]).
 
--spec new(ID :: binary(),
-          Owner :: binary(),
+-spec new(ID :: blockchain_state_channel_v1:id(),
+          Owner :: libp2p_crypto:pubkey_bin(),
           ExpireAtBlock :: pos_integer(),
           OriginalAmtDC :: non_neg_integer(),
           TotalAmtDC :: non_neg_integer(),
@@ -57,19 +57,19 @@ new(ID, Owner, ExpireAtBlock, OriginalAmount, TotalAmount, Nonce) ->
        nonce=Nonce
       }.
 
--spec id(state_channel_v2()) -> binary().
+-spec id(state_channel_v2()) -> blockchain_state_channel_v1:id().
 id(#ledger_state_channel_v2{id=ID}) ->
     ID.
 
--spec id(ID :: binary(), SC :: state_channel_v2()) -> state_channel_v2().
+-spec id(ID :: blockchain_state_channel_v1:id(), SC :: state_channel_v2()) -> state_channel_v2().
 id(ID, SC) ->
     SC#ledger_state_channel_v2{id=ID}.
 
--spec owner(state_channel_v2()) -> binary().
+-spec owner(state_channel_v2()) -> libp2p_crypto:pubkey_bin().
 owner(#ledger_state_channel_v2{owner=Owner}) ->
     Owner.
 
--spec owner(Owner :: binary(), SC :: state_channel_v2()) -> state_channel_v2().
+-spec owner(Owner :: libp2p_crypto:pubkey_bin(), SC :: state_channel_v2()) -> state_channel_v2().
 owner(Owner, SC) ->
     SC#ledger_state_channel_v2{owner=Owner}.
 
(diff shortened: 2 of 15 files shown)


src/ledger/v1/blockchain_ledger_state_channel_v2.erl
```diff
@@ -0,0 +1,222 @@
...
+    state_channel/1,
+    close_state/1
+]).
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-record(ledger_state_channel_v2, {
+    id :: binary(),
```
vihu, line 27:
```suggestion
    id :: blockchain_state_channel_v1:id(),
```

src/ledger/v1/blockchain_ledger_state_channel_v2.erl
```diff
@@ -0,0 +1,222 @@
...
+    owner :: binary(),
```
vihu, line 28:
```suggestion
    owner :: libp2p_crypto:pubkey_bin(),
```

vihu: Would need to reflect the specs in the fun heads as well...

src/ledger/v1/blockchain_ledger_state_channel_v2.erl
```diff
@@ -0,0 +1,222 @@
...
+    expire_at_block :: pos_integer(),
+    amount :: non_neg_integer(),
+    nonce :: non_neg_integer(),
+    closer :: libp2p_crypto:pubkey_bin(),
+    sc :: blockchain_state_channel_v1:state_channel(),
```
vihu, line 33: When is the `sc` in this record actually getting set? It's not in the `new/5` function, neither there is a setter for it...

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2232,30 +2258,54 @@ find_scs_by_owner(Owner, Ledger) ->
...
     {ok, CurrHeight} = ?MODULE:current_height(Ledger),
-    Routing = blockchain_ledger_state_channel_v1:new(ID, Owner, CurrHeight+ExpireWithin, Nonce),
-    Bin = blockchain_ledger_state_channel_v1:serialize(Routing),
     Key = state_channel_key(ID, Owner),
+    Bin = case ?MODULE:config(?sc_version, Ledger) of
+        {ok, 2} ->
+            Routing = blockchain_ledger_state_channel_v2:new(ID, Owner,
+                                                             CurrHeight+ExpireWithin,
+                                                             Amount, Nonce),
+            blockchain_ledger_state_channel_v1:serialize(Routing);
```
vihu, line 2282: This is `v2` no?

src/ledger/v1/blockchain_ledger_state_channel_v2.erl
vihu: Oh I see, only when `close_proposal` is called `sc` gets set. In that case it can go undefined as well?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -842,6 +842,10 @@ validate_var(?sc_grace_blocks, Value) ->
     validate_int(Value, "sc_grace_blocks", 1, 100, false);
 validate_var(?dc_payload_size, Value) ->
     validate_int(Value, "dc_payload_size", 1, 32, false);
+validate_var(?sc_version, Value) ->
+    validate_int(Value, "sc_version", 1, 10, false);
+validate_var(?sc_overcommit, Value) ->
+    validate_int(Value, "sc_overcommit", 1, 10, false); %% XXX should this be a float?
```
vihu, line 848: > %% XXX should this be a float?
Please no, when we do get to submit it, let's just do 2x or whatever, instead of 1.5343 :trollface:

mrallen1: originally @Vagabond said it should be 2.5, but I'd prefer an integer too.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1423,13 +1425,27 @@ maybe_gc_scs(Ledger) ->
                                Ledger,
                                SCsCF,
                                fun({KeyHash, BinSC}, Acc) ->
-                                       SC = blockchain_ledger_state_channel_v1:deserialize(BinSC),
-                                       ExpireAtBlock = blockchain_ledger_state_channel_v1:expire_at_block(SC),
+                                       {Mod, SC} = deserialize_state_channel(BinSC),
+                                       ExpireAtBlock = Mod:expire_at_block(SC),
                                        case (ExpireAtBlock + Grace) < Height of
```
mrallen1, line 1430: @Vagabond wanted to make sure that grace was accounted for before entries are removed from the ledger, so just want to double check that here

src/ledger/v1/blockchain_ledger_state_channel_v2.erl
mrallen1: yes, i added to the spec for the record

src/ledger/v1/blockchain_ledger_v1.erl
mrallen1: fixed

src/transactions/v1/blockchain_txn_vars_v1.erl
Vagabond: I'm ok with 2x

src/ledger/v1/blockchain_ledger_state_channel_v2.erl
```diff
@@ -0,0 +1,249 @@
...
+            case maybe_dispute(Closer, closer(SCEntry), SC, state_channel(SCEntry)) of
+                closed ->
+                    SCEntry#ledger_state_channel_v2{closer=Closer, sc=SC, close_state=closed};
+                {dispute, NewCloser, NewSC} ->
+                    %% store the "latest" (as judged by nonce)
+                    SCEntry#ledger_state_channel_v2{closer=NewCloser, sc=NewSC, close_state=dispute}
+            end;
+        dispute ->
+            %% already marked as dispute, just keep it that way
+            SCEntry
```
Vagabond, line 147: There still might be a newer version here with more up to date payouts

src/ledger/v1/blockchain_ledger_v1.erl
Vagabond: I think we can GC a v2 if if the expire_at + grace time is less than the start of the current epoch.

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -394,7 +394,7 @@ import(Chain, SHA,
                                       Hash = blockchain_block:hash_block(Block),
                                       ok = blockchain_ledger_v1:maybe_gc_pocs(Chain, Ledger2),
 
-                                      ok = blockchain_ledger_v1:maybe_gc_scs(Ledger2),
+                                      ok = blockchain_ledger_v1:maybe_gc_scs(Chain, Ledger2),
```
Vagabond, line 397: You don't need the chain AND the ledger, the ledger is inside the chain...

src/ledger/v1/blockchain_ledger_state_channel_v2.erl
```diff
@@ -143,8 +143,17 @@ close_proposal(Closer, SC, SCEntry) ->
...
             end;
         dispute ->
-            %% already marked as dispute, just keep it that way
-            SCEntry
+            %% already marked as dispute
+            %% Check to see if the nonce is updated, if so replace
+            CurrentNonce = blockchain_state_channel_v1:nonce(SC),
+            PreviousNonce = blockchain_state_channel_v1:none(state_channel(SCEntry)),
+
+            case CurrentNonce > PreviousNonce andalso Closer == closer(SCEntry) of
```
Vagabond, line 151: I'm not sure the closers need to be the same?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1443,15 +1445,19 @@ maybe_gc_scs(Ledger) ->
...
-                                                       case blockchain_ledger_state_channel_v2:close_state(SC) of
-                                                           undefined -> [KeyHash | Acc]; %% due to tests must handle
-                                                           dispute -> [KeyHash | Acc]; %% confiscate overcommit
-                                                           closed ->
-                                                               SC0 = blockchain_ledger_state_channel_v2:state_channel(SC),
-                                                               Owner = blockchain_state_channel_v1:owner(SC0),
-                                                               Credit = calc_remaining_dcs(SC0),
-                                                               ok = credit_dc(Owner, Credit, Ledger),
-                                                               [KeyHash | Acc]
+                                                       case (ExpireAtBlock + Grace) < EpochStart of
```
Vagabond, line 1448: :+1:

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
mrallen1: OK

src/ledger/v1/blockchain_ledger_state_channel_v2.erl
mrallen1: They are during `maybe_dispute` but maybe that is also wrong?

Vagabond: I think as long as the closer is a participant in the state channel, and the state channel is signed by the owner, it's ok

mrallen1: ok

PR #529, opened 2020-06-29
Add a test for the case where staking keys are defined but not in txn

Vagabond:


test/blockchain_price_oracle_SUITE.erl
```diff
@@ -1249,6 +1249,17 @@ staking_key_add_gateway(Config) ->
 
     ?assertMatch({error,payer_invalid_staking_key}, blockchain_txn:is_valid(SignedPayerAddGatewayTx02, Chain)),
 
+    %% check no staking key fails
+    AddGatewayTx000 = blockchain_txn_add_gateway_v1:new(Owner, Gateway),
+
+    %% set the fees on the base txn and then sign the various txns
+    AddGatewayTx001 = blockchain_txn_add_gateway_v1:fee(AddGatewayTx000, AddGatewayTxFee),
```
andymck, line 1256: prob should recalculate fees for the new txn as the fees being used were calculated for a txn with a payer specified which will result in some difference. 

that said the staking server is validated before the fees and thus the point is irrelevant for this test...

Vagabond: yeah, sort of incidental really

PR #533, opened 2020-07-02
[WIP] State channel v2 tests

vihu:


vihu:
This is no longer WIP, all (most) tests for state channels would live in miner and router.

PR #534, opened 2020-07-02
Add ledger:next_oracle_price/1

Vagabond:
This function can be used to determine if a price oracle change is pending, and when it will become effective (minimum time).

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -102,6 +102,7 @@
     maybe_recalc_price/2,
     add_oracle_price/2,
     current_oracle_price/1,
+    next_oracle_prices/1,
     current_oracle_price_list/1,
 
     apply_raw_changes/2,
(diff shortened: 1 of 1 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2362,6 +2363,48 @@ current_oracle_price(Ledger) ->
...
+
+    %% get the prices currently too new to be considered
+    PendingPrices = trim_price_list(StartScan, Prices),
+
+    %% get the prices expiring in the next DelaySecs
+    ExpiringPrices = [ P || P <- Prices, blockchain_ledger_oracle_price_entry:timestamp(P) < EndScan],
+
+    %% walk the pending prices and see if the price will change, what it will change to and when
+    tl(lists:reverse(lists:foldl(
+      fun(T, [{Price, _Time}|_]=Acc) ->
```
mrallen1, line 2393: There's a lot going on in this fold fun - any chance you could refactor it to make it easier to understand what's going on?

Vagabond: I tried to comment it more, but I couldn't find a good way to refactor it that added any real value

Vagabond: I tried to comment it more, but I couldn't find a good way to refactor it that added any real value

PR #535, opened 2020-07-06
Remove txn/staking from add/assert cli

madninja:
The txn add_gateway and txn assert_location cli commands were previously expected to take fee/staking fee values. Since the user can no longer easily calculate these, they're no fetched from the chain using the proper txn calculate functions. 

**NOTE**: This now requires a synced chain before the user can add/assert their cloud miner

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -45,7 +45,8 @@
 
     init_assumed_valid/2,
 
-    add_gateway_txn/4, assert_loc_txn/6,
+    add_gateway_txn/2, add_gateway_txn/4,
+    assert_loc_txn/4, assert_loc_txn/6,
 
     add_snapshot/2, get_snapshot/2, find_last_snapshot/1,
     find_last_snapshots/2,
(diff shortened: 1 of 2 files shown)


src/blockchain.erl
```diff
@@ -1671,15 +1673,42 @@ add_gateway_txn(OwnerB58, PayerB58, Fee, StakingFee) ->
     SignedTxn = blockchain_txn_add_gateway_v1:sign_request(Txn, SigFun),
     {ok, blockchain_txn:serialize(SignedTxn)}.
 
+%% @doc Creates a signed add_gatewaytransaction with this blockchain's keys as
```
andymck, line 1676: Do we need to keep the add_gateway_txn/4 version of this ?

src/blockchain.erl
```diff
@@ -1701,6 +1730,35 @@ assert_loc_txn(H3String, OwnerB58, PayerB58, Nonce, StakingFee, Fee) ->
     SignedTxn = blockchain_txn_assert_location_v1:sign_request(Txn, SigFun),
     {ok, blockchain_txn:serialize(SignedTxn)}.
 
+%% @doc Creates a signed assert_location transaction using the keys of
```
andymck, line 1733: Do we need to keep the assert_loc_txn/6 version of this ?

src/blockchain.erl
madninja: Yes because one is for the cli which assumes a synced chain, the other for the mobile app which talks to the api to get fee information

src/blockchain.erl
madninja: Yes because one is for the cli which assumes a synced chain, the other for the mobile app which talks to the api to get fee information

PR #538, opened 2020-07-09
Fix validation for some old legacy txns with bad instances on chain

Vagabond:
In the past people submitted some bad transactions with invalid fields.
This change enables stricter validation of those txns, but is done
behind a chain var so that the historically bad transactions will be
valid when syncing.

### DIFF
--- a/src/transactions/v1/blockchain_txn_create_htlc_v1.erl
+++ b/src/transactions/v1/blockchain_txn_create_htlc_v1.erl
@@ -150,9 +150,18 @@ is_valid(Txn, Chain) ->
     PubKey = libp2p_crypto:bin_to_pubkey(Payer),
     BaseTxn = Txn#blockchain_txn_create_htlc_v1_pb{signature = <<>>},
     EncodedTxn = blockchain_txn_create_htlc_v1_pb:encode_msg(BaseTxn),
-    case blockchain_txn:validate_fields([{{payee, ?MODULE:payee(Txn)}, {address, libp2p}},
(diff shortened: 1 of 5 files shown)


src/transactions/v1/blockchain_txn_create_htlc_v1.erl
```diff
@@ -150,9 +150,18 @@ is_valid(Txn, Chain) ->
     PubKey = libp2p_crypto:bin_to_pubkey(Payer),
     BaseTxn = Txn#blockchain_txn_create_htlc_v1_pb{signature = <<>>},
     EncodedTxn = blockchain_txn_create_htlc_v1_pb:encode_msg(BaseTxn),
-    case blockchain_txn:validate_fields([{{payee, ?MODULE:payee(Txn)}, {address, libp2p}},
-                                         {{hashlock, ?MODULE:hashlock(Txn)}, {binary, 32, 64}},
-                                         {{address, ?MODULE:address(Txn)}, {binary, 32, 33}}]) of
+    FieldValidation = case blockchain:config(?extended_txn_field_validation, Ledger) of
+                          {ok, true} ->
```
madninja, line 154: Maybe use a version number for extended field validation so we can reuse this var when we need better validation elsewhere later?

Vagabond: sounds quite defeatist, but it's probably not a bad idea

vihu: @madninja d529afa, we get 50 tries

PR #544, opened 2020-07-22
Add MIC to offer msg

vihu:
MIC appears to be required by the router when handling the state_channel offer message. It's extracted from the payload depending on the message type when we construct the sc offer.


vihu:
No longer needed, closing.

PR #546, opened 2020-07-24
Convert PayerB58 of [] to <<>>

fvasquez:
Needs testing.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -1711,7 +1711,7 @@ add_gateway_txn(OwnerB58, PayerB58) ->
 %% will have the chain vars )
 -spec assert_loc_txn(H3String::string(),
                      OwnerB58::string(),
-                     PayerB58::string() | undefined | [],
+                     PayerB58::string() | undefined,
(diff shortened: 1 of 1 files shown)


src/blockchain.erl
```diff
@@ -1711,7 +1711,7 @@ add_gateway_txn(OwnerB58, PayerB58) ->
 %% will have the chain vars )
 -spec assert_loc_txn(H3String::string(),
                      OwnerB58::string(),
-                     PayerB58::string() | undefined,
+                     PayerB58::string() | undefined | [],
```
madninja, line 1714: `string()` and `[]` are the same thing here so no need to add the empty list

src/blockchain.erl
```diff
@@ -1721,6 +1721,7 @@ assert_loc_txn(H3String, OwnerB58, PayerB58, Nonce, StakingFee, Fee) ->
     Payer = case PayerB58 of
                 undefined -> <<>>;
                 <<>> -> <<>>;
```
madninja, line 1723: Remove this clause since an empty binary should never be passed in

PR #547, opened 2020-07-24
Roll to a new state channel (if available) when current is exhausted

mrallen1:
Problem to solve: currently when a state channel is exhausted, we drop the packet. Instead, roll to a state channel with capacity and use that instead.

### DIFF
--- a/src/state_channel/blockchain_state_channels_server.erl
+++ b/src/state_channel/blockchain_state_channels_server.erl
@@ -203,7 +203,8 @@ handle_cast({reject_offer, SCOffer, HandlerPid}, State) ->
     ok = send_rejection(HandlerPid),
     {noreply, State};
 handle_cast({offer, SCOffer, HandlerPid},
-            #state{active_sc_id=ActiveSCID, state_channels=SCs, owner={_Owner, OwnerSigFun}, chain=Chain}=State) ->
+            #state{active_sc_id=ActiveSCID, state_channels=SCs,
+                   owner={_Owner, OwnerSigFun}, chain=Chain}=State) ->
     lager:info("Got offer: ~p, active_sc_id: ~p", [SCOffer, ActiveSCID]),
 
     PayloadSize = blockchain_state_channel_offer_v1:payload_size(SCOffer),
@@ -228,12 +229,47 @@ handle_cast({offer, SCOffer, HandlerPid},
             case (TotalDCs + NumDCs) > DCAmount andalso
                  application:get_env(blockchain, prevent_sc_overspend, true) of
                 true ->
-                    %% will overspend so drop
-                    %% TODO we should switch to the next state channel here
-                    lager:error("Dropping this packet because it will overspend DC ~p, (cost: ~p, packet: ~p)",
-                                [DCAmount, NumDCs, SCOffer]),
-                    ok = send_rejection(HandlerPid),
-                    {noreply, State};
+                    %% will overspend
+                    %% let's check if we *have* another state channel we can use
+                    %% remove the active state channel from candidates
+                    Candidates = maps:without([ActiveSCID], SCs),
+                    %% list of state channel entries
+                    Filter = fun({_ID, {SC, _Skewed}}) ->
+                                     TtlDCs = blockchain_state_channel_v1:total_dcs(SC),
+                                     DCAmt = blockchain_state_channel_v1:amount(SC),
+                                     blockchain_state_channel_v1:state(SC) == open andalso
+                                        DCAmt >= TtlDCs + NumDCs
+                             end,
+                    case sort_and_filter_scs(Candidates, Filter) of
+                        [] ->
+                            %% there are no other state channels which will work
+                            %% so send rejection
+                            lager:error("Dropping this packet because it will overspend DC ~p, (cost: ~p, packet: ~p)",
+                                        [DCAmount, NumDCs, SCOffer]),
+                            ok = send_rejection(HandlerPid),
+                            lager:debug("Setting state channel ~p to pending close.",
+                                       [ActiveSCID]),
+                            NewSCs = set_sc_to_pending_close(ActiveSCID, ActiveSC, Skewed, SCs),
+                            {noreply, State#state{active_sc_id=undefined,
+                                                  state_channels=NewSCs}};
+
+                        SortedAndFilteredSCs when length(SortedAndFilteredSCs) >= 1 ->
+                            {CandID, {CandSC, CandSkewed}} = hd(SortedAndFilteredSCs),
+                            lager:info("Current SC exhausted. Now using ~p, routing: ~p, hotspot: ~p",
+                                       [CandID, Routing, Hotspot]),
+                            {ok, NewSC} = send_purchase(CandSC, Hotspot, HandlerPid, PacketHash,
+                                       PayloadSize, Region, Ledger, OwnerSigFun),
+                            NewState = maybe_add_stream(Hotspot, HandlerPid,
+                                       State#state{active_sc_id=CandID,
+                                            state_channels=maps:put(CandID, {NewSC, CandSkewed}, SCs)}),
+                            erlang:monitor(process, HandlerPid),
+                            ok = maybe_broadcast_new_banner(NewState),
+                            lager:debug("Setting state channel ~p to pending close.",
+                                       [ActiveSCID]),
+                            UpdatedSCs = set_sc_to_pending_close(ActiveSCID, ActiveSC, Skewed, SCs),
+                            {noreply, NewState#state{state_channels = UpdatedSCs}}
+                    end;
+
                 false ->
                     lager:info("Routing: ~p, Hotspot: ~p", [Routing, Hotspot]),
 
@@ -309,6 +345,14 @@ terminate(_Reason, _State) ->
 %% ------------------------------------------------------------------
 %% Internal Function Definitions
 %% ------------------------------------------------------------------
+-spec set_sc_to_pending_close(SCID :: blockchain_state_channel_v1:id(),
+                              SC :: blockchain_state_channel_v1:state_channel(),
+                              Skewed :: skewed:skewed(),
+                              ChannelMap :: state_channels()) -> UpdatedMap :: state_channels().
+set_sc_to_pending_close(SCID, SC, Skewed, ChannelMap) ->
+    NewSC = blockchain_state_channel_v1:state(pending_close, SC),
+    maps:put(SCID, {NewSC, Skewed}, ChannelMap).
+
 -spec process_packet(ClientPubkeyBin :: libp2p_crypto:pubkey_bin(),
                      Packet :: blockchain_helium_packet_v1:packet(),
                      SC :: blockchain_state_channel_v1:state_channel(),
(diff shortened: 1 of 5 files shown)


src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -226,12 +226,43 @@ handle_cast({offer, SCOffer, HandlerPid},
...
+                                     DCAmt = blockchain_state_channel_v1:amount(SC),
+                                     blockchain_state_channel_v1:state(SC) /= closed andalso
+                                        DCAmt >= TtlDCs + NumDCs
+                             end,
+                    case sort_and_filter_scs(maps:to_list(Candidates), Filter) of
+                        [] -> %% there are no other state channels which will work
+                            lager:error("Dropping this packet because it will overspend DC ~p, (cost: ~p, packet: ~p)",
+                                        [DCAmount, NumDCs, SCOffer]),
+                            ok = send_rejection(HandlerPid),
+                            %% FIXME: Should we close the active sc here too?
```
vihu, line 245: I think so? Is there any point in keeping the currently active SC and waiting for either the expiration to happen or GC to trigger? Since we are the owner here, I believe it's okay to issue a close_state_channel here. We should test this path though and make sure it doesn't cause any issues.

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -226,12 +226,43 @@ handle_cast({offer, SCOffer, HandlerPid},
...
+                        [{CandID, {CandSC, CandSkewed}} | _Tail] ->
+                            %% not sure about the sequencing here - should we maybe broadcast a new banner
+                            %% and _then_ send the purchase, or keep it as is?
+                            lager:info("Current SC exhausted. Now using ~p, routing: ~p, hotspot: ~p",
+                                       [CandID, Routing, Hotspot]),
+                            {ok, NewSC} = send_purchase(CandSC, Hotspot, HandlerPid, PacketHash,
+                                       PayloadSize, Region, Ledger, OwnerSigFun),
+                            NewState = maybe_add_stream(Hotspot, HandlerPid,
+                                       State#state{active_sc_id=CandID,
+                                                   state_channels=maps:put(CandID, {NewSC, CandSkewed}, SCs)}),
```
vihu, line 256:
```suggestion
                                                   state_channels=maps:put(CandID, {NewSC, CandSkewed}, Candidates)}),
```

vihu: Or we should remove the `ActiveSC` from the state_channels map here?

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -226,12 +226,43 @@ handle_cast({offer, SCOffer, HandlerPid},
...
+                                     blockchain_state_channel_v1:state(SC) /= closed andalso
+                                        DCAmt >= TtlDCs + NumDCs
+                             end,
+                    case sort_and_filter_scs(maps:to_list(Candidates), Filter) of
+                        [] -> %% there are no other state channels which will work
+                            lager:error("Dropping this packet because it will overspend DC ~p, (cost: ~p, packet: ~p)",
+                                        [DCAmount, NumDCs, SCOffer]),
+                            ok = send_rejection(HandlerPid),
+                            %% FIXME: Should we close the active sc here too?
+                            {noreply, State};
```
vihu, line 246: I wonder if we should set `active_sc_id=undefined` here?

src/state_channel/blockchain_state_channels_server.erl
vihu: I think this is probably fine the way it is, the maps:remove will happen on close. Switching the `active_sc_id` should suffice I suppose.

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -769,3 +792,13 @@ update_sc_summary(ClientPubkeyBin, PayloadSize, Ledger, SC) ->
             %% Update summaries
             blockchain_state_channel_v1:update_summary_for(ClientPubkeyBin, NewSummary, SC)
     end.
+
+maybe_broadcast_new_banner(State, Chain) ->
```
vihu, line 796:
```suggestion
maybe_broadcast_new_banner(#state{chain=Chain}=State) ->
```

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -502,13 +526,7 @@ check_state_channel_expiration(BlockHeight, #state{owner={Owner, OwnerSigFun},
...
     NewState = State#state{active_sc_id=NewActiveSCID, state_channels=NewStateChannels},
 
-    case blockchain:config(sc_version, blockchain:ledger(Chain)) of
-        {ok, 2} ->
-            %% Switching active sc, broadcast banner
-            ok = broadcast_banner(active_sc(NewState), NewState);
-        _ ->
-            ok
-    end,
+    ok = maybe_broadcast_new_banner(NewState, Chain),
```
vihu, line 529:
```suggestion
    ok = maybe_broadcast_new_banner(NewState),
```

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -449,14 +480,7 @@ update_state_sc_close(Txn, #state{db=DB, scf=SCF, state_channels=SCs, chain=Chai
...
 
-    case blockchain:config(sc_version, blockchain:ledger(Chain)) of
-        {ok, 2} ->
-            %% Switching active sc, broadcast banner
-            lager:info("broadcasting banner: ~p", [active_sc(NewState)]),
-            ok = broadcast_banner(active_sc(NewState), NewState);
-        _ ->
-            ok
-    end,
+    ok = maybe_broadcast_new_banner(NewState, Chain),
```
vihu, line 483:
```suggestion
    ok = maybe_broadcast_new_banner(NewState),
```

src/state_channel/blockchain_state_channels_server.erl
mrallen1: I think it's fine to leave it in the state until it gets garbage collected.

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -226,12 +226,43 @@ handle_cast({offer, SCOffer, HandlerPid},
...
+                            {noreply, State#state{active_sc_id=undefined}};
+                        [{CandID, {CandSC, CandSkewed}} | _Tail] ->
+                            lager:info("Current SC exhausted. Now using ~p, routing: ~p, hotspot: ~p",
+                                       [CandID, Routing, Hotspot]),
+                            {ok, NewSC} = send_purchase(CandSC, Hotspot, HandlerPid, PacketHash,
+                                       PayloadSize, Region, Ledger, OwnerSigFun),
+                            NewState = maybe_add_stream(Hotspot, HandlerPid,
+                                       State#state{active_sc_id=CandID,
+                                                   state_channels=maps:put(CandID, {NewSC, CandSkewed}, SCs)}),
+                            %% TODO: We should notify the router that a new state channel is needed
```
vihu, line 257: I think we can punt on this and just have the state channel close, that way as soon as it closes, the sc_worker in router will look at it's active scs (from the ledger) and be able to open one as required.

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -228,12 +228,44 @@ handle_cast({offer, SCOffer, HandlerPid},
...
+                            erlang:monitor(process, HandlerPid),
+                            ok = maybe_broadcast_new_banner(NewState),
+
+                            %% Let's close the active channel since we rolled to a new one...
+                            ok = close_state_channel(ActiveSC, Owner, OwnerSigFun),
```
Vagabond, line 265: We may not be able to close this yet as it might not have hit the expiry time, so we should hang on this until the time has arrived to close it.

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -228,12 +229,47 @@ handle_cast({offer, SCOffer, HandlerPid},
...
+                            {noreply, State#state{active_sc_id=undefined,
+                                                  state_channels=NewSCs}};
+
+                        SortedAndFilteredSCs when length(SortedAndFilteredSCs) >= 1 ->
+                            {CandID, {CandSC, CandSkewed}} = hd(SortedAndFilteredSCs),
+                            lager:info("Current SC exhausted. Now using ~p, routing: ~p, hotspot: ~p",
+                                       [CandID, Routing, Hotspot]),
+                            {ok, NewSC} = send_purchase(CandSC, Hotspot, HandlerPid, PacketHash,
+                                       PayloadSize, Region, Ledger, OwnerSigFun),
+                            NewState = maybe_add_stream(Hotspot, HandlerPid,
```
vihu, line 262: I feel like this whole particular clause could be a bit better, like we update state three times here? It's a little hard to keep track of what is eventually getting updated imo.

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -471,12 +508,12 @@ update_state_sc_close(Txn, #state{db=DB, scf=SCF, state_channels=SCs, chain=Chai
...
 check_state_channel_expiration(BlockHeight, #state{owner={Owner, OwnerSigFun},
                                                    active_sc_id=ActiveSCID,
-                                                   chain=Chain,
                                                    state_channels=SCs}=State) ->
     NewStateChannels = maps:map(
                         fun(_ID, {SC, Skewed}) ->
                                 ExpireAt = blockchain_state_channel_v1:expire_at_block(SC),
-                                case ExpireAt =< BlockHeight andalso blockchain_state_channel_v1:state(SC) == open of
+                                case ExpireAt =< BlockHeight andalso
+                                            blockchain_state_channel_v1:state(SC) /= closed of
```
vihu, line 516: Why was this changed?

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -495,22 +532,17 @@ check_state_channel_expiration(BlockHeight, #state{owner={Owner, OwnerSigFun},
                         _ ->
                             {ActiveSC, _ActiveSCSkewed} = maps:get(ActiveSCID, NewStateChannels),
                             case blockchain_state_channel_v1:state(ActiveSC) of
```
vihu, line 534: It looks like we're flipping cases here unnecessarily? Aren't we doing the same thing as before here?

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -43,7 +43,7 @@
 
 -type state_channel() :: #blockchain_state_channel_v1_pb{}.
 -type id() :: binary().
--type state() :: open | closed.
+-type state() :: open | pending_close | closed.
```
madninja, line 46: are you missing an entry in to_json for this?

mrallen1: good catch

vihu:
@mrallen1 should we close this in favor of https://github.com/helium/blockchain-core/pull/560 ?

mrallen1:
> @mrallen1 should we close this in favor of #560 ?

Yes I think that's a good idea.

PR #548, opened 2020-07-29
Bug fixes for SC V2

vihu:
* As a courtesy for the router, we now reject any offer that the state_channel_server gets when it has no active_sc.
* The validation for `state_channel_open_v1` had a mismatch with the absorb. We were _only_ checking the fee while trying to validate the txn, which would pass, however, the absorb would fail as it would try to debit dcs which the router doesn't have. Instead, we should check the dc_balance in sc_open txn validation _before_ we move to the absorb phase. This fixes it, I think.

The validation fix is guarded behind `sc_open_validation_bugfix` chain variable as we already have some incorrect state channel open txns on the chain.

### DIFF
--- a/src/transactions/v1/blockchain_txn_state_channel_open_v1.erl
+++ b/src/transactions/v1/blockchain_txn_state_channel_open_v1.erl
@@ -147,23 +147,28 @@ absorb(Txn, Chain) ->
     Owner = ?MODULE:owner(Txn),
     ExpireWithin = ?MODULE:expire_within(Txn),
     Nonce = ?MODULE:nonce(Txn),
-    Original = ?MODULE:amount(Txn),
+    OriginalAmount = ?MODULE:amount(Txn),
     TxnFee = ?MODULE:fee(Txn),
     case blockchain_ledger_v1:debit_fee(Owner, TxnFee, Ledger, AreFeesEnabled) of
         {error, _Reason}=Error ->
             Error;
         ok ->
-            Amount = case blockchain_ledger_v1:config(?sc_overcommit, Ledger) of
-                        {ok, Overcommit} -> Original * Overcommit;
-                        _ -> 0
-                     end,
+            Amount = actual_amount(OriginalAmount, Ledger),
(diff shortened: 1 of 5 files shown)


src/transactions/v1/blockchain_txn_state_channel_open_v1.erl
```diff
@@ -222,12 +227,17 @@ do_is_valid_checks(Txn, Chain) ->
...
                                                             TxnFee = ?MODULE:fee(Txn),
+                                                            OriginalAmount = ?MODULE:amount(Txn),
+                                                            ActualAmount = ?MODULE:actual_amount(OriginalAmount, Ledger),
                                                             ExpectedTxnFee = ?MODULE:calculate_fee(Txn, Chain),
                                                             case ExpectedTxnFee =< TxnFee orelse not AreFeesEnabled of
                                                                 false ->
                                                                     {error, {wrong_txn_fee, ExpectedTxnFee, TxnFee}};
                                                                 true ->
-                                                                    blockchain_ledger_v1:check_dc_or_hnt_balance(Owner, TxnFee, Ledger, AreFeesEnabled)
+                                                                    %% Check whether the actual amount (overcommit *
```
macpie, line 237: This is not the end of the world, but it made it look like you could do an implicit burn yo get your DC which we do not do in absorb. So txn would get reject at the absorb phase

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -853,6 +853,12 @@ validate_var(?sc_version, Value) ->
     validate_int(Value, "sc_version", 1, 10, false);
 validate_var(?sc_overcommit, Value) ->
     validate_int(Value, "sc_overcommit", 1, 10, false); %% integer multiplier of amount
+validate_var(?sc_open_validation_bugfix, Value) ->
```
madninja, line 856: We said we were going to make these integers

vihu: Yup, updated here: 36b8f7f

PR #550, opened 2020-08-03
Fix state channel causality issues

vihu:
This re-introduces `effect_of` temporal relation between two state channels. On chain, we are incorrectly marking state channels as having conflicts when in fact they don't.

Also introduced here is `sc_causality_fix` chain var, which would ensure that we still adhere to existing behavior if the var is not set.

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -301,7 +301,7 @@
 -define(sc_overcommit, sc_overcommit).
 %% Number of attempts we get to fix state_channel bugs, we'll set it to 1, max = 50
 -define(sc_open_validation_bugfix, sc_open_validation_bugfix).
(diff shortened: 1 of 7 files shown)


include/blockchain_vars.hrl
```diff
@@ -301,6 +301,8 @@
 -define(sc_overcommit, sc_overcommit).
 %% Number of attempts we get to fix state_channel bugs, we'll set it to 1, max = 50
 -define(sc_open_validation_bugfix, sc_open_validation_bugfix).
+%% Number of attempts we get to fix state_channel causality bugs, we'll set it to 1, max = 50
```
Vagabond, line 304:
```suggestion
%% Number of attempts we get to fix state_channel causality bugs, we'll set it to 1, max = 1
```

PR #552, opened 2020-08-05
Potential fix for undefined chain on dial_success

vihu:
- Also ignore framed stream close return


src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -347,12 +358,12 @@ handle_banner(Banner, Stream, State) ->
             case is_valid_sc(BannerSC, State) of
                 {error, causal_conflict} ->
                     lager:error("causal_conflict for banner sc: ~p", [BannerSC]),
-                    ok = libp2p_framed_stream:close(Stream),
+                    _ = libp2p_framed_stream:close(Stream),
```
mrallen1, line 361: Why did you go to a throwaway `_` response from the function call instead of `ok` here?

vihu: Well, ideally, we'd be tracking the stream better but I think this works okay. The problem that @lthiery was running into was that we were trying to close an already closed stream, but this will just ignore the return from `stream:close`.

PR #554, opened 2020-08-10
Macpie/reject offer

macpie:


src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -245,6 +244,10 @@ handle_cast({offer, SCOffer, HandlerPid},
                     {noreply, NewState}
             end
     end;
+handle_cast({reject_offer, SCOffer, HandlerPid}, State) ->
```
Vagabond, line 247: Can this simply be removed?

Vagabond: I mean this clause of handle_cast if we're going to do rejections directly.

macpie: Yes we can remove it (I did not do it cause I was just testing it first)

PR #555, opened 2020-08-10
Try to reduce the memory footprint of snapshots

Vagabond:


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -770,8 +770,22 @@ active_gateways(Ledger) ->
...
+%% garbage
 snapshot_gateways(Ledger) ->
-    lists:sort(maps:to_list(active_gateways(Ledger))).
+    AGwsCF = active_gateways_cf(Ledger),
+    lists:reverse(cache_fold(
+                    Ledger,
+                    AGwsCF,
+                    fun({Address, Binary}, Acc) ->
+                            Gw = blockchain_ledger_gateway_v2:deserialize(Binary),
+                            %% XXX this might be too expensive
```
Vagabond, line 783: I don't know this is a good idea

Vagabond:
I was also experimenting with doing the snapshot generation in a subprocess with a max_heap set, which stopped the OOM but I wasn't sure how pick a good max_heap setting. Failing to generate the snapshot because of memory pressure is not the worst thing in the world, compared to an OOM.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -770,8 +770,20 @@ active_gateways(Ledger) ->
       #{}
      ).
 
+%% note that instead of calling lists:sort(maps:to_list(active_gateways())) here
+%% we construct the list in the same order without an intermediate map to make less
+%% garbage
 snapshot_gateways(Ledger) ->
-    lists:sort(maps:to_list(active_gateways(Ledger))).
+    AGwsCF = active_gateways_cf(Ledger),
+    lists:reverse(cache_fold(
```
evanmcc, line 778: this needs to be lists:sort for determinism

evanmcc: ignore this, sorry

PR #557, opened 2020-08-10
don't deserialize snap data to save memory and compat

evanmcc:
creating snapshots results in a huge amount of GC pressure, which can OOM smaller hotspots.

this PR shouldn't create as much pressure, as a step on the road to streaming the information to disk piecewise, allowing less even less GC pressure.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -145,6 +145,17 @@
     snapshot_threshold_txns/1,
     load_threshold_txns/2,
 
+    snapshot_raw_gateways/1,
+    load_raw_gateways/2,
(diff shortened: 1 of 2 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3097,6 +3116,33 @@ load_pocs(PoCs, Ledger) ->
       maps:from_list(PoCs)),
     ok.
 
+snapshot_raw(CF, Ledger) ->
+    lists:sort(
```
Vagabond, line 3120: You can probably simply lists:reverse here if you want them in lexical order and save the complexity of the sort

mrallen1:
@Vagabond this is now rebased (and squashed) against current master

PR #560, opened 2020-08-10
Minimal state channel rollover

mrallen1:
When a state channel has no more capacity, try to use a different one. Implement an API for router to use to get the current active count (modulo some headroom factor).

### DIFF
--- a/src/state_channel/blockchain_state_channels_server.erl
+++ b/src/state_channel/blockchain_state_channels_server.erl
@@ -173,7 +173,7 @@ handle_call(active_sc_id, _From, #state{active_sc_id=ActiveSCID}=State) ->
 handle_call(get_active_sc_count, _From, #state{active_sc_id=undefined}=State) ->
     {reply, 0, State};
 handle_call(get_active_sc_count, _From, #state{state_channels=SCs}=State) ->
-    HeadroomFactor = case application:get_env(blockchain, sc_headroom_factor, 10) of
+    HeadroomFactor = case application:get_env(blockchain, sc_headroom_factor, 11) of
                          {ok, X} -> X;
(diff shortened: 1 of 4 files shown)


Vagabond:
Picking the next state channel should check for overspends.

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -165,6 +170,24 @@ handle_call(active_sc, _From, State) ->
...
+handle_call(get_active_sc_count, _From, #state{state_channels=SCs}=State) ->
+    HeadroomFactor = case application:get_env(blockchain, sc_headroom_factor, 11) of
+                         {ok, X} -> X;
+                         X -> X
+                     end,
+    Count = maps:fold(fun(_ID, {SC, _Skewed}, Acc) ->
+                             SCState = blockchain_state_channel_v1:state(SC),
+                             DCAmt = blockchain_state_channel_v1:amount(SC),
+                             TtlDCs = blockchain_state_channel_v1:total_dcs(SC),
+                             Headroom = DCAmt div HeadroomFactor,
```
Vagabond, line 184: Why do we divide here?

vihu: Yeah, dunno why, addressed here: a0a42cc

vihu:
> Picking the next state channel should check for overspends.

Added a filter here: a0a42cc

src/state_channel/blockchain_state_channels_server.erl
mrallen1: Because I wanted the headroom to be a percentage of the total DC amount, but just defining it as an integer works.

PR #568, opened 2020-08-12
More sc optimizations

vihu:


src/state_channel/blockchain_state_channels_db_owner.erl
```diff
@@ -30,12 +34,16 @@
 ]).
 
 -define(DB_FILE, "state_channels.db").
+-define(WRITE_INTERVAL, 1000). % milliseconds
```
vihu, line 37: We should make this configurable, app env maybe?

mrallen1: ok

PR #583, opened 2020-08-17
handle dup txns which result the txn payload existing in both the valid and invalid lists returned from validate

andymck:
If dup txns are submitted by a client, depending on the txn type the dup may fail validations.  An example would be a payment txn, if a dup is submitted it will fail validation due to an invalid nonce.

As part of `blockchain_txn:validate` the txn mgr is returned two lists...the valid txn list and the invalid txn list.  For each cached txn in the txn mgr, we then check which list each txn is in...if a cached txn is in the valid list then its considered valid, if its in the invalid list its considered invalid.  This check is performed using the txn payload itself, the payload is the key.

However in the case of a dup payment txn for example, the txn payload will exist in both the valid and invalid list.  And as txn mgr checks the invalids first, it will result in both payment txns being declared invalid in the txn mgr,  the invalid response being returned to ETL for both and neither txn being submitted upstream.

This PR fixes that by identifying this scenario of the txn being in both lists and then accepting the txn with the lowest txn mgr key as being valid and any others being declared invalid.

An alternative approach would be to pass the txn keys from txn mgr out to blockchain_txn:validate but that looked very messy and makes for poor seperation of concerns.

### DIFF
--- a/src/transactions/blockchain_txn_mgr.erl
+++ b/src/transactions/blockchain_txn_mgr.erl
@@ -343,11 +343,9 @@ process_cached_txns(Chain, CurBlockHeight, SubmitF, _Sync, IsNewElection, NewGro
     %% validate the cached txns
     {ValidTransactions, InvalidTransactions} = blockchain_txn:validate(Txns, Chain),
     ok = lists:foreach(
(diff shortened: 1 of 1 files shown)


andymck:
superseded by https://github.com/helium/blockchain-core/pull/586

PR #587, opened 2020-08-20
Implementation of HIP-10

Vagabond:


src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -789,6 +780,12 @@ get_gateway_owner(Address, Ledger) ->
     end.
 
 
+share_of_dc_rewards( Key, Vars=#{dc_remainder := DCRemainder}) ->
+    erlang:round(DCRemainder * ((maps:get(Key, Vars) / (maps:get(poc_challengers_percent, Vars) + maps:get(poc_challengees_percent, Vars) + maps:get(poc_witnesses_percent, Vars)))));
```
madninja, line 784: Why not match the known var entries in the head?

PR #588, opened 2020-08-20
Downgrade "Unknown LoRaWAN Join" from error to warning

lthiery:
Would be nice to drop this from error logs.


Vagabond:
Could even be a notice level log message

lthiery:
ok I think I fixed the commit history.

PR #589, opened 2020-08-20
add code for finding the first available block

evanmcc:
add a find function so that we can find a particular block higher than a certain height to make it easier to select the correct block interval to target when we're missing the start block.


src/blockchain.erl
```diff
@@ -595,6 +596,19 @@ has_block(Block, #blockchain{db=DB, blocks=BlocksCF,
             Error
     end.
 
+find_first_block_after(MinHeight, #blockchain{db=DB, heights=HeightsCF}) ->
+    {ok, Iter} = rocksdb:iterator(DB, HeightsCF, []),
```
Vagabond, line 600: You can just jump the iterator to the closest key with `{seek, <<MinHeight:64/integer-unsigned-big>>}`

PR #590, opened 2020-08-24
Fix dialyzer

vihu:


src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -780,11 +780,10 @@ get_gateway_owner(Address, Ledger) ->
     end.
 
 
-share_of_dc_rewards( Key, Vars=#{dc_remainder := DCRemainder}) ->
-    erlang:round(DCRemainder * ((maps:get(Key, Vars) / (maps:get(poc_challengers_percent, Vars) + maps:get(poc_challengees_percent, Vars) + maps:get(poc_witnesses_percent, Vars)))));
-share_of_dc_rewards(_, _) ->
-    %% DC remainder not set
-    0.
+share_of_dc_rewards(_Key, #{dc_remainder := 0}) ->
```
evanmcc, line 783: will this key always be present? do we maybe still need the catchall?

vihu: dialyzer gets mad and complains that it won't ever reach the catch-all

PR #592, opened 2020-08-25
bump h3 so we can clean properly again

evanmcc:


evanmcc:
see https://github.com/helium/erlang-h3/pull/21

PR #593, opened 2020-08-25
Add witness and receipt validation for poc txns

vihu:
- This adds witness and receipt validation depending on observed SNR and Frequency values guarded behind `poc_version >= 9` and `data_aggregation_version >= 2`
- [x] Merge https://github.com/helium/proto/pull/50 and update dep


src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -41,6 +41,9 @@
 
 -type txn_poc_receipts() :: #blockchain_txn_poc_receipts_v1_pb{}.
 -type deltas() :: [{libp2p_crypto:pubkey_bin(), {float(), float()}}].
+-type frequencies() :: [float()].
+
+-define(FREQS, [903.9, 904.1, 904.3, 904.5, 904.7, 904.9, 905.1, 905.3]).
```
Vagabond, line 46: So this is not quite finished, as the frequencies depend on the region

src/blockchain_utils.erl
```diff
@@ -244,6 +248,39 @@ free_space_path_loss(Loc1, Loc2) ->
...
+free_space_path_loss(Loc1, Loc2, Ledger) ->
+    %% TODO support regional parameters for non-US based hotspots
+    %% TODO support variable Dt,Dr values for better FSPL values
+    %% FSPL = 10log_10(Dt*Dr*((4*pi*f*d)/(c))^2)
+
+    case blockchain:config(?poc_version, Ledger) of
+        {ok, V} when V >= 9 ->
+            %% Do the new and correct calculation
+            Distance = blockchain_utils:distance(Loc1, Loc2),
+            10*math:log10((1.8*1.8)*math:pow((4*math:pi()*(?FREQUENCY*100000)*(Distance*100000))/(299792458), 2));
```
Vagabond, line 263: We need to parameterize the frequency here, which I guess we should be capturing along with channel/datarate.

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -519,7 +618,15 @@ good_quality_witnesses(Element, Ledger) ->
                                  WitnessGwLoc = blockchain_ledger_gateway_v2:location(WitnessGw),
                                  WitnessParentIndex = h3:parent(WitnessGwLoc, ParentRes),
                                  WitnessRSSI = blockchain_poc_witness_v1:signal(Witness),
-                                 FreeSpacePathLoss = blockchain_utils:free_space_path_loss(WitnessGwLoc, ChallengeeLoc),
+
+                                 FreeSpacePathLoss = case blockchain:config(?poc_version, Ledger) of
+                                                         {ok, V} when V >= 9 ->
+                                                             WitnessFreq = blockchain_poc_witness_v1:frequency(Witness),
```
Vagabond, line 624: What if we simply did not store "filtered" witnesses/receipts in the ledger and made good_quality_witnesses a no-op in v9? Then we could leave the 915 filtering in for v8 and earlier and avoid filling up the ledger with junk.

Vagabond: I think if we go down this road we'll need to re-run all the filtering we did before, or you'll end up in a situation where the FSPL check is fine but the RSSI/channel are not and we'll have to thread those as well.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1194,6 +1196,47 @@ update_gateway_oui(Gateway, OUI, Nonce, Ledger) ->
...
+insert_witnesses(PubkeyBin, Witnesses, Ledger) ->
+    case blockchain:config(?poc_version, Ledger) of
+        %% only works with poc-v9 and above
+        {ok, V} when V >= 9 ->
+            case ?MODULE:find_gateway_info(PubkeyBin, Ledger) of
+                {error, _}=Error ->
+                    Error;
+                {ok, GW0} ->
+                    GW1 = lists:foldl(fun(POCWitness, GW) ->
+                                              case erlang:is_record(POCWitness, blockchain_poc_witness_v1_pb) of
```
Vagabond, line 1211: You could simply match the record in the fun head here

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -891,6 +1039,212 @@ vars(Ledger) ->
...
+    case maps:get(CeilSNR, ?SNR_CURVE, undefined) of
+        undefined ->
+            scale_unknown_snr(CeilSNR);
+        V ->
+            V
+    end.
+
+-spec get_channels(Txn :: txn_poc_receipts(),
+                   Chain :: blockchain:blockchain()) -> {ok, [non_neg_integer()]} | {error, any()}.
+get_channels(Txn, Chain) ->
```
Vagabond, line 1211: You should probably supply the ledger here and not assume the chain and the ledger height you want here are the same.

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -891,6 +1039,212 @@ vars(Ledger) ->
...
+
+    Secret = ?MODULE:secret(Txn),
+    case blockchain_gateway_cache:get(Challenger, Ledger) of
+        {error, Reason}=Error ->
+            lager:warning([{poc_id, POCID}],
+                          "poc_receipts error find_gateway_info, challenger: ~p, reason: ~p",
+                          [Challenger, Reason]),
+            Error;
+        {ok, GwInfo} ->
+            LastChallenge = blockchain_ledger_gateway_v2:last_poc_challenge(GwInfo),
```
Vagabond, line 1225: This line could return the wrong value given a too-new ledger and a short poc interval

src/ledger/v1/blockchain_ledger_v1.erl
vihu: Updated in 72cad9e

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Added in 72cad9e

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Presumably passing in the ledger should fix this? 72cad9e

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -715,13 +715,12 @@ absorb(Txn, Chain) ->
...
                         %% This isn't ideal, but we need to do delta calculation _before_ we delete the poc
                         %% as new calculate_delta calls back into check_is_valid_poc
-                        lists:foldl(fun({Gateway, Delta}, _Acc) ->
-                                            blockchain_ledger_v1:update_gateway_score(Gateway, Delta, Ledger)
-                                    end,
-                                    ok,
-                                    ?MODULE:deltas(Txn, Chain));
-                        %% Rely on the poc gc to delete pocs
-                        %% blockchain_ledger_v1:delete_poc(LastOnionKeyHash, Challenger, Ledger);
+                        ok = lists:foldl(fun({Gateway, Delta}, _Acc) ->
```
Vagabond, line 718: so the intention here is to badmatch if the accumulator is not `ok`

vihu: oh no, not sure why I matched it to ok. I will remove it, good catch

vihu: Updated b6bf946

Vagabond: Why is it even a fold?

vihu: Could change it to a foreach?

evanmcc (approved):

Vagabond (approved):

PR #594, opened 2020-08-25
Restore backward compat for sc_open txn absorb

vihu:


Vagabond:
if this syncs to the head, +1 from me

PR #595, opened 2020-08-26
[DNM] Add snapshot load-height cli cmd

vihu:


src/cli/blockchain_cli_snapshot.erl
```diff
@@ -48,11 +50,12 @@ register_all_cmds() ->
 snapshot_usage() ->
     [["snapshot"],
      ["blockchain snapshot commands\n\n",
-      "  snapshot take   - Take a snapshot at the current ledger height.\n",
-      "  snapshot load   - Load a snapshot from a file.\n"
-      "  snapshot diff   - Load two snapshots from files and find changes.\n"
-      "  snapshot info   - Show information about a snapshot in a file.\n"
-      "  snapshot list   - Show information about the last 5 snapshots.\n"
+      "  snapshot take          - Take a snapshot at the current ledger height.\n",
+      "  snapshot load-file     - Load a snapshot from a file.\n"
```
evanmcc, line 54: one last nit.  could you put the commands on subsequent lines?

src/cli/blockchain_cli_snapshot.erl
```diff
@@ -184,3 +187,57 @@ snapshot_list(["snapshot", "list"], [], []) ->
     end;
 snapshot_list(_, _, _) ->
     usage.
+
+snapshot_load_height_cmd() ->
+    [
+     [["snapshot", "load-height"], [], [], fun snapshot_load_height/3],
+     [["snapshot", "load-height", '*'], [], [], fun snapshot_load_height/3],
+     [["snapshot", "load-height", '*'], [],
+      [{revalidate, [{shortname, "r"}, {longname, "revalidate"}]}],
```
evanmcc, line 196: does this allow us to load the most recent with the `revalidate` flag?

src/cli/blockchain_cli_snapshot.erl
vihu: will do

src/cli/blockchain_cli_snapshot.erl
```diff
@@ -184,3 +187,57 @@ snapshot_list(["snapshot", "list"], [], []) ->
...
+    end;
+snapshot_load_height(["snapshot", "load-height", BH], [], [{revalidate, _}]) ->
+    Chain = blockchain_worker:blockchain(),
+    {ok, Block} = blockchain:get_block(list_to_integer(BH), Chain),
+    case blockchain_block_v1:snapshot_hash(Block) of
+        <<>> ->
+            Text = io_lib:format("No snapshot at block height: ~p", [BH]),
+            [clique_status:alert([clique_status:text(Text)])];
+        Hash ->
+            ok = application:set_env(blockchain, force_resync_validation, true),
```
evanmcc, line 236: can we factor all three loads (file, height, lastest) into a single utility function so we can be sure that we're doing the same thing each time?

src/cli/blockchain_cli_snapshot.erl
vihu: Not yet, I was running into some more issues with the `load-height` command, trying to address those now. Once that's done, I'll add the support to load the most recent with the `revalidate` flag as well.

src/cli/blockchain_cli_snapshot.erl
vihu: I can try, each of them differ slightly in implementation though..

evanmcc: They shouldn't, though?  you should be doing the same routine for each once you have a snapshot and validation flag to work with, I think.

vihu: I see what you mean, not enough :tea: yet

evanmcc:
@vihu this looks good, are there still issues to resolve with merging it?

vihu:
> @vihu this looks good, are there still issues to resolve with merging it?

Yes, don't merge it yet. It doesn't quite do the intended job and I wasn't able to nail it down so abandoned it at the time. Let me rebase, update and retest the cli rework here.

vihu:
This doesn't work as intended, will revisit if need be in future.

PR #597, opened 2020-08-27
Track packet offer times and annotate corresponding packets

Vagabond:


src/state_channel/blockchain_state_channel_handler.erl
```diff
@@ -139,33 +139,34 @@ handle_data(client, Data, State) ->
     end,
     {noreply, State};
 handle_data(server, Data, State=#state{pending_packet_offers=PendingOffers, pending_offer_limit=PendingOfferLimit}) ->
+    Time = erlang:system_time(millisecond),
```
macpie, line 142: why not micro?

macpie: I guess milli is enough for what we are doing

Vagabond: Erlang timers are in milliseconds

PR #598, opened 2020-08-27
fix the last block test

evanmcc:


evanmcc:
feel free to close if this isn't right right fix for SC correctness reasons

Vagabond:
What exactly crashes out here?

evanmcc:
mismatch between the ledger and the block store causes the GC to crash out when it tries to get the block.

evanmcc:
fixed elsewhere.

PR #599, opened 2020-08-27
try to be more memory-efficient when loading snapshots

evanmcc:


src/blockchain.erl
```diff
@@ -1554,6 +1570,18 @@ get_snapshot(Height, #blockchain{db=DB, snapshots=SnapshotsCF}=Blockchain) ->
             Error
     end.
 
+-spec have_snapshot(integer(), blockchain()) ->
```
vihu, line 1573: spec would be wrong cuz `Error -> Error`

evanmcc: will fix

PR #600, opened 2020-08-29
Fix incorrect causality check

vihu:
- Fix the causality check in sc client
- Add a debug application env var to output conflicting scs from sc client
- Add `is_causally_newer` fun to state channel module, enhance eunits for the same

### DIFF
--- a/src/state_channel/blockchain_state_channels_client.erl
+++ b/src/state_channel/blockchain_state_channels_client.erl
@@ -138,7 +138,7 @@ handle_cast({banner, Banner, HandlerPid}, State=#state{pubkey_bin=PubkeyBin, sig
         BannerSC ->
             case is_valid_sc(BannerSC, State) of
                 {error, causal_conflict} ->
-                    lager:error("causal_conflict for banner sc: ~p", [BannerSC]),
+                    lager:error("causal_conflict for banner sc_id: ~p", [blockchain_state_channel_v1:id(BannerSC)]),
                     _ = libp2p_framed_stream:close(HandlerPid),
                     ok = append_state_channel(BannerSC, State),
                     {noreply, State};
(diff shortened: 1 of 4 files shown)


src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -744,13 +744,14 @@ is_causally_correct_sc(SC, State) ->
             false;
         {ok, [KnownSC]} ->
             %% Check if SC is causally correct
-            Check = (caused == blockchain_state_channel_v1:compare_causality(KnownSC, SC) orelse
-                     equal == blockchain_state_channel_v1:compare_causality(KnownSC, SC)),
-            lager:info("causality check: ~p, this sc: ~p, known_sc: ~p", [Check, SC, KnownSC]),
+            Check = (conflict /= blockchain_state_channel_v1:compare_causality(KnownSC, SC)),
```
mrallen1, line 747: :+1:

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -912,6 +928,22 @@ num_dcs_for(PubkeyBin, SC) ->
...
+-spec debug_multiple_scs(SC :: blockchain_state_channel_v1:state_channel(),
+                         KnownSCs :: [blockchain_state_channel_v1:state_channel()]) -> ok.
+debug_multiple_scs(SC, KnownSCs) ->
+    case application:get_env(blockchain, debug_multiple_scs, false) of
+        false ->
+            %% false by default, don't write it out
+            ok;
+        true ->
+            BinSC = term_to_binary(SC),
+            BinKnownSCs = term_to_binary(KnownSCs),
```
mrallen1, line 941: Does `known` mean previous or "current?" If so can we use that instead?

vihu: I mostly had this stanza for debugging from the files where I look at the causal relationship between `KnownSCs` and the `SC` we received to end up with multiple copies.

PR #605, opened 2020-09-03
Improve performance of client side validation of state channels

Vagabond:


vihu (commented):
Nice, can we add the `quick_compare_causality` to the eunits as well please? They serve as a decent sanity check imo.

PR #606, opened 2020-09-03
Fix causality check typos

vihu:


Vagabond:
If we can re-sync with this, then we can merge this. If it breaks sync it needs a chain var.

vihu:
Ok, I was successfully able to resync with revalidation with this from block `450001`

Vagabond:
If we can run another resync (since more SCs have closed since then), and it works, we should merge it

vihu:
> If we can run another resync (since more SCs have closed since then), and it works, we should merge it

Sure, I'll do past ~150K blocks or so

vihu:
> If we can run another resync (since more SCs have closed since then), and it works, we should merge it

Ok, was able to sync from block `450001` to the latest head with revalidation running this fix.

Vagabond:
we should merge it then?

vihu:
> we should merge it then?

We can I think, I've resynced twice with this (from block 450001). I can do one more run for completeness sake and go back to just before we activated state channels. If that syncs, _then_ we can 100% merge it. Will run it over the weekend.

vihu:
Ok @Vagabond I think this is mergable. Appears to sync just fine, I've synced twice from `450001` and once from `339457` (that covers sc_v1 channels too).

evanmcc (approved):

PR #607, opened 2020-09-04
Export an id type for use in the router

mrallen1:


src/transactions/v1/blockchain_txn_state_channel_open_v1.erl
```diff
@@ -38,9 +38,10 @@
 -endif.
 
 -type txn_state_channel_open() :: #blockchain_txn_state_channel_open_v1_pb{}.
--export_type([txn_state_channel_open/0]).
+-type id() :: binary().
```
vihu, line 41: I think we have a spec for id already defined in `blockchain_state_channel_v1`

mrallen1: I know we do but it's not right to tag _this_ transaction id as the same as the id assigned to a state channel, right?

vihu: Huh, I thought they were the same IDs, but ok. Not a huge deal to have this spec here I guess.

vihu (approved):

PR #611, opened 2020-09-10
Fix FSPL calculation

refugeesus:
Want to get this in with 593 if possible.

Fixing FSPL calculation. The actual value of FSPL is the power required at a minimum in free space for two radios to hear each other (sorta not really the definition).

Depending on how this value is used in other modules, we may want to subtract transmit power somewhere else? (i.e. calculating link budget of which FSPL is a component)

### DIFF
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -242,7 +242,10 @@ score_tagged_gateways(Height, Ledger) ->
 free_space_path_loss(Loc1, Loc2) ->
     Distance = blockchain_utils:distance(Loc1, Loc2),
     %% TODO support regional parameters for non-US based hotspots
-    ?TRANSMIT_POWER - (32.44 + 20*math:log10(?FREQUENCY) + 20*math:log10(Distance) - ?MAX_ANTENNA_GAIN - ?MAX_ANTENNA_GAIN).
+    %% TODO support variable Dt,Dr values for better FSPL values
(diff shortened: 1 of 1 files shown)


vihu:
We would need to guard the new fspl calculation behind a chain var, `poc_version >= 9` (since that's what's intended for #593) and propagate that throughout. Otherwise we'd break existing gateway histograms and poc receipt txns, I think.

vihu (approved):

src/blockchain_utils.erl
```diff
@@ -244,6 +245,39 @@ free_space_path_loss(Loc1, Loc2) ->
...
+free_space_path_loss(Loc1, Loc2, Ledger) ->
+    %% TODO support regional parameters for non-US based hotspots
+    %% TODO support variable Dt,Dr values for better FSPL values
+    %% FSPL = 10log_10(Dt*Dr*((4*pi*f*d)/(c))^2)
+
+    case poc_version(Ledger) of
+        {ok, V} when V >= 9 ->
+            %% Do the new and correct calculation
+            Distance = blockchain_utils:distance(Loc1, Loc2),
+            10*math:log10((1.8*1.8)*math:pow((4*math:pi()*(?FREQUENCY*100000)*(Distance*100000))/(299792458), 2));
```
Vagabond, line 260: We should parametrize the frequency here so it works in 868 and the 400 mhz regions

PR #620, opened 2020-09-16
Fix dialyzer for poc_id

madninja:


madninja:
this passes dialyzed but fails tests like it did before the poc_id breakage.

vihu (approved):

PR #623, opened 2020-09-18
Testing single arity get_channels

vihu:
I'm only adding this as a suggestion since we are using this in some simulations which (for now) does not actually use the chain at all instead solely relying on the ledger. Having a `get_channels/1` fun will be useful once we add the request block hash to poc receipts.


vihu:
Closing this in favor of https://github.com/helium/blockchain-core/pull/631

PR #624, opened 2020-09-18
fix snapshot diffs for serialization changes

evanmcc:
also do more work to filter out encoding-related non-changes.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -1433,6 +1433,7 @@ crosscheck(Blockchain, Recalc) ->
             Error
     end.
 
+-spec compare(LedgerA :: blockchain_ledger_v1:ledger(), LedgerB :: blockchain_ledger_v1:ledger()) -> ok | {error, any()}.
 compare(LedgerA, LedgerB) ->
     {ok, FPA} = blockchain_ledger_v1:raw_fingerprint(LedgerA, true),
     {ok, FPB} = blockchain_ledger_v1:raw_fingerprint(LedgerB, true),
--- a/src/ledger/v1/blockchain_ledger_gateway_v2.erl
+++ b/src/ledger/v1/blockchain_ledger_gateway_v2.erl
@@ -328,60 +328,60 @@ print(Address, Gateway, Ledger, Verbose) ->
     ] ++ Scoring.
 
 add_witness({poc_receipt,
-             WitnessPubkeyBin,
+             WitnessAddress,
              WitnessGW = #gateway_v2{nonce=Nonce},
              POCWitness,
              Gateway = #gateway_v2{witnesses=Witnesses}}) ->
     RSSI = blockchain_poc_receipt_v1:signal(POCWitness),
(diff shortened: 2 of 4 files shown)


src/blockchain.erl
```diff
@@ -1438,6 +1433,42 @@ crosscheck(Blockchain) ->
             Error
     end.
 
+compare(LedgerA, LedgerB) ->
```
vihu, line 1436:
```suggestion
-spec compare(LedgerA :: blockchain_ledger_v1:ledger(), LedgerB :: blockchain_ledger_v1:ledger()) -> ok | {error, any()}.
compare(LedgerA, LedgerB) ->
```

vihu: also wondering if `compare/2` should be put in ledger module..

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -590,8 +590,8 @@ raw_fingerprint(#ledger_v1{mode = Mode} = Ledger, Extended) ->
             false ->
                 {ok, #{<<"ledger_fingerprint">> => LedgerHash}};
             _ ->
-                [GWsHash, EntriesHash, DCEntriesHash, HTLCsHash,
-                 PoCsHash, SecuritiesHash, RoutingsHash, StateChannelsHash, SubnetsHash, _] = L,
+                [_, GWsHash, EntriesHash, DCEntriesHash, HTLCsHash,
```
vihu, line 593: Not sure I understand why we ignore `DefaultHash` here, isn't the same we put in the `core_fingerprint` value every time we calculated above?

evanmcc: all of these are calculated about, we're just matching out all the hashes here to have some bindings to use.  `DefaultHash` is already bound, so why force a match calculation when we rebind it?

vihu: makes sense, thanks

src/ledger/v1/blockchain_ledger_gateway_v2.erl
```diff
@@ -328,60 +328,60 @@ print(Address, Gateway, Ledger, Verbose) ->
...
-        {ok, Witness=#witness{nonce=Nonce, count=Count, hist=Hist}} ->
+    case lists:keytake(WitnessAddress, 1, Witnesses) of
+        {value, {_, Witness=#witness{nonce=Nonce, count=Count, hist=Hist}}, Witnesses1} ->
             %% nonce is the same, increment the count
-            Gateway#gateway_v2{witnesses=maps:put(WitnessPubkeyBin,
-                                                  Witness#witness{count=Count + 1,
-                                                                  hist=update_histogram(RSSI, Hist),
-                                                                  recent_time=TS},
-                                                  Witnesses)};
+            Gateway#gateway_v2{witnesses=[{WitnessAddress,
```
vihu, line 341: the witnesses list here in gateway_v2 does not need to be sorted?

PR #627, opened 2020-09-24
Calculate DC rewards in previous epoch grace blocks

mrallen1:
Problem to solve: Currently when a state channel closes in the grace block period before an epoch ends, that close is never considered for rewards _during_ that epoch. The next epoch does not look back at those blocks, so no rewards are generated from DC activity when a state channel close occurs during these blocks.

Solution: The next epoch reward cycle needs to look back into those blocks to find state channel closes and then reward them appropriately.

### DIFF
--- a/src/transactions/v1/blockchain_txn_rewards_v1.erl
+++ b/src/transactions/v1/blockchain_txn_rewards_v1.erl
@@ -157,7 +157,12 @@ absorb(Txn, Chain) ->
 calculate_rewards(Start, End, Chain) ->
     {ok, Ledger} = blockchain:ledger_at(End, Chain),
     Vars = get_reward_vars(Start, End, Ledger),
-    case get_rewards_for_epoch(Start, End, Chain, Vars, Ledger) of
+    %% Previously, if a state_channel closed in the grace blocks before an
+    %% epoch ended, then it wouldn't ever get rewarded.
+    {ok, PreviousGraceBlockDCRewards} = collect_dc_rewards_from_previous_epoch_grace(Start, End,
+                                                                                     Chain, Vars,
+                                                                                     Ledger),
+    case get_rewards_for_epoch(Start, End, Chain, Vars, Ledger, PreviousGraceBlockDCRewards) of
         {error, _Reason}=Error ->
             Error;
         {ok, POCChallengersRewards, POCChallengeesRewards, POCWitnessesRewards, DCRewards} ->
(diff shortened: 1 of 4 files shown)


src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -157,7 +157,12 @@ absorb(Txn, Chain) ->
 calculate_rewards(Start, End, Chain) ->
     {ok, Ledger} = blockchain:ledger_at(End, Chain),
     Vars = get_reward_vars(Start, End, Ledger),
-    case get_rewards_for_epoch(Start, End, Chain, Vars, Ledger) of
+    %% Previously, if a state_channel closed in the grace blocks before an
+    %% epoch ended, then it wouldn't ever get rewarded.
+    PreviousGraceBlockDCRewards = collect_dc_rewards_from_previous_epoch_grace(Start, End,
```
vihu, line 162: This can potentially give you an `{error, something}` looks like? We should handle that when passing it over?

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -704,6 +681,24 @@ normalize_witness_rewards(WitnessRewards, #{epoch_reward := EpochReward,
         WitnessRewards
     ).
 
+collect_dc_rewards_from_previous_epoch_grace(Start, End, Chain,
```
vihu, line 684: I know these are internal functions, but still feels like we should spec them. Thoughts?

vihu:
@mrallen1 do you think we can extend eunit `dc_rewards_v3_test` to invoke the changes here?

src/transactions/v1/blockchain_txn_rewards_v1.erl
mrallen1: yeah good point, thanks

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -704,6 +681,30 @@ normalize_witness_rewards(WitnessRewards, #{epoch_reward := EpochReward,
...
+                                             #{sc_grace_blocks := Grace,
+                                               reward_version := RV} = Vars,
+                                             Ledger) when RV > 3 ->
+    scan_grace_block(Start - Grace, Start, End, Vars, Chain, Ledger, #{});
+collect_dc_rewards_from_previous_epoch_grace(_Start, _End, _Chain, _Vars, _Ledger) -> {ok, #{}}.
+
+scan_grace_block(Current, Start, _End, _Vars, _Chain, _Ledger, Acc)
+                                           when Current == Start + 1 -> {ok, Acc};
+scan_grace_block(Current, Start, End, Vars, Chain, Ledger, Acc) ->
+    case blockchain:get_block(Current, Chain) of
```
Vagabond, line 698: We should check the snapshot generation code to make sure it pulls enough blocks to make this reliable.

Vagabond: https://github.com/helium/blockchain-core/blob/master/src/ledger/v1/blockchain_ledger_snapshot_v1.erl#L424 We probably need to extend this to make sure it covers the last *grace* blocks in the last epoch.

mrallen1: OK, good catch - thanks

vihu (approved):

Vagabond (approved):

PR #602, opened 2020-09-01
Use bloom filter to improve packet times

macpie:

PR #630, opened 2020-09-25
Rework state channel handler with offer queueing

vihu:
This has a portion of https://github.com/helium/blockchain-core/pull/602 without the bloom filter stuff just to test whether it makes offer/packet handling more robust


vihu:
Closing this as we'd land https://github.com/helium/blockchain-core/pull/602 which is a superset of the work here

PR #631, opened 2020-09-25
Add single arity get_channels

vihu:

PR #632, opened 2020-09-25
Add is_valid entries for poc_receipts/witnesses

madninja:
Passing in a ledger to txn_receipts_v1:to_json options will have the resulting json include is_valid entries for the receipts and witnesses.


madninja:
Rebase on #631 once it merges

vihu (approved):
This appears to work and resync

PR #635, opened 2020-09-29
client side fix for v1 state channels

evanmcc:
when we went to state channel v2, we broke v1's by requiring a banner to send any packets.  This code tries to fix this by sending queued packets on dial success instead of waiting for a banner message.


Vagabond (approved):

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -249,7 +218,7 @@ handle_info({dial_success, AddressOrOUI, Stream}, #state{chain=undefined}=State)
 handle_info({dial_success, OUIOrAddress, Stream}, State0) ->
     erlang:monitor(process, Stream),
     State1 = add_stream(OUIOrAddress, Stream, State0),
-    {noreply, State1};
+    {noreply, maybe_send_packets(OUIOrAddress, Stream, State1)};
```
Vagabond, line 221: Actually we might not want to do this in every case, only in v1's case

vihu (approved):
lgtm

PR #638, opened 2020-10-01
move purchase to a call

evanmcc:
we need some backpressure here, to avoid OOMing due to large purchase messages piling up.


Vagabond (approved):
This looks like it should help by leaving packets on the socket until we can handle them.

vihu (approved):

mrallen1 (approved):
lgtm - should give us some room to explore other solutions/ideas

PR #642, opened 2020-10-05
Fix nonce state-keeping

evanmcc:
make sure that we update the nonce on v2 as well because the root hash is also changing, otherwise this update gets ignored sometimes


vihu (commented):
I think this must be guarded behind a chain var, eventual state channel(s) show up in the sc close txn, although since this happens in `process_packet` maybe it's fine? We should definitely run a validation resync though just to be certain..

evanmcc:
I don't think that it needs a var.  Mostly everyone would be picking this update up later with the next packet, and the state channel server side always does the right thing, since it's pulling directly from the state here.

I agree that we should sync it back a ways just to make sure.

vihu (approved):
lgtm, confirmed resync as well :+1:

Vagabond:
Yes this should not require a resync as it affects the production path of state channels, not their validation.

Vagabond (approved):

PR #643, opened 2020-10-06
When closing a disputed state channel slash overcommit & normalize payouts

Vagabond:
Prior to this change, gated behind rewards v4, when a state channel
closed in dispute state, the DCs were paid out 1:1, even if the channel
closed with a total above the amount of DC in the state channel.
Additionally, the overcommit would be shared between all the
participants of the state channel, but that incentivizes the router to
pack the state channel with affiliated hotspots to try to recover some
of the overcommit and payout. With v4 we fully slash the overcommit and
proportionally pay out the participants (if the state channel is
overpaid), after truncating any large participant payouts to the size of
the state channel.


src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -302,6 +304,32 @@ add_payload(Payload, SC, Skewed) ->
     end.
 
 
+normalize(#blockchain_state_channel_v1_pb{summaries=Summaries}=SC) ->
+    Total = amount(SC),
```
vihu, line 308: Shouldn't this be `total_dcs(SC)`?

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -302,6 +304,32 @@ add_payload(Payload, SC, Skewed) ->
...
+                                                           {AccTotal + Amt, [Summary|AccSummaries]}
+                                                   end
+                                           end, {0, []}, Summaries),
+
+    %% then scale rewards proportionally if they collectively sum to more than the total amount
+    FinalSummaries = case SumTotal > Total of
+                         true ->
+                             lists:map(fun(Summary) ->
+                                               %% use trunc here so rounding up cannot inflate DC counts
+                                               NewAmt = trunc((blockchain_state_channel_summary_v1:num_dcs(Summary) / SumTotal) * Total),
```
vihu, line 325: There's a non-zero chance that `SumTotal` here may be equal to `0` and :boom:

evanmcc: in this case `SumTotal` is greater than `Total`, is there a chance that `Total` will ever be negative?  if not I think we're OK wrt zero

src/state_channel/v1/blockchain_state_channel_v1.erl
vihu: Unless this is the "credits" that the state channel was funded with, in which case it's fine as is.

Vagabond: Yes, it's the total amount the SC was funded with (not including the overcommit)

src/state_channel/v1/blockchain_state_channel_v1.erl
Vagabond: Yes, it'd be possible to get 0 here. Remember that this is not the SC that appears on-chain, this is a temporary normalized form that we only use when calculating rewards so it doesn't really matter if 0 pops out here, I don't think.

Vagabond:
Please add tests.

Vagabond:
This needs tests.

vihu (approved):

PR #649, opened 2020-10-13
re-add majority voting

evanmcc:
reverting the old unanimous voting code.

### DIFF
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -42,6 +42,7 @@
     index_of/2,
 
     verify_multisig/3,
+    count_votes/3,
     poc_per_hop_max_witnesses/1
 ]).
 
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -472,7 +472,7 @@ poc_per_hop_max_witnesses(Ledger) ->
 
 
 majority(N) ->
-    N div 2 + 1.
+    (N div 2) + 1.
 
 %% ------------------------------------------------------------------
 %% EUNIT Tests
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -430,10 +430,9 @@ verify_multisig(Artifact, Sigs, Keys) ->
     Total = length(Keys),
     lager:debug("sigs ~p keys ~p", [Sigs, Keys]),
     Votes = count_votes(Artifact, Keys, Sigs),
-    %% this code is still good, uncomment to get majority voting
-    %% Majority = majority(Total),
-    %% lager:debug("votes ~p, majority: ~p", [Votes, Majority]),
-    Votes == Total.
+    Majority = majority(Total),
+    lager:info("votes ~p, majority: ~p", [Votes, Majority]),
+    Votes >= Majority.
 
 count_votes(Artifact, MultiKeys, Proofs) ->
     count_votes(Artifact, MultiKeys, Proofs, 0).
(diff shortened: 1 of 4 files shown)


src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -419,21 +419,24 @@ validate_master_keys(Txn, Gen, Artifact, Ledger) ->
...
                     KeyProofs =
                         case multi_key_proofs(Txn) of
                             [] ->
                                 throw({error, no_multi_keys_proofs});
                             Ps ->
                                 Ps
                         end,
-                    [case verify_key(Artifact, Key, KeyProof) of
+                    Votes = blockchain_utils:count_votes(Artifact, MultiKeys, KeyProofs),
+                    case Votes >= length(ProofKeys) of
```
Vagabond, line 432: What are we actually checking for here, maybe add some comments?

evanmcc: will do.

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -279,13 +279,13 @@ is_valid(Txn, Chain) ->
                                 %% handle the case where this gets set before
                                 %% the keys are set
                                 case blockchain_ledger_v1:multi_keys(Ledger) of
-                                    {ok, MultiKeys} ->
+                                    {ok, MultiKeys} when MultiKeys /= [] ->
```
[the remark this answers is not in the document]
madninja: ah ok... however the behavior still is that if there are _no_ multi keys we'd fall back to the master_key?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -419,21 +419,37 @@ validate_master_keys(Txn, Gen, Artifact, Ledger) ->
...
+                    %% keys, and to ensure that all new keys have one proof associated with them.
+                    {ok, OldMultiKeys} = blockchain_ledger_v1:multi_keys(Ledger),
+                    %% remove all existing keys from the new list. They don't need to re-prove
+                    %% themselves and cannot 'vote' for the change here.  Their votes as to the
+                    %% signedness of the transaction are counted elsewhere.
+                    ProofKeys = MultiKeys -- OldMultiKeys,
                     KeyProofs =
-                        case multi_key_proofs(Txn) of
+                        %% deduplicate the proofs
+                        case lists:usort(multi_key_proofs(Txn)) of
```
Vagabond, line 434: Would they be duplicated in normal operation somehow, or is this in case of an attack?

Vagabond: ECDSA does not give deterministic signatures, so deduping will not remove repeated signatures by the same key. EdDSA *is* deterministic happily, so this is not a concern with ed25519 keys.

Vagabond (approved):

src/blockchain_utils.erl
```diff
@@ -464,8 +471,8 @@ poc_per_hop_max_witnesses(Ledger) ->
     end.
 
 
-%% majority(N) ->
-%%     N div 2 + 1.
+majority(N) ->
+    N div 2 + 1.
```
vihu, line 475:
```suggestion
    (N div 2) + 1.
```

src/transactions/v1/blockchain_txn_vars_v1.erl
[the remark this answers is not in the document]
evanmcc: yes

Vagabond (approved):

vihu (approved):

PR #652, opened 2020-10-14
Fix state channel suite

vihu:
The only important change here is test specific, just adding an amount when opening a state channel.


test/blockchain_state_channel_SUITE.erl
```diff
@@ -1521,15 +1521,15 @@ multi_sc_gc_test(Config) ->
     DevNonce0 = crypto:strong_rand_bytes(2),
     Packet0 = blockchain_ct_utils:join_packet(?APPKEY, DevNonce0, 0.0),
     ok = ct_rpc:call(GatewayNode1, blockchain_state_channels_client, packet, [Packet0, [], 'US915']),
+    timer:sleep(timer:seconds(3)),
```
evanmcc, line 1524: wouldn't it make more sense to add extra time to the wait right below this?

Vagabond (approved):

PR #653, opened 2020-10-15
Add timeout for outbound dials

vihu:
On my local system I noticed that the sc client state would keep packets in the waiting queue forever if for some reason the outbound dial to one of the router(s) was either unverified or stuck; this should clear it after 30s. It's debatable whether we should be dropping them but if you can't succeed in 30s there's not much to do there.

### DIFF
--- a/src/state_channel/blockchain_state_channels_client.erl
+++ b/src/state_channel/blockchain_state_channels_client.erl
@@ -552,7 +552,7 @@ dial(Swarm, Address) when is_list(Address) ->
                       ok;
                   {'DOWN', R, process, P, _Reason} ->
                       Self ! {dial_fail, Address, _Reason}
-              after 30000 ->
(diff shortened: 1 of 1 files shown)


src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -552,6 +552,9 @@ dial(Swarm, Address) when is_list(Address) ->
                       ok;
                   {'DOWN', R, process, P, _Reason} ->
                       Self ! {dial_fail, Address, _Reason}
+              after 30000 ->
```
evanmcc, line 555: make this an env parameter please, so we can tune it for tests.

vihu: something like `application:get_env(blockchain, sc_packet_dial_timeout, 30000)` should do the trick?

evanmcc (approved):

PR #654, opened 2020-10-15
Filter downed packet streams

vihu:
I've altered `find_stream` in the state channel client to `find_alive_stream`; this should return `undefined` when the stream has died for any reason and it went unhandled in the client. This should also allow `handle_packet` to add packet to the waiting queue instead of trying to send an offer over a dead stream.

Open to any suggestions. Also, wondering if we should also invoke `delete_stream` if the stream process isn't alive?

### DIFF
--- a/src/state_channel/blockchain_state_channels_client.erl
+++ b/src/state_channel/blockchain_state_channels_client.erl
@@ -307,13 +307,19 @@ handle_info({blockchain_event, {add_block, BlockHash, _Syncing, Ledger}},
                                            end
                                    end, [], SCs),
     {noreply, State#state{pending_closes=lists:usort(PendingCloses ++ ClosingChannels ++ ExpiringChannels)}};
-handle_info({'DOWN', _Ref, process, Pid, _}, #state{streams=Streams}=State) ->
+handle_info({'DOWN', _Ref, process, Pid, _}, #state{streams=Streams, packets=Packets}=State) ->
     FilteredStreams = maps:filter(fun(_Name, {unverified, Stream}) ->
                                           Stream /= Pid;
                                      (_Name, Stream) ->
                                           Stream /= Pid
                                   end, Streams),
-    {noreply, State#state{streams=FilteredStreams}};
+
+    %% Keep the streams which don't have downed pid, given we're monitoring correctly
+    FilteredPackets = maps:filter(fun(StreamPid, _PacketQueue) ->
+                                          StreamPid /= Pid
+                                  end, Packets),
+
+    {noreply, State#state{streams=FilteredStreams, packets=FilteredPackets}};
 handle_info(_Msg, State) ->
     lager:warning("rcvd unknown info msg: ~p", [_Msg]),
     {noreply, State}.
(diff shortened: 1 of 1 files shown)


vihu:
I guess we can make `find_alive_stream` return `{stream_val(), state()}` and call delete stream if the process is dead maybe?

Vagabond:
But we monitor the stream, in what case does the DOWN message not fire?

vihu:
> But we monitor the stream, in what case does the DOWN message not fire?

Hmm, right, it looks like when the stream goes down normally, 1. we don't delete it from the streams/packets maps? 2. we don't send ourselves a msg in the client either.

vihu:
Oh, maybe the real problem here is that when we do get a `DOWN` for a stream we don't filter the other maps (`packets` map specifically), we only filter out the `streams` map.

Vagabond:
> Oh, maybe the real problem here is that when we do get a `DOWN` for a stream we don't filter the other maps (`packets` map specifically), we only filter out the `streams` map.

This sounds worth investigating.

Vagabond:
So how does this help, exactly?

vihu:
> So how does this help, exactly?

This just ensures that if there is a `pid -> packet_queue` entry in the packets map, the pid is an alive process. There isn't any particular check in `handle_packet` in sc client that if it found a stream, it would just try send an offer/packet, even if the stream is a dead process.

With the fix:

```
(miner@127.0.0.1)14> [erlang:is_process_alive(P) || P <- maps:keys(erlang:element(9, sys:get_state(blockchain_state_channels_client)))].
[true,true]
(miner@127.0.0.1)15> [erlang:exit(P, kill) || P <- maps:keys(erlang:element(9, sys:get_state(blockchain_state_channels_client)))].      
[true,true]
(miner@127.0.0.1)16> [erlang:is_process_alive(P) || P <- maps:keys(erlang:element(9, sys:get_state(blockchain_state_channels_client)))].
[true]
(miner@127.0.0.1)17> [erlang:is_process_alive(P) || P <- maps:keys(erlang:element(9, sys:get_state(blockchain_state_channels_client)))].
[true,true]
```

Without the fix:

```
(miner@127.0.0.1)3> [erlang:is_process_alive(P) || P <- maps:keys(erlang:element(9, sys:get_state(blockchain_state_channels_client)))].
[true,true]
(miner@127.0.0.1)4> [erlang:exit(P, kill) || P <- maps:keys(erlang:element(9, sys:get_state(blockchain_state_channels_client)))].      
[true,true]
(miner@127.0.0.1)5> [erlang:is_process_alive(P) || P <- maps:keys(erlang:element(9, sys:get_state(blockchain_state_channels_client)))].
[false,false,true]
(miner@127.0.0.1)6> [erlang:is_process_alive(P) || P <- maps:keys(erlang:element(9, sys:get_state(blockchain_state_channels_client)))].
[false,false,true]
(miner@127.0.0.1)7> [erlang:is_process_alive(P) || P <- maps:keys(erlang:element(9, sys:get_state(blockchain_state_channels_client)))].
[false,false,true,true]
```

Vagabond (approved):

PR #659, opened 2020-10-20
Implement transfer hotspot transaction

mrallen1:
Problem to solve: implement the transfer hotspot transaction

CH: https://app.clubhouse.io/hlm/story/8742/transfer-ownership-transaction
Proto: https://github.com/helium/proto/pull/54


src/transactions/v1/blockchain_txn_transfer_hotspot_v1.erl
```diff
@@ -0,0 +1,314 @@
...
+    case {?MODULE:is_valid_seller(Txn),
+          ?MODULE:is_valid_buyer(Txn),
+          (is_integer(HNT) andalso HNT >= 0),
+          seller_owns_gateway(Txn, Ledger),
+          buyer_has_enough_hnt(Txn, Ledger)} of
+        {false, _, _, _, _} -> {error, bad_seller_signature};
+        {_, false, _, _, _} -> {error, bad_buyer_signature};
+        {_, _, false, _, _} -> {error, invalid_hnt_to_seller};
+        {_, _, _, false, _} -> {error, gateway_not_owned_by_seller};
+        {_, _, _, _, false} -> {error, buyer_insufficient_hnt_balance};
```
vihu, line 167: We should also invalidate `buyer=seller` transactions for ownership transfer

mrallen1: Ah, yeah good idea

vihu:
`blockchain_txn.erl` needs some updates, specifically missing type definition for hotspot transfer txn and wrap_txn.

vihu:
@mrallen1 I've updated `blockchain_txn` and added a basic common test to check validity of the txn.

src/blockchain_utils.erl
```diff
@@ -470,6 +471,22 @@ poc_per_hop_max_witnesses(Ledger) ->
             ?POC_PER_HOP_MAX_WITNESSES
     end.
 
+%%--------------------------------------------------------------------
+%% @doc Given a list of tuples of zero arity functions that return a
```
Vagabond, line 475: A railroad by any other name would smell as sweet?

src/transactions/v1/blockchain_txn_transfer_hotspot_v1.erl
```diff
@@ -0,0 +1,319 @@
...
+                  {fun() -> is_integer(HNT) andalso HNT >= 0 end,
+                                          {error, invalid_hnt_to_seller}},
+                  {fun() -> seller_owns_gateway(Txn, Ledger) end,
+                                          {error, gateway_not_owned_by_seller}},
+                  {fun() -> buyer_has_enough_hnt(Txn, Ledger) end,
+                                          {error, buyer_insufficient_hnt_balance}}],
+    case blockchain_utils:fold_condition_checks(Conditions) of
+        {error, _Reason} = Err -> Err;
+        ok ->
+            AreFeesEnabled = blockchain_ledger_v1:txn_fees_active(Ledger),
```
Vagabond, line 174: Can we not do the fees as a condition as well?

src/transactions/v1/blockchain_txn_transfer_hotspot_v1.erl
```diff
@@ -0,0 +1,319 @@
...
+    Fee = ?MODULE:fee(Txn),
+    HNTToSeller = ?MODULE:hnt_to_seller(Txn),
+
+    {ok, GWInfo} = blockchain_gateway_cache:get(Gateway, Ledger),
+    %% fees here are in DC (and perhaps converted to HNT automagically)
+    case blockchain_ledger_v1:debit_fee(Buyer, Fee, Ledger, AreFeesEnabled) of
+        {error, _Reason} = Error -> Error;
+        ok ->
+            %% this call confusingly accepts HNT not DC values
+            ok = blockchain_ledger_v1:debit_fee_from_account(Buyer, HNTToSeller, Ledger),
```
Vagabond, line 199: Would it be better to use `debit_from_account` and require a nonce to prevent replay attacks? Imagine if A transfers a hotspot to B and then B transfers it back to A (perhaps a refund/RMA). The original A->B transaction could be replayed by anyone, including B, no?

Vagabond:
This all looks ok now, what we are missing is actually applying this txn to a chain and checking everything changes as intended.

Also I think this branch needs a rebase.

mrallen1:
ok ill rebase here and add miner integration test(s)

mrallen1:
rebased and force pushed

src/transactions/blockchain_txn.erl
```diff
@@ -185,7 +187,9 @@ wrap_txn(#blockchain_txn_state_channel_close_v1_pb{}=Txn) ->
 wrap_txn(#blockchain_txn_price_oracle_v1_pb{}=Txn) ->
     #blockchain_txn_pb{txn={price_oracle_submission, Txn}};
 wrap_txn(#blockchain_txn_gen_price_oracle_v1_pb{}=Txn) ->
-    #blockchain_txn_pb{txn={gen_price_oracle, Txn}}.
+    #blockchain_txn_pb{txn={gen_price_oracle, Txn}};
+wrap_txn(#blockchain_txn_transfer_hotspot_v1_pb{}=Txn) ->
+    #blockchain_txn_pb{txn={hotspot_transfer, Txn}}.
```
vihu, line 192:
```suggestion
    #blockchain_txn_pb{txn={transfer_hotspot, Txn}}.
```

vihu: proto says `transfer_hotspot`

test/blockchain_hotspot_transfer_SUITE.erl
```diff
@@ -0,0 +1,304 @@
...
+    SellerSignedTxn = blockchain_txn_transfer_hotspot_v1:sign_seller(Txn, SellerSigFun),
+    BuyerSignedTxn = blockchain_txn_transfer_hotspot_v1:sign_buyer(SellerSignedTxn, BuyerSigFun),
+    ct:pal("SignedTxn: ~p", [BuyerSignedTxn]),
+    ct:pal("IsValidSeller: ~p", [blockchain_txn_transfer_hotspot_v1:is_valid_seller(BuyerSignedTxn)]),
+
+    {error, gateway_not_owned_by_seller} = blockchain_txn:is_valid(BuyerSignedTxn, Chain),
+
+    ok.
+
+replay_transaction(Config) ->
```
vihu, line 270: This is good, but we should also do a test by manually creating a block via `test_utils:create_block/2`, similar to some tests done in `payment_v2` for example. So we are sure that the replayed txn doesn't appear in a subsequent block. Although it's a bit tricky as the "seller" won't own the hotspot once the first valid txn appears, but atleast we'll know that they appear in a block. (This is how I found out about the proto not lining up in transactions btw :wink: )

Vagabond: Even better, try a transfer of A->B then a transfer from B->A and try to replay the original transaction.

src/transactions/v1/blockchain_txn_transfer_hotspot_v1.erl
```diff
@@ -0,0 +1,357 @@
...
+        {ok, GwInfo} ->
+            GwOwner = blockchain_ledger_gateway_v2:owner_address(GwInfo),
+            Seller == GwOwner
+    end.
+
+-spec gateway_not_stale(txn_transfer_hotspot(), blockchain_ledger_v1:ledger()) -> boolean().
+gateway_not_stale(#blockchain_txn_transfer_hotspot_v1_pb{gateway=GW}, Ledger) ->
+    {ok, StaleHeight} = blockchain_ledger_v1:config(?transfer_hotspot_stale_poc_blocks, Ledger),
+    case blockchain_gateway_cache:get(GW, Ledger) of
+        {error, _} -> false;
```
Vagabond, line 249: So this transaction will not work until the chain var is set?

mrallen1: As it's written now, yes - I could add a default value to use here if we want, or I guess the default value could be 0

src/transactions/v1/blockchain_txn_transfer_hotspot_v1.erl
```diff
@@ -244,7 +246,7 @@ seller_owns_gateway(#blockchain_txn_transfer_hotspot_v1_pb{gateway=GW,
 
 -spec gateway_not_stale(txn_transfer_hotspot(), blockchain_ledger_v1:ledger()) -> boolean().
 gateway_not_stale(#blockchain_txn_transfer_hotspot_v1_pb{gateway=GW}, Ledger) ->
-    {ok, StaleHeight} = blockchain_ledger_v1:config(?transfer_hotspot_stale_poc_blocks, Ledger),
+    StaleInterval = get_config_or_default(?transfer_hotspot_stale_poc_blocks, Ledger),
```
vihu, line 249: nit, maybe just do something like `get_staleness_var_or_default(Ledger)`, it doesn't necessarily need to be arity 2 imo.

vihu (approved):
I think this looks good now. Nice work!

PR #662, opened 2020-10-22
Add support for HIP15

vihu:
Refer: https://github.com/helium/HIP/blob/master/0015-beaconing-rewards.md

- Introduce two new chain variables `poc_reward_decay_rate` (r) and `witness_redundancy` (n)
- The reward formula for transmitter (from hip15) is essentially the poc_challengee_reward unit and the formula for receiver is the poc_witness_reward unit
- There are no defaults here for either of the new chain variables, so we must set both in order for the change to trigger

TODO:
- [x] Fix eunits
- [x] Add more tests

### DIFF
--- a/src/transactions/v1/blockchain_txn_rewards_v1.erl
+++ b/src/transactions/v1/blockchain_txn_rewards_v1.erl
@@ -25,7 +25,8 @@
     absorb/2,
     calculate_rewards/3,
     print/1,
-    to_json/2
+    to_json/2,
+    legit_witnesses/6
 ]).
 
 -ifdef(TEST).
@@ -202,10 +203,12 @@ calculate_rewards(Start, End, Chain) ->
 %%--------------------------------------------------------------------
 -spec print(txn_rewards()) -> iodata().
 print(undefined) -> <<"type=rewards undefined">>;
-print(#blockchain_txn_rewards_v1_pb{start_epoch=Start, end_epoch=End,
+print(#blockchain_txn_rewards_v1_pb{start_epoch=Start,
+                                    end_epoch=End,
                                     rewards=Rewards}) ->
+    PrintableRewards = [blockchain_txn_reward_v1:print(R) || R <- Rewards],
     io_lib:format("type=rewards start_epoch=~p end_epoch=~p rewards=~p",
-                  [Start, End, Rewards]).
+                  [Start, End, PrintableRewards]).
 
 -spec to_json(txn_rewards(), blockchain_json:opts()) -> blockchain_json:json_object().
 to_json(Txn, _Opts) ->
@@ -291,6 +294,17 @@ get_reward_vars(Start, End, Ledger) ->
         {ok, R4} -> R4;
         _ -> 1
     end,
+
+    WitnessRedundancy = case blockchain:config(?witness_redundancy, Ledger) of
+                            {ok, WR} -> WR;
+                            _ -> undefined
+                        end,
+
+    DecayRate = case blockchain:config(?poc_reward_decay_rate, Ledger) of
+                    {ok, R} -> R;
+                    _ -> undefined
+                end,
+
     EpochReward = calculate_epoch_reward(Start, End, Ledger),
     #{
         monthly_reward => MonthlyReward,
@@ -306,7 +320,9 @@ get_reward_vars(Start, End, Ledger) ->
         sc_grace_blocks => SCGrace,
         sc_version => SCVersion,
         poc_version => POCVersion,
-        reward_version => RewardVersion
+        reward_version => RewardVersion,
+        witness_redundancy => WitnessRedundancy,
+        poc_reward_decay_rate => DecayRate
     }.
 
 -spec calculate_epoch_reward(pos_integer(), pos_integer(), blockchain_ledger_v1:ledger()) -> float().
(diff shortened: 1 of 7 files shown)


src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -291,6 +291,12 @@ get_reward_vars(Start, End, Ledger) ->
         {ok, R4} -> R4;
         _ -> 1
     end,
+
+    WitnessRedundancy = case blockchain:config(?witness_redundancy, Ledger) of
+                            {ok, WR} -> WR;
+                            _ -> 1
```
Vagabond, line 297: Still not sure why we just don't make this 5?

Vagabond:
Here are the scaling factors from HIP 15 in erlang code:

```
(N - (1 - math:pow(R, W-N)))/W.

1 + (1 - math:pow(R, W-N))
```

abhay:
relevant GH Issue #673

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -291,6 +293,17 @@ get_reward_vars(Start, End, Ledger) ->
...
     end,
+
+    WitnessRedundancy = case blockchain:config(?witness_redundancy, Ledger) of
+                            {ok, WR} -> WR;
+                            _ -> ?DEFAULT_WITNESS_REDUNDANCY
+                        end,
+
+    DecayRate = case blockchain:config(?poc_reward_decay_rate, Ledger) of
+                    {ok, R} -> R;
+                    _ -> 0
```
Vagabond, line 304: Guessing this should be 1.0 by default?

test/blockchain_reward_SUITE.erl
```diff
@@ -0,0 +1,521 @@
...
+        Challenger,
+        <<"secret">>,
+        <<"onion_key_hash">>,
+        <<"block_hash">>,
+        [P1, P2, P3]
+    ),
+    ct:pal("Txn: ~p", [Txn]),
+
+    %% Construct a block for the poc receipt txn WITHOUT validation
+    {ok, Block2} = test_utils:create_block(ConsensusMembers, [Txn], #{}, false),
```
Vagabond, line 445: we disable validation so we don't need to deal with pathing?

Vagabond (commented):
Initial impression of this looks very good. I will try to go through it more deeply soon.

test/blockchain_reward_SUITE.erl
vihu: Yeah, pretty much. Also didn't really see any need to do a poc request then receipt cycle since we only care about the rewards here. lmk if that's acceptable...

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -511,18 +519,32 @@ poc_challengees_rewards_(Version, [Elem|Path], StaticPath, Txn, Chain, Ledger, I
                            %% while we don't have a receipt for this node, we do know
                            %% there were witnesses or the path continued which means
                            %% the challengee transmitted
-                           maps:put(Challengee, I+1, Acc0);
+                           case poc_challengee_reward_unit(WitnessRedundancy, DecayRate, Witnesses) of
+                               {error, _} ->
+                                   %% Old behavior
+                                   maps:put(Challengee, I+1, Acc0);
+                               {ok, ToAdd} ->
+                                   lager:info("+++++++++++MARKER1"),
```
evanmcc, line 527: rm debugging?

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -604,11 +677,21 @@ poc_witnesses_rewards(Transactions,
                                               [] ->
                                                   Acc1;
                                               ValidWitnesses ->
+                                                  ToAdd = case {WitnessRedundancy, DecayRate} of
```
evanmcc, line 680: can we inline some of the documentation here?  it's clear enough what it's doing, but I'm at sea wrt why

src/transactions/v1/blockchain_txn_rewards_v1.erl
vihu: My bad, removed in 11ad15a

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -826,6 +909,47 @@ normalize_dc_rewards(DCRewards0, #{epoch_reward := EpochReward,
...
+                %% Get channels without validation
+                {ok, Channels} = blockchain_txn_poc_receipts_v1:get_channels(Txn, Chain),
+                ElemPos = blockchain_utils:index_of(Elem, StaticPath),
+                WitnessChannel = lists:nth(ElemPos, Channels),
+                ValidWitnesses = blockchain_txn_poc_receipts_v1:valid_witnesses(Elem, WitnessChannel, Ledger),
+                lager:debug("ValidWitnesses: ~p",
+                            [[blockchain_utils:addr2name(blockchain_poc_witness_v1:gateway(W)) || W <- ValidWitnesses]]),
+                ValidWitnesses
+            catch What:Why:ST ->
+                      lager:error("failed to calculate poc_challengees_rewards, error ~p:~p:~p", [What, Why, ST]),
```
evanmcc, line 937: is it possible to do this differently depending on how many blocks back you have?  in which case we might want to determine if we need more snap history to avoid drift

src/transactions/v1/blockchain_txn_rewards_v1.erl
vihu: Ok, added 5f87842

src/transactions/v1/blockchain_txn_rewards_v1.erl
vihu: Well, this isn't _new_ code, I simply moved this piece out in order to meck it in the test. I will however run a resync with revalidation with this just to be certain.

evanmcc (approved):

wavyog:
I know it's going to work

PR #665, opened 2020-10-27
Improvements to handling of failed txns

andymck:
Handle exit msgs when checking the result of txn validations

Added a fallback error handler

Added tests

### DIFF
--- a/src/transactions/blockchain_txn.erl
+++ b/src/transactions/blockchain_txn.erl
@@ -299,6 +299,7 @@ separate_res([{T, ok} | Rest], Chain, V, I) ->
             lager:warning("invalid txn while absorbing ~p : ~p / ~s", [type(T), Reason, print(T)]),
             separate_res(Rest, Chain, V, [{T, Reason} | I])
     end;
+
 separate_res([{T, Err} | Rest], Chain, V, I) ->
     case Err of
         {error, {bad_nonce, {_NonceType, Nonce, LedgerNonce}}} when Nonce > LedgerNonce + 1 ->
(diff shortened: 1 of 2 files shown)


test/blockchain_simple_SUITE.erl
```diff
@@ -2363,6 +2365,147 @@ replay_oui_test(Config) ->
...
+    {[], [{SignedPoCReqTxn0, crash_in_pmap}]} = blockchain_txn:validate([SignedPoCReqTxn0], Chain),
+
+    %% meck out the pmap function to simulate it returning an error msg in an unexpected format {error, unexpected_reason, unexpected_details}
+    %% this will default to a generic error msg
+    meck:expect(blockchain_utils, pmap, fun(_, _) -> [{SignedPoCReqTxn0, {crash, something_unexpected}}] end),
+    {[], [{SignedPoCReqTxn0, txn_failed}]} = blockchain_txn:validate([SignedPoCReqTxn0], Chain),
+
+    ?assert(meck:validate(blockchain_utils)),
+    meck:unload(blockchain_utils),
+
```
Vagabond, line 2505: Might be nice to test once all the mecks are gone things work normally as well?

src/transactions/blockchain_txn.erl
```diff
@@ -310,7 +311,18 @@ separate_res([{T, Err} | Rest], Chain, V, I) ->
...
+        {'EXIT', {{_Why,{error, CrashReason}}, _Stack}} when is_atom(CrashReason)->
+            lager:warning("crashed txn ~p : ~p / ~s", [type(T), CrashReason, print(T)]),
+            %% any other error means we drop it
+            separate_res(Rest, Chain, V, [{T, CrashReason} | I]);
+        {'EXIT', CrashReason} when is_atom(CrashReason)->
+            lager:warning("crashed txn ~p : ~p / ~s", [type(T), CrashReason, print(T)]),
+            %% any other error means we drop it
+            separate_res(Rest, Chain, V, [{T, CrashReason} | I]);
+        Error->
+            lager:warning("failed txn ~p : ~p / ~s", [type(T), Error, print(T)]),
```
Vagabond, line 324: we should log this as a bug so we can find and fix any that are not EXITs or 2-tuple errors.

test/blockchain_simple_SUITE.erl
andymck: yeah good point, will add that

src/transactions/blockchain_txn.erl
andymck: roger

src/transactions/blockchain_txn.erl
```diff
@@ -310,7 +311,18 @@ separate_res([{T, Err} | Rest], Chain, V, I) ->
         {error, InvalidReason} = Error ->
             lager:warning("invalid txn ~p : ~p / ~s", [type(T), Error, print(T)]),
             %% any other error means we drop it
```
evanmcc, line 313: clean up the comments here?  we should just have a comment that says "since this is critical code, always make sure we have a catchall clause just in case" or something like that

evanmcc (approved):

src/transactions/blockchain_txn.erl
andymck: roger

Vagabond (approved):

PR #666, opened 2020-10-27
Limit sc client heap size to 128MB

vihu:


vihu:
Closing in favor of #667

PR #668, opened 2020-10-27
Use correct sc after nonce update

vihu:


Vagabond (approved):

evanmcc (approved):
this looks like an actual bug rather than just a test fix

PR #670, opened 2020-10-30
WIP: Add new targeting module

vihu:
This adds poc challenge rate limiting with logarithmic scaling on
network size. It's controlled using several newly introduced chain
variables.

This allows us to ensure consistent block times and gives us more
control on how many POCs we want to have per block as the network
continues to grow.

New variables:

- poc_beacons_per_day: Number of times a hotspot is allowed to beacon in a day
- poc_log_base: Step size to control logarithm scale base
- poc_rate_k: Extra control parameter as the network grows
- poc_consider_inactive: Hotspot considered inactive if (current_height - last_poc_challenge >= poc_consider_inactive)

There is also a corresponding eqc included here which tests how many
hotspots are getting targeted depending on the current ledger.

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -360,5 +360,15 @@
 %%
 -define(data_aggregation_version, data_aggregation_version).
 
+%% ------------------------------------------------------------------
 %% Multi-key
 -define(use_multi_keys, use_multi_keys).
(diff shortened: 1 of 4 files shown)


vihu:
Closing since we will do a new pass on targeting at some point

PR #677, opened 2020-11-12
Add support for hip17

vihu:
Refer: https://github.com/helium/HIP/blob/master/0017-hex-density-based-transmit-reward-scaling.md

TODO:
- [x] Switch h3 to master branch
- [x] Cross verify reward calculcations
- [x] Run performance checks
- [x] Add support for interactivity based on `last_poc_challenge`


src/blockchain_hex.erl
```diff
@@ -0,0 +1,217 @@
...
+-type var_map() :: #{non_neg_integer() => map()}.
+-type hex_resolutions() :: [non_neg_integer()].
+-type locations() :: [h3:h3_index()].
+
+%%--------------------------------------------------------------------
+%% Public functions
+%%--------------------------------------------------------------------
+
+-spec densities(Ledger :: blockchain_ledger_v1:ledger()) -> densities().
+densities(Ledger) ->
```
Vagabond, line 18: So the HIP mentions 'interactive hotspots', would it be better to build this map iteratively throughout the epoch as receipts come in (and age out inactive hotspots if they become stale?)

src/blockchain_hex.erl
```diff
@@ -0,0 +1,217 @@
...
+        12 => #{n => N12, density_tgt => Tgt12, density_max => Max12}
+    }.
+
+%%--------------------------------------------------------------------
+%% Internal functions
+%%--------------------------------------------------------------------
+
+-spec filtered_locations(Ledger :: blockchain_ledger_v1:ledger()) -> locations().
+filtered_locations(Ledger) ->
+    AG = blockchain_ledger_v1:active_gateways(Ledger),
```
Vagabond, line 81: This is something we should absolutely never use as the list keeps getting bigger and slower.

src/blockchain_hex.erl
```diff
@@ -0,0 +1,217 @@
+-module(blockchain_hex).
+
+-export([var_map/1, scale/4, densities/1]).
+
+-include("blockchain_vars.hrl").
+
+-type density_map() :: #{h3:h3_index() => pos_integer()}.
+-type densities() :: {density_map(), density_map()}.
```
Vagabond, line 8:
```suggestion
-type densities() :: {UnclippedDensities :: density_map(), ClippedDensities :: density_map()}.
```

Vagabond:
I think we should merge the h3dex branch into this and then rewrite this code to use that rather than the active_gateways thing.

src/blockchain_hex.erl
```diff
@@ -0,0 +1,217 @@
...
+-spec scale(
+    Location :: h3:h3_index(),
+    TargetRes :: non_neg_integer(),
+    UnclippedDensities :: density_map(),
+    ClippedDensities :: density_map()
+) -> float().
+scale(Location, TargetRes, UnclippedDensities, ClippedDensities) ->
+    lists:foldl(
+        fun(R, Acc) ->
+            Parent = h3:parent(Location, R),
```
Vagabond, line 36: This crashes if the hex is a lower resolution than R

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3090,6 +3109,123 @@ clean_all_hexes(Ledger) ->
...
+                               Ledger :: ledger()) -> Results :: h3dex().
+%% @doc Given a hex find candidate gateways in the span to the next adjacent
+%% hex. N.B. May return an empty map.
+lookup_gateways_from_hex(Hex, Ledger) ->
+    H3CF = h3dex_cf(Ledger),
+    cache_fold(Ledger, H3CF,
+               fun({Key, GWs}, Acc) ->
+                       maps:put(key_to_h3(Key), binary_to_term(GWs), Acc)
+               end, #{}, [
+                          {start, find_lower_bound_hex(Hex)},
```
Vagabond, line 3159: Is this missing a {seek tuple?

vihu: Updated in c29b687

src/blockchain_hex.erl
```diff
@@ -243,8 +243,11 @@ filter_interactive_gws(GWs, InteractiveBlocks, Ledger) ->
     lists:filter(fun(GWAddr) ->
                          case blockchain_ledger_v1:find_gateway_info(GWAddr, Ledger) of
                              {ok, GWInfo} ->
-                                 LastChallenge = blockchain_ledger_gateway_v2:last_poc_challenge(GWInfo),
-                                 (CurrentHeight - LastChallenge) =< InteractiveBlocks;
+                                 case blockchain_ledger_gateway_v2:last_poc_challenge(GWInfo) of
+                                     undefined -> false;
```
mrallen1, line 247: ahh yeah that's a good catch, thanks

src/blockchain.erl
```diff
@@ -94,7 +94,8 @@
 
 -define(BC_UPGRADE_FUNS, [fun upgrade_gateways_v2/1,
```
evanmcc, line 95: still need to add the mapping here to the `BC_UPGRADE_NAMES` def in `include/blockchain.hrl`

src/blockchain_hex.erl
```diff
@@ -0,0 +1,374 @@
...
+scale(Location, VarMap, TargetRes, Ledger) ->
+    case lookup(Location, ?SCALE_MEMO_TBL) of
+        {ok, Scale} -> Scale;
+        not_found ->
+            memoize(?SCALE_MEMO_TBL, Location,
+                    calculate_scale(Location, VarMap, TargetRes, Ledger))
+    end.
+
+
+%% TODO: This ought to be stored in the ledger because it won't change much? ever?
```
evanmcc, line 68: technically it is stored in the ledger as a series of chain vars.  I think that we allow binaries so you could just change it to a compressed serialzied map or something?

looks to me like this only gets called once per block on rewards, so probably not super worth optimizing

src/blockchain_hex.erl
```diff
@@ -0,0 +1,374 @@
...
+    case maps:size(Locations) of
+        0 ->
+            {#{}, #{}};
+        _ ->
+            UpperBoundRes = lists:max([h3:get_resolution(H3) || H3 <- maps:keys(Locations)]),
+            LowerBoundRes = h3:get_resolution(H3Root),
+
+            [Head | Tail] = lists:seq(UpperBoundRes, LowerBoundRes, -1),
+
+            %% find parent hexs to all hotspots at highest resolution in chain variables
```
evanmcc, line 226: hexs asdkjhasdlkgna';dflgnasd'fkl;anmdsf;'alkjngva'dlfb

src/blockchain_hex.erl
```diff
@@ -0,0 +1,374 @@
...
+                          maps:map(
+                            fun(_K, V) ->
+                                    filter_interactive_gws(V, InteractiveBlocks, Ledger)
+                            end, Locations)
+                  end,
+
+    %% Calculate clipped and unclipped densities
+    densities(H3Index, VarMap, Interactive, Ledger).
+
+-spec densities(
```
evanmcc, line 210: how hard would it be to make it so there's a precalc vs. on demand runtime switch so we could bench them against each other?

src/blockchain_hex.erl
mrallen1: ok

src/blockchain.erl
mrallen1: done

src/blockchain_hex.erl
mrallen1: I think @vihu's original code _did_ build a totally pre-calculated map but we kind of ripped it out when @Vagabond pointed out that most of it wouldn't be worthwhile and we'd have to keep rebuilding/refreshing it

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -250,11 +254,43 @@ get_rewards_for_epoch(Current, End, Chain, Vars, Ledger, ChallengerRewards, Chal
...
                 false ->
-                    get_rewards_for_epoch(Current+1, End, Chain, Vars, Ledger,
-                                          poc_challengers_rewards(Transactions, Vars, ChallengerRewards),
-                                          poc_challengees_rewards(Transactions, Vars, Chain, Ledger, ChallengeeRewards),
-                                          poc_witnesses_rewards(Transactions, Vars, Chain, Ledger, WitnessRewards),
-                                          dc_rewards(Transactions, End, Vars, Ledger, DCRewards))
+                    case blockchain_hex:var_map(Ledger) of
+                        {error, _} ->
+                            %% do the old thing
+                            get_rewards_for_epoch(Current+1, End, Chain, Vars, Ledger,
```
evanmcc, line 260: why not just one of these with 
```erlang
VarMap = case blockchain_hex:var_map(Ledger) of {ok, VM} -> VM; _ -> #{} end,
```
above it?  Are we actually doing anything different?

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -559,7 +622,14 @@ poc_challengees_rewards_(#{poc_version := Version}=Vars,
...
                                            maps:put(Challengee, I+3, Acc0);
                                        {ok, ToAdd} ->
-                                           maps:put(Challengee, I+ToAdd, Acc0)
+                                           TxScale = maybe_calc_tx_scale(Challengee,
+                                                                         DensityTgtRes,
+                                                                         ChallengeeLoc,
+                                                                         VarMap,
+                                                                         Ledger),
+                                           X = maps:put(Challengee, I+(ToAdd * TxScale), Acc0),
+                                           lager:info("MARKER1 +++++++~p", [print_intermediate_reward_map(X)]),
```
evanmcc, line 631: don't forget to do a debug logging sweep at the end.

src/transactions/v1/blockchain_txn_rewards_v1.erl
mrallen1: Planning to go through this today and see what we can do to make this suck less.

src/blockchain_hex.erl
Vagabond: My concern was the total memory needed for a global map. Even this local map building, I think, can throw away more than it does right now by discarding higher resolution hexes once we've calculated the lower resolution densities.

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -597,29 +677,39 @@ poc_challengees_rewards_(#{poc_version := Version}=Vars,
...
+                                           maps:put(Challengee, I+(ToAdd * TxScale), Acc0)
                                    end;
                                true ->
                                    case poc_challengee_reward_unit(WitnessRedundancy, DecayRate, Witnesses) of
                                        {error, _} ->
                                            %% Old behavior
                                            maps:put(Challengee, I+1, Acc0);
                                        {ok, ToAdd} ->
-                                           maps:put(Challengee, I+ToAdd, Acc0)
+                                           TxScale = maybe_calc_tx_scale(Challengee,
```
Vagabond, line 693: Does this need to be done here for versions < 4?

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -680,25 +777,56 @@ poc_witnesses_rewards(Transactions,
...
+                                                                                undefined ->
+                                                                                    Acc2;
+                                                                                ChallengeeLoc ->
+                                                                                    Witness = blockchain_poc_witness_v1:gateway(WitnessRecord),
+                                                                                    %% The witnesses get scaled by the value of their transmitters
+                                                                                    RxScale = blockchain_hex:scale(ChallengeeLoc,
+                                                                                                                   VarMap,
+                                                                                                                   D,
+                                                                                                                   Ledger),
+                                                                                    I = maps:get(Witness, Acc2, 0),
```
Vagabond, line 813: HIP 15 still is also done here, right?

src/transactions/v1/blockchain_txn_rewards_v1.erl
vihu: No, I will update this to reflect old behavior (prior to hip15 even). Also true for the successive branch.

vihu: Plz check 6e53fd9

src/transactions/v1/blockchain_txn_rewards_v1.erl
vihu: Updated 9032342

evanmcc (approved):
yay

PR #678, opened 2020-11-17
Codify implicit sorting of oracle submissions before txn close & fix …

Vagabond:
…issue

The implicit txn ordering behaviour if a transaction did not have a
defined sort order can cause instability in the sort order when a new
transaction is added, because the undefined sort order was the same as
the last sort order. This could lead to the transaction without a
defined sort order sorting before the transaction with a defined sort
order (because of the other sort dimensions like nonces, etc).This means
that adding another transaction to the end of the sort order list would
cause the transaction with the implicit sort order to definitely sort
after the transaction it may have sorted before earlier, invalidating
the block.

This change establishes a definitive sort order for the oracle price
report txn to mirror the implicit sort order before the addition of the
hotspot transfer function. Additionally it makes the undefined sort
order one past the maximum to prevent this bug recurring and it add a
log message to the sync_handler to make this failure clearer.

This *may* expose other implicit sort ordering bugs, so a thorough
resync (or at least ordering check of txns in blocks) is recommended.

### DIFF
--- a/src/transactions/blockchain_txn.erl
+++ b/src/transactions/blockchain_txn.erl
@@ -109,8 +109,9 @@
     {blockchain_txn_payment_v2, 19},
     {blockchain_txn_state_channel_open_v1, 20},
     {blockchain_txn_update_gateway_oui_v1, 21},
-    {blockchain_txn_state_channel_close_v1, 22},
-    {blockchain_txn_transfer_hotspot_v1, 23}
+    {blockchain_txn_price_oracle_v1, 22},
+    {blockchain_txn_state_channel_close_v1, 23},
+    {blockchain_txn_transfer_hotspot_v1, 24}
 ]).
 
 block_delay() ->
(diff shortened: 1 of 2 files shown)


mrallen1 (approved):
Great catch, thank you

src/transactions/blockchain_txn.erl
```diff
@@ -636,7 +637,7 @@ type_order(Txn) ->
     Type = type(Txn),
     case lists:keyfind(Type, 1, ?ORDER) of
         {Type, Index} -> Index;
-        false -> erlang:length(?ORDER)
+        false -> erlang:length(?ORDER) + 1
```
evanmcc, line 640: wouldn't it be safer to just error here?

evanmcc (approved):

vihu (approved):

src/transactions/blockchain_txn.erl
Vagabond: We'd have to audit that all other transactions have a defined order and make sure we establish that ordering, but yes.

vihu:
For reference, I was able to successfuly resync with revalidation from block `585361` and pass the supposed doom block `587899`

PR #682, opened 2020-11-25
Add sort order for routing_v1, token_burn_v1

madninja:
Also added txn_buncld and gen_price_oracle (both unused) to have the
complete list there.

routing_v1: blocks 552480, 550829, and  490139


Vagabond:
I would suggest we also remove the default +1 sorting order for unlisted txns as @evanmcc suggested

evanmcc (approved):

PR #697, opened 2020-12-14
Fix h3dex corruption on re-asserting location

vihu:
- Refactor cases on assert location
- Allow siblings (N) to go to 0
- Delete existing h3dex and redo bootstrap_h3dex


src/blockchain.erl
```diff
@@ -95,6 +95,7 @@
 -define(BC_UPGRADE_FUNS, [fun upgrade_gateways_v2/1,
                           fun bootstrap_hexes/1,
                           fun upgrade_gateways_oui/1,
+                          fun bootstrap_h3dex/1,
```
evanmcc, line 98: can we add a comment here noting that the duplication is not an error?

src/blockchain_hex.erl
```diff
@@ -312,7 +312,7 @@ limit(Res, VarMap, OccupiedCount) ->
     DensityMax = maps:get(max, VarAtRes),
     DensityTgt = maps:get(tgt, VarAtRes),
     N = maps:get(n, VarAtRes),
-    Max = max((OccupiedCount - N), 1),
+    Max = max((OccupiedCount - N + 1), 1),
```
evanmcc, line 315: moar parens

evanmcc (approved):
other than the nits I mentioned, looks good so far.

src/blockchain.erl
vihu: added 10d662b

src/blockchain_hex.erl
vihu: added 10d662b

PR #706, opened 2020-12-17
Expose tagged_witnesses to use with ETL

vihu:
Didn't really want to mess around with `valid_witnesses` because it requires changing critical logic in _several_ other places. Instead wrote a separate function `tagged_witnesses` which does NOT filter witnesses instead just adds a boolean and a reason tag to each witness and passed it down to the `to_json` function for both poc receipt txn and poc path element.

Unfortunately this introduces some code duplication but I think we can live with it (assuming that this indeed works)

### DIFF
--- a/src/transactions/v1/blockchain_poc_witness_v1.erl
+++ b/src/transactions/v1/blockchain_poc_witness_v1.erl
@@ -152,7 +152,7 @@ print(#blockchain_poc_witness_v1_pb{
 
 -spec to_json(poc_witness(), blockchain_json:opts()) -> blockchain_json:json_object().
 to_json(Witness, Opts) ->
-    Base = #{
+    Base0 = #{
              gateway => ?BIN_TO_B58(gateway(Witness)),
              timestamp => timestamp(Witness),
              signal => signal(Witness),
@@ -162,9 +162,13 @@ to_json(Witness, Opts) ->
              channel => ?MAYBE_UNDEFINED(channel(Witness)),
              datarate => ?MAYBE_UNDEFINED(?MAYBE_LIST_TO_BINARY(datarate(Witness)))
             },
-    case lists:keyfind(is_valid, 1, Opts) of
+    Base = case lists:keyfind(is_valid, 1, Opts) of
+               false -> Base0;
+               {is_valid, Valid} -> Base0#{ is_valid => Valid }
+           end,
+    case lists:keyfind(invalid_reason, 1, Opts) of
         false -> Base;
-        {is_valid, Valid} -> Base#{ is_valid => Valid }
+        {invalid_reason, InvalidReason} -> Base#{ invalid_reason => InvalidReason }
     end.
 
 
--- a/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
+++ b/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
@@ -36,6 +36,7 @@
     poc_id/1,
     good_quality_witnesses/2,
     valid_witnesses/3,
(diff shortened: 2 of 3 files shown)


src/transactions/v1/blockchain_poc_witness_v1.erl
```diff
@@ -162,9 +162,13 @@ to_json(Witness, Opts) ->
...
             },
-    case lists:keyfind(is_valid, 1, Opts) of
+    Base = case lists:keyfind(is_valid, 1, Opts) of
+               false -> Base0;
+               {is_valid, Valid} -> Base0#{ is_valid => Valid }
+           end,
+    case lists:keyfind(reason, 1, Opts) of
         false -> Base;
-        {is_valid, Valid} -> Base#{ is_valid => Valid }
+        {reason, Reason} -> Base#{ reason => Reason }
```
madninja, line 171: Ideally we'd have just a single field to encompass valid/invalid/reason but since we don't want to break json compatibility, can we rename reason to `invalid_reason` since the Reason value is only there when is_valid is false right?

madninja: And just to verify, Reason is a _string_ right?

madninja (changes requested):

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1244,6 +1266,82 @@ valid_witnesses(Element, Channel, Ledger) ->
                          end
                  end, Witnesses).
 
+%% This is an ETL specific function, it does NOT filter witnesses. It adds a tag to each witness specifying a reason why
+%% a witness was considered invalid
+-spec tagged_witnesses(Element :: blockchain_poc_path_element_v1:poc_element(),
+                       Channel :: non_neg_integer(),
+                       Ledger :: blockchain_ledger_v1:ledger()) -> tagged_witnesses().
+tagged_witnesses(Element, Channel, Ledger) ->
```
madninja, line 1274: Thanks for this special ETL function! The only concern I'd have is that it's not actually used anywhere but in the to_json? 

i..e there are no tests to ensure that this function works correctly?

madninja (changes requested):
One change request and question, left in comments

src/transactions/v1/blockchain_poc_witness_v1.erl
vihu: `reason` would be a binary, excerpt from a transaction for reference: https://gist.github.com/vihu/4126bfee0c871eb9cae5da9700a59f68

I tagged the witnesses with an `ok` reason when they are valid, just to keep things symmetric, if it's too much data on the wire we can remove it.

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Well, there's an eunit in the module with a _very_ basic test (I didn't edit that here). Let me check how awful adding a test for this is.

vihu: @madninja Updated in e24090d to not send `<<"ok">>` on valid witnesses. An example response for a [poc receipt txn](https://explorer.helium.com/txns/0RJecbABN5JsoeI3Sfai7pTe7_5h9PfpRsM0GN0VxQA) is [here](https://gist.github.com/vihu/aa63f4475b29aac54639a9f6b65371df).

Also changed the key to `invalid_reason`.

madninja (approved):

PR #707, opened 2021-01-04
Dont insert self witnesses with poc-v11

vihu:
This extends witness filtering to not insert a challengee as a witness of itself. We've seen that happening on rare occasions. This change is guarded behind `poc_version` chain var which would have to be incremented to 11 for this to take effect.

Note that the witness gc would remove any self witness records from the ledger, this does not address that.

### DIFF
--- a/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
+++ b/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
@@ -737,7 +737,7 @@ absorb(Txn, Chain) ->
                     {error, not_found} ->
                         %% Older poc version, don't add witnesses
                         ok;
-                    {ok, POCVersion} when POCVersion >= 10 ->
+                    {ok, POCVersion} when POCVersion >= 11 ->
                         %% Add filtered witnesses and remove self-witnessing with poc-v10
                         ok = valid_path_elements_fold(fun(Element, {FilteredWitnesses, FilteredReceipt}, _) ->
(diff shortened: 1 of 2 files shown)


src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -737,6 +737,23 @@ absorb(Txn, Chain) ->
                     {error, not_found} ->
                         %% Older poc version, don't add witnesses
                         ok;
+                    {ok, POCVersion} when POCVersion >= 10 ->
```
evanmcc, line 740: isn't the poc version already 10?  won't this break compat?

evanmcc: also why isn't it the same var as in the miner?

vihu: It is indeed 10, I updated it to be `>= 11` and also changed the poc_version max limit to go to 20.

f302a3c

vihu: As for the miner side, I was thinking that it won't matter at the time of construction? Could be wrong about that though. So self insert would not be allowed when we bump poc_version to 11 (because of this) and miner won't allow self witnessing once we land miner#592 was my thinking..

vihu:
This has been handled on the miner side https://github.com/helium/miner/pull/592, closing.

PR #715, opened 2021-01-12
detect missing request block hash

madninja:
Right around the edge of introducing request block hashes it looks like there is at least one receipts transaction with an empty/undefined request block hash.

folding over tagged path element did not take this into account and would fail with unfound channels. This fix ignores unfound channels completely (i.e. turns a blind eye to the missing request block)


vihu:
Can we find the poc_receipt/poc_request txn pair where the request hash is missing? Probably easier to lookup on ETL and once we have the hash I'll do some digging on the chain and see what happened. Don't want to jump to conclusions without taking a look at the txns themselves here.

vihu (approved):
lgtm, nice catch. Hindsight, I feel we should have enforced to do the tagging only after v10 was active, oh well.

evanmcc (approved):

PR #679, opened 2020-11-18
add rough commit hooking mechanism

evanmcc:
this compiles, but probably doesn't work.  however, it's roughly what I want it to do.

PR #716, opened 2021-01-12
Andymck/grpc validator support

andymck:
Support branch for sibyl and grpc validation services

Required by sibyl and miner PRs including:

https://github.com/helium/miner/pull/595
https://github.com/helium/sibyl/pull/1

Incorporates @evanmcc commit hooks PR https://github.com/helium/blockchain-core/pull/679 along with a number of other changes

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -136,7 +136,7 @@
     lookup_gateways_from_hex/2,
     add_gw_to_hex/3,
     remove_gw_from_hex/3,
-    add_commit_hook/3, add_commit_hook/4,
+    add_commit_hook/4, add_commit_hook/5,
     remove_commit_hook/2,
 
     %% snapshot save/restore stuff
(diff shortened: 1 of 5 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2378,6 +2402,18 @@ add_oui(Owner, Addresses, Filter, Subnet, Ledger) ->
             ok = cache_put(Ledger, SubnetCF, Subnet, <<OUI:32/little-unsigned-integer>>)
     end.
 
+-spec get_routes(ledger()) -> {ok, [blockchain_ledger_routing_v1:routing()]}
+                                                   | {error, any()}.
```
evanmcc, line 2406: this spec seems weird? it seems possible here to return `{ok, {error, any()}}`.  maybe case the return and check for errors, then wrap?

andymck: specs define it as returning {ok, [ListOfRouting]} or {error, any()} but upon second look I think I can drop the {error, any()} spec altogether....

evanmcc (approved):

Issue #717, opened 2021-01-14
Investigate why unclipped density had missing location key


vihu:
Error excerpt from blockchain-etl

> 3}],[]},{blockchain_hex,'-calculate_scale/4-fun-0-',5,[{file,"/opt/blockchain-etl/_build/default/lib/blockchain/src/blockchain_hex.erl"},{line,166}]},{lists,foldl,3,[{file,"lists.erl"},{line,1263}]},{blockchain_hex,scale,4,[{file,"/opt/blockchain-etl/_build/default/lib/blockchain/src/blockchain_hex.erl"},{line,65}]},{blockchain_hex,scale,2,[{file,"/opt/blockchain-etl/_build/default/lib/blockchain/src/blockchain_hex.erl"},{line,47}]},{be_db_gateway,maybe_reward_scale,2,[{file,"/opt/blockchain-etl/src/be_db_gateway.erl"},{line,285}]},{be_db_gateway,mk_gateway,2,[{file,"/opt/blockchain-etl/src/be_db_gateway.erl"},{line,277}]},{be_db_gateway,mk_gateway_hash,2,[{file,"/opt/blockchain-etl/src/be_db_gateway.erl"},{line,158}]}]},{gen_server,call,[<0.6058.3>,{with_transaction,#Fun<be_db_follower.1.29441482>},infinity]}}
     Offender:   [{pid,<0.1342.0>},{id,db_follower},{mfargs,{blockchain_follower,start_link,[[{follower_module,{be_db_follower,[{base_dir,"data"}]}}]]}},{restart_type,permanent},{shutdown,5000},{child_type,worker}]

PR #718, opened 2021-01-14
Handle failed scale calc for ETL

vihu:
Fixes https://github.com/helium/blockchain-core/issues/717

This isn't an ideal fix but we already do a try/catch in the rewards txn when the scale/4 calculation fails. This just does it so that ETL can also continue to ingest blocks.


madninja (approved):
Tested on dev-etl. Seems to work just fine

PR #719, opened 2021-01-18
Folding over active gateways cleanup

madninja:
While trying to speed up etl gateway change detection, I saw some repetition so squashed it with an active_gatways_fold function

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -789,33 +789,30 @@ scan_threshold_txns(Ledger, CF) ->
                     {iterate_upper_bound, <<"$threshold_txn`">>}]),
     lists:reverse(L).
 
--spec active_gateways(ledger()) -> active_gateways().
-active_gateways(Ledger) ->
+-spec active_gateways_fold(ledger(), 
+                            Fun :: fun((Key::binary(), Gateway::blockchain_ledger_gateway_v2:gateway(), Acc::any()) -> NewAcc::any()), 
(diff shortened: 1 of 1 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -789,33 +789,30 @@ scan_threshold_txns(Ledger, CF) ->
...
+                            Fun :: fun((Key::binary(), Gateway::blockchain_ledger_gateway_v2:gateway(), Acc::any()) -> NewAcc::any()), 
+                            OriginalAcc::any()) -> FinalAcc::any().
+active_gateways_fold(Ledger, Fun, Acc0) ->
     AGwsCF = active_gateways_cf(Ledger),
     cache_fold(
       Ledger,
       AGwsCF,
       fun({Address, Binary}, Acc) ->
               Gw = blockchain_ledger_gateway_v2:deserialize(Binary),
-              maps:put(Address, Gw, Acc)
```
evanmcc, line 800: delete this bit and use cf_fold in the other function?

madninja: Uh it is deleted? the change diff reads odd

evanmcc: https://github.com/helium/blockchain-core/blob/90518edd66fbdaa8483f2e9599f9973f8c135a76/src/ledger/v1/blockchain_ledger_v1.erl#L516

that function is effectively your `active_gateways_fold` function, just generalized, no?

PR #721, opened 2021-01-22
Extend PoC validation

vihu:
The goal here is to extend PoC validation by utilizing witness_count that are captured in the gateway ledger record.

#### TODO:
- [x] Introduce a new chain variable (`poc_witnessing_max`)
- [x] Extend poc validation to filter witnesses who have reached `poc_witnessing_max`
~~- [ ] Maybe allow poc_version to go higher than `10`, propose 50 and do the extra validation with `poc_version=11` + `max_witness_count`~~

### DIFF
--- a/src/ledger/v1/blockchain_ledger_gateway_v2.erl
+++ b/src/ledger/v1/blockchain_ledger_gateway_v2.erl
@@ -26,6 +26,7 @@
     has_witness/2,
     clear_witnesses/1,
     remove_witness/2,
+    lookup_witness_count/2,
     witnesses/1,
     witnesses_plain/1,
     witness_hist/1, witness_recent_time/1, witness_first_time/1,
@@ -483,6 +484,16 @@ has_witness(#gateway_v2{witnesses=Witnesses}, WitnessAddr) ->
         _ -> true
     end.
 
+-spec lookup_witness_count(gateway(), libp2p_crypto:pubkey_bin()) -> undefined | non_neg_integer().
+lookup_witness_count(#gateway_v2{witnesses=Witnesses}, WitnessAddr) ->
+    case lists:keyfind(WitnessAddr, 1, Witnesses) of
+        false ->
+            %% no witness found
+            undefined;
+        {_, Witness} ->
+            witness_count(Witness)
+    end.
+
 -spec witnesses(gateway()) -> #{libp2p_crypto:pubkey_bin() => gateway_witness()}.
 witnesses(Gateway) ->
     maps:from_list(Gateway#gateway_v2.witnesses).
@@ -494,6 +505,10 @@ witnesses_plain(Gateway) ->
 -spec witness_hist(gateway_witness()) -> erlang:error(no_histogram) | histogram().
 witness_hist(Witness) ->
     maps:from_list(Witness#witness.hist).
+
+-spec witness_count(gateway_witness()) -> undefined | non_neg_integer().
+witness_count(Witness) ->
+    Witness#witness.count.
 
 -spec witness_recent_time(gateway_witness()) -> undefined | non_neg_integer().
 witness_recent_time(Witness) ->
--- a/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
+++ b/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
@@ -681,10 +681,10 @@ tagged_path_elements_fold(Fun, Acc0, Txn, Ledger, Chain) ->
                                     _ ->
                                         {lists:nth(ElementPos - 1, Path), lists:nth(ElementPos - 1, Channels), lists:nth(ElementPos, Channels)}
                                 end,
-        
+
                                 FilteredReceipt = valid_receipt(PreviousElement, Element, ReceiptChannel, Ledger),
                                 TaggedWitnesses = tagged_witnesses(Element, WitnessChannel, Ledger),
-        
+
                                 Fun(Element, {TaggedWitnesses, FilteredReceipt}, Acc)
                         end, Acc0, lists:zip(lists:seq(1, length(Path)), Path));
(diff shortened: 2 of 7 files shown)


src/ledger/v1/blockchain_ledger_gateway_v2.erl
```diff
@@ -483,6 +484,16 @@ has_witness(#gateway_v2{witnesses=Witnesses}, WitnessAddr) ->
         _ -> true
     end.
 
+-spec lookup_witness_count(gateway(), libp2p_crypto:pubkey_bin()) -> undefined | non_neg_integer().
+lookup_witness_count(#gateway_v2{witnesses=Witnesses}, WitnessAddr) ->
+    case lists:keyfind(WitnessAddr, 1, Witnesses) of
+        false ->
+            %% no witness found
+            undefined;
```
evanmcc, line 492: should this just return 0?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1409,6 +1418,26 @@ get_channels(Txn, Chain) ->
...
     end.
 
+-spec check_witness_allowed_count(Gateway :: blockchain_ledger_gateway_v2:gateway(),
+                                  WitnessPubkeyBin :: libp2p_crypto:pubkey_bin(),
+                                  Ledger :: blockchain_ledger_v1:ledger()) -> boolean().
+check_witness_allowed_count(Gateway, WitnessPubkeyBin, Ledger) ->
+    case blockchain:config(?poc_witnessing_max, Ledger) of
+        {ok, AllowedCount} ->
+            %% var is set, compare this witness count with the gateway record
+            case blockchain_ledger_gateway_v2:lookup_witness_count(Gateway, WitnessPubkeyBin) of
```
evanmcc, line 1428: this would be simplified by returning 0 when the witness is missing

evanmcc (approved):

maco2035:
Anyone got info on this?

vihu:
> Anyone got info on this?

A blog post is being worked upon, but the gist here (also briefly mentioned in the description above) is that we've always had witness counts in the gateway record but never really used those for PoC validation. Ever since the activation of [HIP17](https://github.com/helium/HIP/blob/master/0017-hex-density-based-transmit-reward-scaling.md) which switched to beacon only PoCs, the necessity to add the changes here became obvious. There is no actual benefit to the network if some hotspot A witnesses B over and over again. This PR gives us the ability to tweak that behavior.

maco2035:
I understand it is useless for a hotspot to know one connection only. 100% agree what you said. But that means there is a cap on how much a hotspot can earn ever. This is sort of good and bad? But current network is based on PoC not Data transfer. But if it was to allow hotspots that haven't hit the limit get priority, and then others after that I don't see a problem with it.

cvolkernick:
Agree with @maco2035 as far as the network not being mature enough to be heavily dependent on data transfer only rewards, but such a drastic shift in earnings mechanisms makes me feel like I must have an incorrect understanding of what the changes imply.

vihu:
> But that means there is a cap on how much a hotspot can earn ever.

Recall that we have `witness_refresh_interval: 20000` which basically refreshes a hotspot's witness list in a deterministically random order, essentially making the counts null again. Implying that if for some hotspot A the witness cap is reached, they'll be reset and eligible again once their witness list gets refreshed.

Furthermore, this only applies to a particular pairing of hotspots, i.e. A witness B, X number of times, A would still be a valid witness for some other hotspot C (assuming that A hasn't witnessed C, X number of times). `X` being the value of `poc_witnessing_max`

cvolkernick:
>  There is no actual benefit to the network if some hotspot A witnesses B over and over again

Is it not correct to say it's valuable to demonstrate coverage continuity? That is, to continue to demonstrate over time that the hotspot in question continues to be where it is asserted to be. Hotspot A witnessing hotspot B repeatedly and consistently would suggest hotspot B continues to be where it says it is and has not moved elsewhere since previous beacon.

vihu:
> Is it not correct to say it's valuable to demonstrate coverage continuity? That is, to continue to demonstrate over time that the hotspot in question continues to be where it is asserted to be.

And this will still be true. Once the witness list is refreshed, it's the same behavior as currently on chain. The only introduction here is a witness_count cap (for A, B pair) which holds till the witness list is refreshed.

maco2035:
As I remember the witness_refresh_interval, that refreshes 20 000 is about a few weeks right? And it can be a day or a few weeks. I have seen lists not refresh fast too. But again, is it fair to a small group some farm to have less earnings then a big city? The job is the same, you want coverage, but cities are not the only place.

As far as my knowledge is too, that you can witness someone 50 times in like a week. So this would reduce earnings by 50% for some people who have small amount of witness. Because most of the time the hotspots when broadcasting can be heard by the same hotspots almost always. So people in big cities with more then the witness cap of witness will be able to do better then ones who are capped because they don't have any hotspots to witness in the area.

vihu:
> As I remember the witness_refresh_interval, that refreshes 20 000 is about a few weeks right? And it can be a day or a few weeks. I have seen lists not refresh fast too.

Yes, that's a valid concern and we're planning to reduce `witness_refresh_interval`, currently it's guaranteed that every hotspot on the network would have its witness list refreshed within ~14 days (20K blocks) and I agree that's a bit much. The blog will have proposed new value for `witness_refresh_interval`.

vihu:
> So people in big cities with more then the witness cap of witness will be able to do better then ones who are capped because they don't have any hotspots to witness in the area.

Remember that [HIP17](https://github.com/helium/HIP/blob/master/0017-hex-density-based-transmit-reward-scaling.md) introduced density based reward scaling, so hotspots in big cities with much higher hotspot density already get scale rewards.

PR #726, opened 2021-01-29
Allow routers to update xor filters

macpie:


test/blockchain_simple_SUITE.erl
```diff
@@ -1148,6 +1148,18 @@ routing_test(Config) ->
     Routing4 = blockchain_ledger_routing_v1:new(OUI2, Payer, Addresses0, Filter, <<0,0,4,127,255,254>>, 0),
     ?assertEqual({ok, Routing4}, blockchain_ledger_v1:find_routing(OUI2, Ledger)),
 
+    %% Test if router (in addresses can modify filters)
```
Vagabond, line 1151: Can you test that non-routers cannot modify filters in the OUI?

macpie: sure

Vagabond (approved):
I think this is ok to do and the code looks safe

PR #727, opened 2021-02-02
blockchain validators

evanmcc:
These are the blockchain changes needed for the validators features.  So far, this has elections, and adds validators to the ledger, and does some basic testing.

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -439,3 +439,20 @@
 -define(non_consensus_gateway_capabilities_mask, non_consensus_gateway_capabilities_mask).
 %% the mask value to represent the capabilities of full gateways, defined as an integer and used as a bitmask
 -define(full_gateway_capabilities_mask, full_gateway_capabilities_mask).
+
+%%%
+%%% validators variables
+%%%
+
+-define(validator_version, validator_version).
+-define(validator_minimum_stake, validator_minimum_stake).  % bones
+-define(validator_liveness_interval, validator_liveness_interval).  % blocks
+-define(validator_liveness_grace_period, validator_liveness_grace_period).  % blocks
+-define(validator_penalty_probability_factor, validator_penalty_probability_factor). % float
+-define(stake_withdrawal_cooldown, stake_withdrawal_cooldown). % blocks
+-define(stake_withdrawal_max, stake_withdrawal_max). % blocks
+%% -define(maximum_overstake, maximum_overstake). % float multiple of min stake
+
+-define(penalty_history_limit, penalty_history_limit). % blocks
+-define(dkg_penalty, dkg_penalty). % float
+-define(tenure_penalty, tenure_penalty). % float
--- a/rebar.config
+++ b/rebar.config
@@ -71,6 +71,9 @@
         {deps, [{meck, "0.8.12"}]}
     ]},
     {eqc, [
-        {deps, [{meck, "0.8.12"}]}
+           {erl_opts, [{d, 'TEST'}]},
+           {src_dirs, ["test", "src"]},
+           %% {cover_enabled, false},
+           {deps, [{meck, "0.8.12"}]}
     ]}
 ]}.
--- a/src/blockchain_election.erl
+++ b/src/blockchain_election.erl
@@ -3,7 +3,11 @@
 -export([
          new_group/4,
          has_new_group/1,
-         election_info/2
+         election_info/1, election_info/2,
+         icdf_select/3,
+         adjust_old_group/2,
+         adjust_old_group_v2/2,
+         validator_penalties/2
         ]).
 
 -include("blockchain_vars.hrl").
@@ -11,14 +15,17 @@
 
 -import(blockchain_utils, [normalize_float/1]).
 
--ifdef(TEST).
--export([
-         adjust_old_group/2
-        ]).
--endif.
+-record(val_v1,
+        {
+         prob :: float(),
+         heartbeat :: pos_integer() | undefined,
+         addr :: libp2p_crypto:pubkey_bin()
+        }).
 
 new_group(Ledger, Hash, Size, Delay) ->
     case blockchain_ledger_v1:config(?election_version, Ledger) of
+        {ok, N} when N >= 5 ->
+            new_group_v5(Ledger, Hash, Size, Delay);
         {ok, N} when N >= 4 ->
             new_group_v4(Ledger, Hash, Size, Delay);
         {ok, N} when N >= 3 ->
@@ -152,6 +159,93 @@ new_group_v4(Ledger, Hash, Size, Delay) ->
     Rem = OldGroup0 -- select(OldGroup, OldGroup, min(Remove, length(New)), RemovePct, []),
     Rem ++ New.
 
+-spec new_group_v5(Ledger :: blockchain_ledger_v1:ledger(),
+                   Hash :: binary(),
+                   Size :: non_neg_integer(),
+                   Delay :: non_neg_integer()) ->
+          [libp2p_crypto:pubkey_bin()].
+new_group_v5(Ledger, Hash, Size, Delay) ->
+    %% some complications here in the transfer.
+
+    %% we don't want to clutter this code with a bunch of gateway crap.  so we have a special,
+    %% simple code path just for the transitional time, and then a clean path for afterwards
+
+    %% deterministically set the random seed
+    blockchain_utils:rand_from_hash(Hash),
+
+    {ok, OldGroup0} = blockchain_ledger_v1:consensus_members(Ledger),
+
+    OldLen = length(OldGroup0),
+    {Remove, Replace} = determine_sizes_v2(Size, OldLen, Delay, Ledger),
+
+    Validators0 = validators_filter(Ledger),
+
+    %% remove dupes, sort
+    {OldGroupDeduped, Validators} = val_dedup(OldGroup0, Validators0, Ledger),
+
+    %% random shuffle of all validators, and change format into something idcf can consume
+    Validators1 = [{Addr, Prob}
+                  || #val_v1{addr = Addr, prob = Prob} <- blockchain_utils:shuffle(Validators)],
+
+    lager:debug("validators ~p ~p", [min(Replace, length(Validators1)), an2(Validators1)]),
+
+    %% select replacement nodes from the candidate pool using iterative icdf
+    New = icdf_select(Validators1, min(Replace, length(Validators1)), []),
+    lager:debug("validators new ~p", [an(New)]),
+
+    %% limit the number removed to the number of replacement nodes in the case that we're changing
+    %% size this election and it would end us up with an undersized group
+    NewLen = min(Remove, length(New)),
+    ToRem =
+        case have_gateways(OldGroup0, Ledger) of
+            [] ->
+                %% adjust for bbas and seen votes
+                OldGroupAdjusted = adjust_old_group_v2(OldGroupDeduped, Ledger),
+                lager:debug("old group ~p", [an2(OldGroupAdjusted)]),
+
+                case NewLen == 0 of
+                    %% moving down, we need remove bad nodes, not just take the first Size
+                    true when OldLen > Size ->
+                        icdf_select(lists:keysort(1, OldGroupAdjusted), OldLen - Size, []);
+                    true ->
+                        [];
+                    false ->
+                        icdf_select(lists:keysort(1, OldGroupAdjusted), NewLen, [])
+                end;
+            Gateways ->
+                %% just sort and remove the first removal amount rather than selecting, leaving
+                %% validators in to make up the number on the last round if needed.
+                lists:sublist(lists:sort(Gateways), 1, NewLen)
+        end,
+    lager:debug("to rem ~p", [an(ToRem)]),
+    %% shuffle the order to diffuse any ordering effects over time
+    blockchain_utils:shuffle((OldGroup0 -- ToRem) ++ New).
+
+an(M) ->
+    lists:map(fun blockchain_utils:addr2name/1, M).
+
+an2(M) ->
+    lists:map(fun({Addr, P}) -> {blockchain_utils:addr2name(Addr), P} end, M).
+
+icdf_select(_List, 0, Acc) ->
+    Acc;
+icdf_select(List, ToSelect, _Acc) when ToSelect > length(List) ->
+    {error, not_enough_elements};
+icdf_select(List, ToSelect, Acc) ->
+    {ok, Elt} = blockchain_utils:icdf_select(List, rand:uniform()),
+    icdf_select(lists:keydelete(Elt, 1, List), ToSelect - 1, [Elt | Acc]).
+
+have_gateways(List, Ledger) ->
+    lists:filter(
+      fun(X) ->
+              case blockchain_ledger_v1:get_validator(X, Ledger) of
+                  {ok, _} ->
+                      false;
+                  {error, not_found} ->
+                      true
+              end
+      end, List).
+
 %% all blocks other than the first block in an election epoch contain
 %% information about the nodes that consensus members saw during the
 %% course of the block, and also information about which BBAs were
@@ -160,15 +254,10 @@ new_group_v4(Ledger, Hash, Size, Delay) ->
 %% to the nodes that have not been seen or that are failing to drive
 %% their BBA executions to completion.
 adjust_old_group(Group, Ledger) ->
-    {ok, BBAPenalty} = blockchain_ledger_v1:config(?election_bba_penalty, Ledger),
-    {ok, SeenPenalty} = blockchain_ledger_v1:config(?election_seen_penalty, Ledger),
-
-    Sz = length(Group),
     %% ideally would not have to do this but don't want to change the
     %% interface.
-    Chain = blockchain_worker:blockchain(),
     #{start_height := Start,
-      curr_height := End} = election_info(Ledger, Chain),
+      curr_height := End} = election_info(Ledger),
     %% annotate the ledger group (which is ordered properly), with each one's index.
     {ok, OldGroup} = blockchain_ledger_v1:consensus_members(Ledger),
     {_, Addrs} =
@@ -179,8 +268,78 @@ adjust_old_group(Group, Ledger) ->
           {1, #{}},
           OldGroup),
 
-    Blocks = [begin {ok, Block} = blockchain:get_block(Ht, Chain), Block end
+    Blocks = [begin {ok, Block} = blockchain_ledger_v1:get_block(Ht, Ledger), Block end
               || Ht <- lists:seq(Start + 1, End)],
+
+    Penalties = get_penalties(Blocks, length(Group), Ledger),
+    lager:debug("penalties ~p", [Penalties]),
+
+    %% now that we've accumulated all of the penalties, apply them to
+    %% adjust the score for this group generation
+    lists:map(
+      fun({Score, Loc, Addr}) ->
+              Index = maps:get(Addr, Addrs),
+              Penalty = maps:get(Index, Penalties, 0.0),
+              lager:debug("~s ~p ~p", [blockchain_utils:addr2name(Addr), Score, Penalty]),
+              {normalize_float(Score - Penalty), Loc, Addr}
+      end,
+      Group).
+
+adjust_old_group_v2(Group, Ledger) ->
+    %% annotate the ledger group (which is ordered properly), with each one's index.
+    {ok, OldGroup} = blockchain_ledger_v1:consensus_members(Ledger),
+    Penalties = validator_penalties(OldGroup, Ledger),
+    lager:debug("penalties ~p", [Penalties]),
+
+    %% now that we've accumulated all of the penalties, apply them to
+    %% adjust the score for this group generation
+    lists:map(
+      fun(#val_v1{prob = Prob, addr = Addr}) ->
+              Penalty = maps:get(Addr, Penalties, 0.0),
+              {Addr, normalize_float(Prob + Penalty)}
+      end,
+      Group).
+
+validator_penalties(Group, Ledger) ->
+    {_, Addrs} =
+        lists:foldl(
+          fun(Addr, {Index, Acc}) ->
+                  {Index + 1, Acc#{Index => Addr}}
+          end,
+          {1, #{}},
+          Group),
+
+    #{start_height := Start,
+      curr_height := End} = election_info(Ledger),
+    Blocks = [begin {ok, Block} = blockchain_ledger_v1:get_block(Ht, Ledger), Block end
+              || Ht <- lists:seq(Start + 2, End)],
+
+    IndexedPenalties = get_penalties_v2(Blocks, length(Group), Ledger),
+    maps:fold(fun(I, Pen, Acc) ->
+                      Addr = maps:get(I, Addrs),
+                      Acc#{Addr => Pen}
+              end, #{}, IndexedPenalties).
+
+calc_age_weighted_penalty(Amt, Limit, Height, Instances) ->
+    Tot =
+        lists:foldl(
+          fun(H, Acc) ->
+                  BlocksAgo = Height - H,
+                  case BlocksAgo >= Limit of
+                      true ->
+                          Acc;
+                      _ ->
+                          %% 1 - ago/limit = linear inverse weighting for recency
+                          Acc + (Amt * (1 - (BlocksAgo/Limit)))
+                  end
+          end,
+          0.0,
+          Instances),
+    blockchain_utils:normalize_float(Tot).
+
+get_penalties(Blocks, Sz, Ledger) ->
+    {ok, BBAPenalty} = blockchain_ledger_v1:config(?election_bba_penalty, Ledger),
+    {ok, SeenPenalty} = blockchain_ledger_v1:config(?election_seen_penalty, Ledger),
 
     BBAs = [begin
                 BBA0 = blockchain_block_v1:bba_completion(Block),
@@ -215,33 +374,89 @@ adjust_old_group(Group, Ledger) ->
 
     lager:debug("penalties0 ~p", [Penalties0]),
 
-    Penalties =
+    lists:foldl(
+      fun(Seen, Acc) ->
+              maps:fold(
+                fun(Idx, false, A) ->
+                        maps:update_with(Idx, fun(X) -> X + SeenPenalty end,
+                                         SeenPenalty, A);
+                   (_, _, A) ->
+                        A
+                end,
+                Acc, Seen)
+      end,
+      Penalties0,
+      Seens).
+
+get_penalties_v2(Blocks, Sz, Ledger) ->
+    {ok, BBAPenalty} = blockchain_ledger_v1:config(?election_bba_penalty, Ledger),
+    {ok, SeenPenalty} = blockchain_ledger_v1:config(?election_seen_penalty, Ledger),
+    {ok, PenaltyLimit} = blockchain_ledger_v1:config(?penalty_history_limit, Ledger),
+    {ok, Height} = blockchain_ledger_v1:current_height(Ledger),
+
+    BBAs = [begin
+                BBA0 = blockchain_block_v1:bba_completion(Block),
+                Ht = blockchain_block_v1:height(Block),
+                {Ht, blockchain_utils:bitvector_to_map(Sz, BBA0)}
+            end || Block <- Blocks],
+
+    BBAPenalties =
         lists:foldl(
-          fun(Seen, Acc) ->
+          fun({Ht, BBA}, Acc) ->
+                  %% we want to generate a list for each idx where we have a height entry for each place
+                  %% where the BBA map is false for that index
                   maps:fold(
                     fun(Idx, false, A) ->
-                            maps:update_with(Idx, fun(X) -> X + SeenPenalty end,
-                                             SeenPenalty, A);
-                       (_, _, A) ->
-                            A
-                    end,
-                    Acc, Seen)
+                            maps:update_with(Idx, fun(X) -> [Ht | X]  end, A);
+                       (_, _, A) -> A
+                    end, Acc, BBA)
           end,
-          Penalties0,
+          maps:from_list([{I, []} || I <- lists:seq(1, Sz)]),
+          BBAs),
+
+    lager:debug("bba pens ~p", [BBAPenalties]),
+
+    Penalties0 =
+        maps:map(
+          fun(_Idx, HtList) ->
+                  calc_age_weighted_penalty(BBAPenalty, PenaltyLimit, Height, HtList)
+          end,
+          BBAPenalties),
+
+    Seens = [begin
+                 Seen0 = blockchain_block_v1:seen_votes(Block),
+                 Ht = blockchain_block_v1:height(Block),
+                 %% votes here are lists of per-participant seen
+                 %% information. condense here summarizes them, and a
+                 %% node is only voted against if there are 2f+1
+                 %% votes against it.
+                 {Ht, condense_votes(Sz, Seen0)}
+             end || Block <- Blocks],
+
+    SeenPenalties =
+        lists:foldl(
+          fun({Ht, Seen}, Acc) ->
+                  %% we want to generate a list for each idx where we have a height entry for each place
+                  %% where the BBA map is false for that index
+                  maps:fold(
+                    fun(Idx, false, A) ->
+                            maps:update_with(Idx, fun(X) -> [Ht | X]  end, A);
+                       (_, _, A) -> A
+                    end, Acc, Seen)
+          end,
+          maps:from_list([{I, []} || I <- lists:seq(1, Sz)]),
           Seens),
-    %% now that we've accumulated all of the penalties, apply them to
-    %% adjust the score for this group generation
 
-    lager:debug("penalties ~p", [Penalties]),
+    lager:debug("ct ~p bbas ~p seens ~p", [length(Blocks), BBAs, Seens]),
+    lager:debug("penalties0 ~p", [Penalties0]),
 
-    lists:map(
-      fun({Score, Loc, Addr}) ->
-              Index = maps:get(Addr, Addrs),
-              Penalty = maps:get(Index, Penalties, 0.0),
-              lager:info("~s ~p ~p", [blockchain_utils:addr2name(Addr), Score, Penalty]),
-              {normalize_float(Score - Penalty), Loc, Addr}
+    maps:fold(
+      fun(Idx, HtList, Acc) ->
+              SeenPen = calc_age_weighted_penalty(SeenPenalty, PenaltyLimit, Height, HtList),
+              maps:update_with(Idx, fun(X) -> SeenPen + X end, Acc)
       end,
-      Group).
+      Penalties0,
+      SeenPenalties).
 
 condense_votes(Sz, Seen0) ->
     SeenMaps = [blockchain_utils:bitvector_to_map(Sz, S)
@@ -301,6 +516,38 @@ determine_sizes(Size, OldLen, Delay, Ledger) ->
         false when Size > OldLen ->
             Remove = 0,
             Replace = Size - OldLen;
+        %% shrinking
+        false ->
+            Remove = OldLen - Size,
+            Replace = 0
+    end,
+    {Remove, Replace}.
+
+determine_sizes_v2(Size, OldLen, Delay, Ledger) ->
+    {ok, ReplacementFactor} = blockchain_ledger_v1:config(?election_replacement_factor, Ledger),
+    %% increase this to make removal more gradual, decrease to make it less so
+    {ok, ReplacementSlope} = blockchain_ledger_v1:config(?election_replacement_slope, Ledger),
+    {ok, Interval} = blockchain:config(?election_restart_interval, Ledger),
+    case Size == OldLen of
+        true ->
+            %% lower the replacement ceiling because in big groups, bigger is much much harder
+            MinSize = (2 * ((OldLen - 1) div 3)), % smallest remainder we will allow
+            BaseRemove = floor(Size/ReplacementFactor), % initial remove size
+            Removable = OldLen - MinSize - BaseRemove,
+            %% use tanh to get a gradually increasing (but still clamped to 1) value for
+            %% scaling the size of removal as delay increases
+            %% vihu argues for the logistic function here, for better
+            %% control, but tanh is simple
+            AdditionalRemove = floor(Removable * math:tanh((Delay/Interval) / ReplacementSlope)),
+
+            Remove = Replace = BaseRemove + AdditionalRemove;
+        %% growing
+        false when Size > OldLen ->
+            BaseRemove = floor(Size/ReplacementFactor),
+            Replace0 = Size - OldLen,
+            Remove = max(0, BaseRemove - Replace0),
+            %% by removing a few, this allows us to grow even when there are dead nodes in the group
+            Replace = Replace0 + Remove;
         %% shrinking
         false ->
             Remove = OldLen - Size,
@@ -504,10 +751,13 @@ has_new_group(Txns) ->
             false
     end.
 
-election_info(Ledger, Chain) ->
+election_info(Ledger) ->
     %% grab the current height and get the block.
     {ok, Height} = blockchain_ledger_v1:current_height(Ledger),
-    {ok, Block} = blockchain:get_block(Height, Chain),
+    election_info(Height, Ledger).
+
+election_info(Height, Ledger) when is_integer(Height) ->
+    {ok, Block} = blockchain_ledger_v1:get_block(Height, Ledger),
 
     %% get the election info
     {Epoch, StartHeight0} = blockchain_block_v1:election_info(Block),
@@ -517,7 +767,7 @@ election_info(Ledger, Chain) ->
     StartHeight = max(1, StartHeight0),
 
     %% get the election txn
-    {ok, StartBlock} = blockchain:get_block(StartHeight, Chain),
+    {ok, StartBlock} = blockchain_ledger_v1:get_block(StartHeight, Ledger),
     {ok, Txn} = get_election_txn(StartBlock),
     lager:debug("txn ~s", [blockchain_txn:print(Txn)]),
     ElectionHeight = blockchain_txn_consensus_group_v1:height(Txn),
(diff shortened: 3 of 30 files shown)


evanmcc:
requires https://github.com/helium/proto/pull/59

anthonyra:
After a quick review, I have a couple questions. Specifically on the cooldown portion of a validator due to the comment in the `is_valid` check.

To initiate cooldown it starts with `deactivate_validator` than it goes to `delay_stake` right? Shouldn't `delay_stake` instead be something like `start_cooldown` ... however this brings me to another question about the logic behind `delay_stake`. It seems that the current `delay_stake` does an atomic action where it "pre-fills" the stake to a future (TargetBlock) in the Ledger. Do I understand the mechanics behind that function even?

Now until I fully understand what's happening the following suggestions might be a mute point...
Would it make sense to have a `cooldown_hold` within the validator itself? Therefore, when a owner unstakes or even deactivates the validator the `stake` would go to 0 and the `cooldown_hold` would go to the previous staked value. This hold would maintain the HNT balance until the TargetBlock is reached at which point the atomic swap occurs to return the HNT to the owners account. This would also allow the cooldown to be reversible (in the sense that it could be restaked to that same validator) within the cooldown period. It'll also allow you to check if a validator exists and simply if the stake value is enough.

rebar.config
```diff
@@ -39,7 +39,7 @@
     {erlang_stats, ".*", {git, "https://github.com/helium/erlang-stats.git", {branch, "master"}}},
     {e2qc, ".*", {git, "https://github.com/project-fifo/e2qc", {branch, "master"}}},
     {vincenty, ".*", {git, "https://github.com/helium/vincenty", {branch, "master"}}},
-    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "master"}}},
+    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "pevm/validators"}}},
```
vihu, line 42: TODO: change to master when proto side lands

src/blockchain_election.erl
```diff
@@ -151,6 +160,73 @@ new_group_v4(Ledger, Hash, Size, Delay) ->
...
+                icdf_select(lists:keysort(1, OldGroupAdjusted), NewLen, []);
+            Gateways ->
+                %% just sort and remove the first removal amount rather than selecting, leaving
+                %% validators in to make up the number on the last round if needed.
+                lists:sublist(lists:sort(Gateways), 1, NewLen)
+        end,
+
+    (OldGroup0 -- ToRem) ++ New.
+
+icdf_select(_List, 0, Acc) ->
```
vihu, line 210: Maybe move this piece to `blockchain_utils`? Your call..

src/blockchain_election.erl
```diff
@@ -531,3 +613,54 @@ get_election_txn(Block) ->
...
+
+    maps:fold(
+      fun(Addr, Val = #val_v1{heartbeat = Last}, {Old, Candidates} = Acc) ->
+              Missing = (Height - Last) > (HBInterval + HBGrace),
+              case lists:member(Addr, OldGroup0) of
+                  true ->
+                      OldGw =
+                          case Missing of
+                              true ->
+                                  Val#val_v1{prob = 0.001};
```
vihu, line 650: We should move `0.001` out somewhere instead of hardcoding, also some comments about the workings of this whole function would be nice :)

src/blockchain_utils.erl
```diff
@@ -314,6 +314,7 @@ get_pubkeybin_sigfun(Swarm) ->
 -spec icdf_select([{any(), float()}, ...], float()) -> {ok, any()} | {error, zero_weight}.
 icdf_select(PopulationList, Rnd) ->
     Sum = lists:sum([Weight || {_Node, Weight} <- PopulationList]),
+    lager:info("sum ~p norm ~p", [Sum, normalize_float(Rnd * Sum)]),
```
vihu, line 317: TODO: Remove log line whence done debugging

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1201,6 +1263,45 @@ add_gateway_location(GatewayAddress, Location, Nonce, Ledger) ->
             end
     end.
 
+cg_versions(Ledger) ->
+    case config(?election_version, Ledger) of
+        N when N >= 5 ->
```
vihu, line 1268:
```suggestion
        {ok, N} when N >= 5 ->
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1201,6 +1263,45 @@ add_gateway_location(GatewayAddress, Location, Nonce, Ledger) ->
...
+                              false ->
+                                  maps:update_with(Version, Inc, 1, Acc)
+                          end
+                  end,
+                  #{}),
+            L = maps:to_list(Versions),
+            Tot = lists:sum([Ct || {_V, Ct} <- L]),
+
+            %% reformat counts as percentages
+            [{V, Ct / Tot} || {V, Ct} <- L]
```
vihu, line 1302: Potential divide by 0 :boom: here

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3348,6 +3454,131 @@ bootstrap_gw_denorm(Ledger) ->
       end,
       ignore).
 
+-spec add_validator(ValidatorAddress :: libp2p_crypto:pubkey_bin(),
+                    OwnerAddress :: libp2p_crypto:pubkey_bin(),
+                    Stake :: pos_integer(),
+                    Description :: string(),
+                    Ledger :: ledger()) -> ok | {error, gateway_already_active}.
```
vihu, line 3461:
```suggestion
                    Ledger :: ledger()) -> ok | {error, validator_already_added}.
```

vihu: or just `{error, any()}` I s'pose

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3348,6 +3454,131 @@ bootstrap_gw_denorm(Ledger) ->
...
+        {ok, _} ->
+            {error, validator_already_added};
+        _ ->
+            Val = blockchain_ledger_validator_v1:new(Address, OwnerAddr, Description, Stake),
+            update_validator(Address, Val, Ledger)
+    end.
+
+-spec update_validator(Addr :: libp2p_crypto:pubkey_bin(),
+                       Gw :: blockchain_ledger_validator_v1:validator(),
+                       Ledger :: ledger()) -> ok | {error, _}.
```
vihu, line 3477:
```suggestion
                       Ledger :: ledger()) -> ok | {error, any()}.
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3348,6 +3454,131 @@ bootstrap_gw_denorm(Ledger) ->
...
+
+            %% set status to unstaked
+            Val1 = blockchain_ledger_validator_v1:status(unstaked, Val),
+            %% increment the nonce
+            Val2 = blockchain_ledger_validator_v1:nonce(Nonce + 1, Val1),
+            %% 0 the stake
+            Val3 = blockchain_ledger_validator_v1:stake(0, Val2),
+            %% put the stake HNT into cooldown
+            ok = delay_stake(Owner, Address, Stake, Ledger),
+            update_validator(Address, Val3, Ledger);
```
vihu, line 3514: Question, this wouldn't necessarily "deactivate" (delete) said validator right? Do we want to remove it entirely?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3348,6 +3454,131 @@ bootstrap_gw_denorm(Ledger) ->
...
+        Error -> Error
+    end.
+            
+delay_stake(Owner, Validator, Stake, Ledger) ->
```
vihu, line 3518:
```suggestion
-spec delay_stake(Owner :: libp2p_crypto:pubkey_bin(), Validator: libp2p_crypto:pubkey_bin(), Stake :: non_neg_integer(), Ledger :: ledger()) -> ok | {error, any()}.
delay_stake(Owner, Validator, Stake, Ledger) ->
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3348,6 +3454,131 @@ bootstrap_gw_denorm(Ledger) ->
...
+                binary_to_term(HE);
+            not_found ->
+                []
+            %% just gonna function clause for now
+        end,
+    HeightEntry1 = HeightEntry ++ [{Owner, Validator, Stake}],
+    cache_put(Ledger, DefaultCF, cd_block_name(TargetBlock),
+              term_to_binary(HeightEntry1)).
+
+staked_hnt(Ledger) ->
```
vihu, line 3549:
```suggestion
-spec staked_hnt(Ledger :: ledger()) -> ok | {error, any()}.
staked_hnt(Ledger) ->
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3348,6 +3454,131 @@ bootstrap_gw_denorm(Ledger) ->
...
+      Ledger,
+      validators_cf(Ledger),
+      fun({_Addr, BinVal}, Acc) ->
+              Val = blockchain_ledger_validator_v1:deserialize(BinVal),
+              Acc + blockchain_ledger_validator_v1:stake(Val)
+      end,
+      0
+     ).
+
+cooldown_hnt(Ledger) ->
```
vihu, line 3560:
```suggestion
-spec cooldown_hnt(Ledger :: ledger()) -> ok | {error, any()}.
cooldown_hnt(Ledger) ->
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3464,6 +3695,23 @@ load_delayed_vars(DVars, Ledger) ->
       end, maps:from_list(DVars)),
     ok.
 
+snapshot_delayed_hnt(Ledger) ->
```
vihu, line 3698:
```suggestion
-spec snapshot_delayed_hnt(Ledger :: ledger()) -> ok | {error, any()}.
snapshot_delayed_hnt(Ledger) ->
```

src/ledger/v1/blockchain_ledger_validator_v1.erl
```diff
@@ -0,0 +1,138 @@
...
+-endif.
+
+-record(validator_v1,
+        {
+         address :: libp2p_crypto:pubkey_bin(),
+         owner_address :: libp2p_crypto:pubkey_bin(),
+         stake = 0 :: non_neg_integer(),
+         description = <<>> :: string(),
+         heartbeat = 1 :: pos_integer(),
+         nonce = 1 :: pos_integer(),
```
vihu, line 37: 1 or 0?

src/transactions/blockchain_txn.erl
```diff
@@ -37,7 +37,13 @@
...
              | blockchain_txn_gen_price_oracle_v1:txn_genesis_price_oracle()
              | blockchain_txn_transfer_hotspot_v1:txn_transfer_hotspot()
-             | blockchain_txn_state_channel_close_v1:txn_state_channel_close().
+             | blockchain_txn_state_channel_close_v1:txn_state_channel_close()
+             | blockchain_txn_gen_validator_v1:txn_gen_validator()
+             | blockchain_txn_stake_validator_v1:txn_stake_validator()
+             | blockchain_txn_change_validator_description_v1:txn_change_validator_description()
+             | blockchain_txn_transfer_validator_stake_v1:txn_transfer_validator_stake()
+             | blockchain_txn_unstake_validator_v1:txn_unstake_validator()
+             | blockchain_txn_unstake_validator_v1:txn_validator_heartbeat().
```
vihu, line 46: TODO: 

- I'm just adding this here, we need to handle all the validator transaction dependencies in `depends_on/2` function as well.
- Handle validator txn `actor/1` function
- Handle validator txn `nonce/1` function (if needed)

src/transactions/v1/blockchain_txn_change_validator_description_v1.erl
```diff
@@ -0,0 +1,212 @@
...
+    Validator = addr(Txn),
+    Description = description(Txn),
+    Fee = fee(Txn),
+
+    case blockchain_ledger_v1:debit_fee(Owner, Fee, Ledger, true) of
+        {error, _Reason} = Err -> Err;
+        ok ->
+            case blockchain_ledger_v1:get_validator(Validator, Ledger) of
+                {ok, V} ->
+                    %% make sure that the nonce is correct
```
vihu, line 169: TODO

src/transactions/v1/blockchain_txn_gen_validator_v1.erl
```diff
@@ -0,0 +1,184 @@
...
+    absorb/2,
+    print/1,
+    to_json/2
+]).
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-type txn_genesis_validator() :: #blockchain_txn_gen_validator_v1_pb{}.
```
vihu, line 34: nit, rename to `txn_gen_validator` maybe?

src/transactions/v1/blockchain_txn_stake_validator_v1.erl
```diff
@@ -0,0 +1,249 @@
...
+
+-spec is_valid(txn_stake_validator(), blockchain:blockchain()) ->
+          ok | {error, atom()} | {error, {atom(), any()}}.
+is_valid(Txn, Chain) ->
+    Ledger = blockchain:ledger(Chain),
+    Owner = ?MODULE:owner(Txn),
+    Validator = ?MODULE:validator(Txn),
+    Stake = stake(Txn),
+    Fee = fee(Txn),
+    case {is_valid_owner(Txn), is_valid_validator(Txn)} of
```
vihu, line 143: Missing an `is_valid_description` check, the description size should be =< `validator_description_max_len` chain var. Will apply to all txns where there is a description field.

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
```diff
@@ -0,0 +1,175 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain Transaction Stake Validator ==
```
vihu, line 3:
```suggestion
%% == Blockchain Transaction Validator Heartbeat ==
```

src/transactions/v1/blockchain_txn_transfer_validator_stake_v1.erl
```diff
@@ -0,0 +1,260 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain Transaction Stake Validator ==
```
vihu, line 3:
```suggestion
%% == Blockchain Transaction Transfer Validator Stake ==
```

src/transactions/v1/blockchain_txn_gen_validator_v1.erl
```diff
@@ -0,0 +1,184 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain Transaction Genesis Gateway ==
```
vihu, line 3:
```suggestion
%% == Blockchain Transaction Genesis Validator ==
```

src/transactions/v1/blockchain_txn_change_validator_description_v1.erl
```diff
@@ -0,0 +1,212 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain Transaction Stake Validator ==
```
vihu, line 3:
```suggestion
%% == Blockchain Transaction Change Validator Description ==
```

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -527,38 +527,62 @@ maybe_absorb(Txn, Ledger, _Chain) ->
...
+                      case blockchain_ledger_v1:get_validator(M, Ledger) of
+                          {ok, Val} ->
+                              V = blockchain_ledger_validator_v1:version(Val),
+                              lager:info("val ~p target ~p vers ~p", [Val, Target, V]),
+                              V >= Target;
+                          _Err -> lager:info("err ~p", [_Err]), false
+                      end
+              end,
+              Members);
+        _N ->
```
vihu, line 573:
```suggestion
        {ok, _N} ->
```

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1105,6 +1130,21 @@ validate_var(?density_tgt_res, Value) ->
...
 
+%% validators vars
+validate_var(?validator_minimum_stake, Value) ->
+    validate_int(Value, "validator_minimum_stake", 5000, 100000, false);
+validate_var(?validator_liveness_interval, Value) ->
+    validate_int(Value, "validator_liveness_interval", 5, 200, false);
+validate_var(?validator_liveness_grace_period, Value) ->
+    validate_int(Value, "validator_liveness_grace_period", 1, 100, false);
+validate_var(?validator_description_max_len, Value) ->
+    validate_int(Value, "validator_minimum_stake", 64, 1024, false);
```
vihu, line 1141:
```suggestion
    validate_int(Value, "validator_description_max_len", 64, 1024, false);
```

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: to allow cancellation, we need to leave it there, I think.  also nice for the historical record to keep people from reusing compromised keys

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: spec here is wrong, I'll correct it

src/ledger/v1/blockchain_ledger_validator_v1.erl
evanmcc: 1!  add doesn't set it to 1, I don't think.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3348,6 +3457,217 @@ bootstrap_gw_denorm(Ledger) ->
...
+                        not_found ->
+                            {error, missing_height_entry}
+                    end
+            end;
+        not_found ->
+            {error, not_found};
+        Err -> Err                
+    end.
+
+cooldown_stake(Val, Ledger) ->
```
Vagabond, line 3611: Is this a getter?

Vagabond, line 3611:
```suggestion
-spec cooldown_stake(Val :: blockchain_ledger_validator_v1:validator(), Ledger :: ledger()) -> {ok, non_neg_integer()}.
cooldown_stake(Val, Ledger) ->
```

evanmcc: I was thinking of these three (cooldown, circulating, staked) as queries

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -742,6 +767,32 @@ process_delayed_txns(Block, Ledger, Chain) ->
...
     cache_delete(Ledger, DefaultCF, block_name(Block)),
+
+    %% do the stake cooldown work
+    HeightEntries =
+        case cache_get(Ledger, DefaultCF, cd_block_name(Block), []) of
+            {ok, HE} ->
+                binary_to_term(HE);
+            not_found ->
+                []
+            %% just gonna function clause for now
```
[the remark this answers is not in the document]
evanmcc: copy paste error

src/blockchain_election.erl
evanmcc: probably

src/blockchain_election.erl
```diff
@@ -531,3 +613,54 @@ get_election_txn(Block) ->
...
+
+val_dedup(OldGroup0, Validators0, Ledger) ->
+    %% filter liveness here
+    {ok, HBInterval} = blockchain:config(?validator_liveness_interval, Ledger),
+    {ok, HBGrace} = blockchain:config(?validator_liveness_grace_period, Ledger),
+    {ok, Height} = blockchain_ledger_v1:current_height(Ledger),
+
+    maps:fold(
+      fun(Addr, Val = #val_v1{heartbeat = Last}, {Old, Candidates} = Acc) ->
+              Missing = (Height - Last) > (HBInterval + HBGrace),
```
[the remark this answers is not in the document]
evanmcc: will do

src/blockchain_election.erl
evanmcc: yeah, that's the intention.  I think I might have inverted the logic here.

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: this is old code.  it's not actually possible given the inputs.

src/transactions/v1/blockchain_txn_stake_validator_v1.erl
evanmcc: killed description

include/blockchain_vars.hrl
```diff
@@ -399,3 +399,14 @@
...
 -define(hip17_interactivity_blocks, hip17_interactivity_blocks).
+
+%%%
+%%% validators variables
+%%%
+
+-define(validator_minimum_stake, validator_minimum_stake).  % HNT
+-define(validator_liveness_interval, validator_liveness_interval).  % blocks
+-define(validator_liveness_grace_period, validator_liveness_grace_period).  % blocks
+-define(validator_description_max_len, validator_description_max_len).  % bytes
```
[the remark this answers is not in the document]
evanmcc: removed

src/blockchain_election.erl
```diff
@@ -151,6 +160,73 @@ new_group_v4(Ledger, Hash, Size, Delay) ->
...
+
+    lager:info("validators ~p", [Validators1]),
+    
+    %% replace select with iterative icdf
+    New = icdf_select(Validators1, min(Replace, length(Validators1)), []),
+    lager:info("validators new ~p", [New]),
+    
+    NewLen = min(Remove, length(New)),
+    ToRem = 
+        case have_gateways(OldGroup0, Ledger) of
```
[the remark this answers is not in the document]
evanmcc: yes, it is.  it's not great, but other than writing some special purpose switchover code once we're through the transition, I'm not sure how to not check each time.

src/blockchain_election.erl
evanmcc: yeah, thinking about it again, I made it large

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -527,38 +527,60 @@ maybe_absorb(Txn, Ledger, _Chain) ->
                     ok = blockchain_ledger_v1:delay_vars(Effective, Txn, Ledger),
                     true;
                 V ->
+                    lager:info("v = ~p", [V]),
```
Vagabond, line 530: Debugging

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
```diff
@@ -0,0 +1,177 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain Transaction Validator Heartbeat ==
+%% @end
+%%%-------------------------------------------------------------------
+-module(blockchain_txn_validator_heartbeat_v1).
```
Vagabond, line 6: I dislike the need for this transaction, hopefully we can factor it out as we build out the bits where light gateways will use the validators for services and use that signal for liveness instead.

Vagabond:
Is there no gate on validators being staked as soon as this rolls out to the current chain?

src/transactions/blockchain_txn.erl
```diff
@@ -645,7 +674,8 @@ type_order(Txn) ->
     Type = type(Txn),
     case lists:keyfind(Type, 1, ?ORDER) of
         {Type, Index} -> Index;
-        false -> erlang:length(?ORDER) + 1
+        %% don't implicitly order unknown transactions
```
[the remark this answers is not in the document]
evanmcc: there shouldn't be. we shouldn't be accepting anything that doesn't have an explicit ordering?

src/transactions/v1/blockchain_txn_gen_validator_v1.erl
```diff
@@ -0,0 +1,169 @@
...
+%%--------------------------------------------------------------------
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec absorb(txn_genesis_validator(), blockchain:blockchain()) -> ok | {error, atom()} | {error, {atom(), any()}}.
+absorb(Txn, Chain) ->
+    Ledger = blockchain:ledger(Chain),
+    Address = ?MODULE:address(Txn),
+    Owner = ?MODULE:owner(Txn),
+    Stake = ?MODULE:stake(Txn),
```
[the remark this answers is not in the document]
evanmcc: yes

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -141,9 +141,11 @@ absorb(Txn, Chain) ->
         #{},
         Rewards
     ),
+    lager:info("rewards ~p", [AccRewards]),
```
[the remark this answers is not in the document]
evanmcc: will remove

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
evanmcc: agree.  I'm not wild about it, but since there's no other verified activity right now, it's what we have.

src/transactions/v1/blockchain_txn_stake_validator_v1.erl
```diff
@@ -0,0 +1,284 @@
...
+                                %% make sure the staking amount is high enough
+                                case Stake == MinStake of
+                                    true -> ok;
+                                    false -> throw(stake_too_low)
+                                end,
+                                %% make sure that the owner has enough HNT to stake
+                                case blockchain_ledger_v1:find_entry(Owner, Ledger) of
+                                    {ok, Entry} ->
+                                        Balance = blockchain_ledger_entry_v1:balance(Entry),
+                                        case Balance >= Stake of
```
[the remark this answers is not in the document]
evanmcc: I mean, there's an argument that this shouldn't be here?  it's always possible to move the HNT aside before the transaction lands, so the checks in absorb have to be sufficient in case someone tries that.

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -184,6 +184,15 @@ calculate_rewards(Start, End, Chain) ->
...
                                                  Reward = blockchain_txn_reward_v1:new(Owner, Gateway, Amount, Type),
                                                  [Reward|Acc1]
+                                         end;
+                                    ({validator, Type, Validator}, Amount, Acc1) ->
+                                         case blockchain_ledger_v1:get_validator(Validator, Ledger) of
+                                             {error, _} ->
+                                                 Acc1;
+                                             {ok, Val} ->
+                                                 Owner = blockchain_ledger_validator_v1:owner_address(Val),
+                                                 Reward = blockchain_txn_reward_v1:new(Owner, Validator, Amount, Type),
```
andymck, line 194: Does v2 rewards need brought in here after its merged ?

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -411,6 +420,13 @@ calculate_epoch_reward(_Version, _Start, _End, BlockTime0, ElectionInterval, Mon
                                 map()) -> #{{gateway, libp2p_crypto:pubkey_bin()} => non_neg_integer()}.
 consensus_members_rewards(Ledger, #{epoch_reward := EpochReward,
                                     consensus_percent := ConsensusPercent}) ->
+    GwOrVal =
+        case blockchain:config(?election_version, Ledger) of
+            {ok, N} when N >= 5 ->
+                validator;
```
andymck, line 426: maybe worth adding a type for gateway or validator.  use it in specs, if we need to add another type in the future saves a lot of manual updates

andymck (commented):
a lotta code to get to grips with but on first pass looks good to me

src/transactions/v1/blockchain_txn_consensus_group_v1.erl
```diff
@@ -193,19 +205,50 @@ is_valid(Txn, Chain) ->
...
+                            true ->
+                                blockchain_election:validator_penalties(OldMembers, Chain, Ledger);
+                            false -> #{}
+                        end,
+
+                    lists:foreach(
+                      fun(M) ->
+                              {ok, V} = blockchain_ledger_v1:get_validator(M, Ledger),
+                              V1 = blockchain_ledger_validator_v1:add_penalty(V, CurrHeight, tenure,
+                                                                              1.0, PenaltyLimit),
```
Vagabond, line 240: Should be a chain var, not hardcoded to 1.0

src/blockchain_election.erl
```diff
@@ -531,3 +776,96 @@ get_election_txn(Block) ->
...
+                      %% this clause handles generating the list of new nodes to potentially add (Candidates)
+                      _ ->
+                          lager:debug("name ~p out off ~p", [blockchain_utils:addr2name(Addr), Offline]),
+                          case Offline of
+                              %% don't bother to add to the candidate list
+                              true ->
+                                  Acc;
+                              _ ->
+                                  %% use 5 instead of 1 to make things more lenient
+                                  case max(0.0, 5.0 - Prob) of
```
Vagabond, line 848: 5.0 should be a chain var?

src/transactions/v1/blockchain_txn_consensus_group_v1.erl
```diff
@@ -193,19 +205,50 @@ is_valid(Txn, Chain) ->
...
+                              V2 = case maps:get(M, EpochPenalties, none) of
+                                       none -> V1;
+                                       Penalty ->
+                                           blockchain_ledger_validator_v1:add_penalty(V1, CurrHeight, performance,
```
Vagabond, line 244: If this is 0, don't add it?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3799,16 +3920,230 @@ remove_commit_hook(Atom, #ledger_v1{commit_hooks = Hooks} = Ledger) when is_atom
...
+        end,
+    HeightEntry1 = HeightEntry ++ [{Owner, Validator, Stake}],
+    cache_put(Ledger, DefaultCF, cd_block_name(StakeReleaseHeight),
+              term_to_binary(HeightEntry1)).
+
+cancel_cooldown_stake(Val, Ledger) ->
```
evanmcc, line 4035: remove this as it's no longer possible

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3799,16 +3920,230 @@ remove_commit_hook(Atom, #ledger_v1{commit_hooks = Hooks} = Ledger) when is_atom
...
+            %% set status to unstaked
+            Val1 = blockchain_ledger_validator_v1:status(unstaked, Val),
+            %% zero out the stake
+            Val2 = blockchain_ledger_validator_v1:stake(0, Val1),
+            update_validator(Address, Val2, Ledger);
+        Error -> Error
+    end.
+
+
+-spec activate_validator(libp2p_crypto:pubkey_bin(), ledger()) ->
```
evanmcc, line 3989: no longer used, remove

src/blockchain_election.erl
```diff
@@ -531,3 +776,96 @@ get_election_txn(Block) ->
...
+      fun({Addr, BinVal}, Acc) ->
+              Val = blockchain_ledger_validator_v1:deserialize(BinVal),
+              Stake = blockchain_ledger_validator_v1:stake(Val),
+              Status = blockchain_ledger_validator_v1:status(Val),
+              Version = blockchain_ledger_validator_v1:version(Val),
+              Penalty = blockchain_ledger_validator_v1:calculate_penalty_value(Val, Ledger),
+              HB = blockchain_ledger_validator_v1:last_heartbeat(Val),
+              case Stake >= MinStake andalso Status == staked of
+                  true ->
+                      case LimitVersion == false orelse Version == AllowedVersion of
```
evanmcc, line 800: add a TODO here to mark this as test-only code

rebar.config
```diff
@@ -32,7 +32,7 @@
     {lager, "3.9.1"},
     {erl_base58, "0.0.1"},
     {base64url, "1.0.1"},
-    {libp2p, ".*", {git, "https://github.com/helium/erlang-libp2p.git", {branch, "master"}}},
+    {libp2p, ".*", {git, "https://github.com/helium/erlang-libp2p.git", {branch, "pevm/txn-get-read-opts"}}},
```
Vagabond, line 35: I think this landed?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -272,6 +289,15 @@
           aux_heights :: rocksdb:cf_handle()
          }).
 
+%% This record is for managing validator stake cooldown information in the
+%% ledger
+-record(validator_stake_v1, {
```
Vagabond, line 294: Was there a reason this was not done in a blockchain_ledger_validator_v1 module?

mrallen1: Arguably it should've been done in its own module but it wouldn't really make the intent of the code which uses these records any clearer tbh

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3303,22 +3425,33 @@ get_aux_rewards_(Itr, {ok, Key, BinRes}, Acc) ->
...
+    SL = subledger(Ledger),
+    SL#sub_ledger_v1.validators.
+
 -spec cache_put(ledger(), rocksdb:cf_handle(), binary(), binary()) -> ok.
 cache_put(Ledger, CF, Key, Value) ->
-    {Cache, _GwCache} = context_cache(Ledger),
-    true = ets:insert(Cache, {{CF, Key}, Value}),
-    ok.
+    case context_cache(Ledger) of
+        {undefined, _GwCache} -> ok; %% needed for test cases
```
Vagabond, line 3436: I'm concerned this will cause silent failures, why do the test cases need it?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3799,16 +3932,187 @@ remove_commit_hook(Atom, #ledger_v1{commit_hooks = Hooks} = Ledger) when is_atom
...
+
+-spec deactivate_validator(libp2p_crypto:pubkey_bin(), pos_integer(), ledger()) ->
+          ok | {error, any()}.
+deactivate_validator(Address, StakeReleaseHeight, Ledger) ->
+    case get_validator(Address, Ledger) of
+        {ok, Val} ->
+            Stake = blockchain_ledger_validator_v1:stake(Val),
+            Owner = blockchain_ledger_validator_v1:owner_address(Val),
+
+            %% set status to unstaked
```
Vagabond, line 3981:
```suggestion
            %% set status to cooldown
```

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -468,7 +484,7 @@ consensus_members_rewards(Ledger, #{epoch_reward := EpochReward,
                 fun(Member, Acc) ->
                     PercentofReward = 100/Total/100,
                     Amount = erlang:round(PercentofReward*ConsensusReward),
-                    maps:put({gateway, consensus, Member}, Amount, Acc)
+                    maps:put({GwOrVal, consensus, Member}, Amount, Acc)
```
mrallen1, line 487: If `validator` should be in this tuple, then rewards_v2 should get this change

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -229,6 +229,15 @@ calculate_rewards_(Start, End, Ledger, Chain) ->
                                              {ok, Owner} ->
                                                  Reward = blockchain_txn_reward_v1:new(Owner, Gateway, Amount, Type),
                                                  [Reward|Acc1]
+                                         end;
+                                    ({validator, Type, Validator}, Amount, Acc1) ->
```
mrallen1, line 233: Here's another change that should be in `rewards_v2`

src/transactions/v1/blockchain_txn_gen_validator_v1.erl
```diff
@@ -0,0 +1,169 @@
...
+
+%%--------------------------------------------------------------------
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec sign(txn_genesis_validator(), libp2p_crypto:sig_fun()) -> txn_genesis_validator().
+sign(Txn, _SigFun) ->
+    Txn.
+
+%%--------------------------------------------------------------------
```
mrallen1, line 66: remove

src/transactions/v1/blockchain_txn_gen_validator_v1.erl
```diff
@@ -0,0 +1,169 @@
...
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec hash(txn_genesis_validator()) -> blockchain_txn:hash().
+hash(Txn) ->
+    EncodedTxn = blockchain_txn_gen_validator_v1_pb:encode_msg(Txn),
+    crypto:hash(sha256, EncodedTxn).
+
+%%--------------------------------------------------------------------
```
mrallen1, line 58: anything here that's empty should remove

src/ledger/v1/blockchain_ledger_validator_v1.erl
```diff
@@ -0,0 +1,236 @@
...
+-record(validator_v1,
+        {
+         address :: libp2p_crypto:pubkey_bin(),
+         owner_address :: libp2p_crypto:pubkey_bin(),
+         stake = 0 :: non_neg_integer(),
+         heartbeat = 1 :: pos_integer(),
+         nonce = 1 :: pos_integer(),
+         version = 1 :: pos_integer(),
+         status = staked :: status(),
+         penalties = [] :: [penalty()]
```
mrallen1, line 55: maybe the stake release height should be in this record too?

mrallen1 (commented):
Few items to consider - also I think vihu's suggestions about the actor() and nonce() functions probably should be addressed before this gets merged

include/blockchain_vars.hrl
```diff
@@ -439,3 +439,21 @@
 -define(non_consensus_gateway_capabilities_mask, non_consensus_gateway_capabilities_mask).
 %% the mask value to represent the capabilities of full gateways, defined as an integer and used as a bitmask
 -define(full_gateway_capabilities_mask, full_gateway_capabilities_mask).
+
+%%%
+%%% validators variables
+%%%
+
+-define(validator_version, validator_version).
+-define(validator_minimum_stake, validator_minimum_stake).  % HNT
```
vihu, line 448: Should this be in bones?

src/blockchain_election.erl
```diff
@@ -3,22 +3,29 @@
...
 -import(blockchain_utils, [normalize_float/1]).
 
--ifdef(TEST).
--export([
-         adjust_old_group/2
-        ]).
--endif.
+-record(val_v1,
+        {
+         prob :: float(),
```
vihu, line 20: By default this is undefined? I haven't read further down yet but would it make sense to have some default value here?

src/blockchain_election.erl
```diff
@@ -152,6 +159,85 @@ new_group_v4(Ledger, Hash, Size, Delay) ->
...
+    %% remove dupes, sort
+    {OldGroupDeduped, Validators} = val_dedup(OldGroup0, Validators0, Ledger),
+
+    %% random shuffle of all validators, and change format into something idcf can consume
+    Validators1 = [{Addr, Prob}
+                  || #val_v1{addr = Addr, prob = Prob} <- blockchain_utils:shuffle(Validators)],
+
+    lager:debug("validators ~p ~p", [min(Replace, length(Validators1)), an2(Validators1)]),
+
+    %% replace select with iterative icdf
```
vihu, line 187: What about this?

And it would be really helpful if we could add a few words on what the intentions for the next couple lines below are perhaps. So we select X number of new validators where X is the minimum between the number of validators we wanted to replace and the number of new validators?

vihu: Good chance I read it incorrectly cuz I see NewLen as the minimum between the number to remove and the previous NewLength in the next lines heh

src/blockchain_election.erl
```diff
@@ -152,6 +159,85 @@ new_group_v4(Ledger, Hash, Size, Delay) ->
...
+    New = icdf_select(Validators1, min(Replace, length(Validators1)), []),
+    lager:debug("validators new ~p", [an(New)]),
+
+    NewLen = min(Remove, length(New)),
+    ToRem =
```
vihu, line 192: I think we should split this whole `new_group_v5` function further; I'd even suggest more verbose names like `filter_new_selected` or whatever

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1177,6 +1194,36 @@ validate_var(?light_gateway_capabilities_mask, Value) ->
...
+validate_var(?stake_withdrawl_cooldown, Value) ->
+    %% maybe set this in the test
+    validate_int(Value, "stake_withdrawal_cooldown", 5, 1000000, false);
+validate_var(?stake_withdrawal_max, Value) ->
+    validate_int(Value, "stake_withdrawal_max", 50, 1000, false);
+
+validate_var(?dkg_penalty, Value) ->
+    validate_float(Value, "dkg_penalty", 0.0, 5.0);
+validate_var(?penalty_history_limit, Value) ->
+    %% low end is low for testing and an out if these become corrupted
```
vihu, line 1221:
```suggestion
    %% low end is low for testing and an out if these become corrupted
    %% also low end cannot be 0
```

src/blockchain_election.erl
```diff
@@ -544,3 +789,96 @@ get_election_txn(Block) ->
...
+                  case lists:member(Addr, OldGroup0) of
+                      %% this clause keeps the old group out of the candidate list and additionally
+                      %% marks really bad nodes for removal
+                      true ->
+                          lager:debug("name ~p in off ~p", [blockchain_utils:addr2name(Addr), Offline]),
+                          OldGw =
+                              case Offline of
+                                  true ->
+                                      %% try and make sure offline nodes are selected
+                                      Val#val_v1{prob = 200.0};
```
vihu, line 847: `200.0` seems a bit arbitrary?

include/blockchain_vars.hrl
Vagabond: Yes, I don't think we have anything else denominated in HNT?

evanmcc: the comment is wrong and the stake is in bones

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: @mrallen1 added this, I think, maybe he can clarify?

src/transactions/v1/blockchain_txn_transfer_validator_stake_v1.erl
```diff
@@ -0,0 +1,353 @@
...
+                end,
+                %% make sure the amount is encoded correctly
+                %% and is only specified in the correct case
+                case payment_amount(Txn) of
+                    %% 0 is always ok
+                    0 -> ok;
+                    N when is_integer(N) andalso N >= 0 ->
+                        case new_owner(Txn) /= <<>> of
+                            true -> ok;
+                            false -> throw(amount_set_for_inter_account_transfer)
```
Vagabond, line 212:
```suggestion
                            false -> throw(amount_set_for_intra_account_transfer)
```

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -555,15 +556,30 @@ maybe_absorb(Txn, Ledger, _Chain) ->
     end.
 
 check_members(Members, Target, Ledger) ->
-    lists:all(fun(M) ->
+    case blockchain_ledger_v1:config(?election_version, Ledger) of
+        {ok, N} when N >= 5 ->
+            lists:all(
+              fun(M) ->
+                      case blockchain_ledger_v1:get_validator(M, Ledger) of
```
Vagabond, line 563: So this would allow a chain var to only clear once all the validators are reporting a specific version?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1189,6 +1206,36 @@ validate_var(?full_gateway_capabilities_mask, Value) ->
...
+        _ ->
+            throw({error, {invalid_validator_version, Value}})
+    end;
+validate_var(?validator_minimum_stake, Value) ->
+    validate_int(Value, "validator_minimum_stake", ?bones(5000), ?bones(100000), false);
+validate_var(?validator_liveness_interval, Value) ->
+    validate_int(Value, "validator_liveness_interval", 5, 200, false);
+validate_var(?validator_liveness_grace_period, Value) ->
+    validate_int(Value, "validator_liveness_grace_period", 1, 100, false);
+%% TODO fix this var
```
Vagabond, line 1223: TODO?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3989,16 +4118,187 @@ remove_commit_hook(Atom, #ledger_v1{commit_hooks = Hooks} = Ledger) when is_atom
...
+    %% create or add an stake entry for the owner
+    StakeList =
+        case cache_get(Ledger, DefaultCF, owner_name(Owner), []) of
+            {ok, OE} ->
+                binary_to_term(OE);
+            not_found ->
+                []
+            %% just gonna function clause for now
+        end,
+    NewStakeRec = new_stake_record(Owner, Validator, Stake, StakeReleaseHeight),
```
Vagabond, line 4205: Why do we store this in 2 places in the ledger? So it's easier to find?

src/ledger/v1/blockchain_ledger_v1.erl
mrallen1: we can remove it I think - at the time I was worried about a test failure in the unstake height PR which was triggered by this case being unhandled

src/transactions/v1/blockchain_txn_vars_v1.erl
evanmcc: it's the same mechanism as before, but IMO all this code should be removed.  We've never used it, we should get rid of it, since these things are almost impossible to use now.

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: we have to store it twice (although originally it was two slightly different bits of information that were stored) in order to support the two lookup patterns (1. when the block passes, 2. when we need to look up the stake by validator).  if we take @mrallen1's suggestion above to store the release block in the validator record, we won't need to store it twice.  I can PR that this afternoon.

src/blockchain_election.erl
```diff
@@ -152,6 +159,85 @@ new_group_v4(Ledger, Hash, Size, Delay) ->
...
+                lager:debug("old group ~p", [an2(OldGroupAdjusted)]),
+
+                case NewLen == 0 of
+                    %% moving down, we need remove bad nodes, not just take the first Size
+                    true when OldLen > Size ->
+                        icdf_select(lists:keysort(1, OldGroupAdjusted), OldLen - Size, []);
+                    true ->
+                        [];
+                    false ->
+                        icdf_select(lists:keysort(1, OldGroupAdjusted), NewLen, [])
```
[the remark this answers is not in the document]
evanmcc: I don't think that this is possible, no.

src/transactions/blockchain_txn.erl
Vagabond: I'm not sure we have any dependences worth caring about, and these txns are all nonceless.

Actor() we could do, but I'm not sure its worth it?

PR #729, opened 2021-02-06
Rough draft of HIP-24 implementation

ericmheilman:


DGKSK8LIFE (approved):

vihu:
x-posting from [discord](https://discord.com/channels/404106811252408320/795832567265624095/807763374980726815) for reference:

A few things:

- I'm pretty sure that you can do a similar testing strategy as done in blockchain_payment_v2_SUITE, it works largely independent of the chain and just checks the txn validation constraints
- Don't have two versions of rewards_map (one as a list, other as a map) unless there's an absolutely necessary reason for it
- What happens when there are two competing reward_split transactions? How do we ensure that only one split transaction for a particular account is valid at a given point in time? Should this be tied to the nonce of an account (I'm leaning towards yes)? Presumably your payments would get messed up if there's also debit happening from the same account using the reward_split txn. 
- When its time, we'll probably need to run an end-to-end test on the miner side (there's a payment_v2_SUITE in miner as well for reference)
- Ensure that you're able to sync the chain with this code running (without the vars activated of course)

vihu:
I forgot to mention this earlier, I don't see `blockchain_txn.erl` being updated here (yet). Basically `blockchain_txn` requires ensuring a particular txn's actor, type, wrapping, dependencies etc. (i.e. how do we expect each txn to behave). Would advise to essentially look for places where any particular txn (`payment_v2` for example) is handled in `blockchain_txn.erl` and follow along those breadcrumbs.

Issue #730, opened 2021-02-07
Support an "auxiliary ledger" to be able to quantify the result of consensus changes


Vagabond:
Optionally run a second ledger that has special rules for absorbing txns such that we can test different chain var values or behavior over some period of time. We could add some magic check like `case blockchain_ledger_v1:is_auxiliary` and run conditional code (or set a different chain var on the auxiliary ledger).

We'd absorb blocks/txns into both ledgers, but we'd only consult the main ledger. At the end of the run or the test we could compare the ledgers against each other to see what differs. This would allow for easier validation of new or proposed changes before actually making the change live.

PR #731, opened 2021-02-08
[DNM] [WIP] Aux ledger

vihu:
Will track #730


Vagabond:
Why is aux just not a new mode on the existing ledger?

vihu:
Closing since the strategy here is not quite correct

PR #736, opened 2021-02-16
Add eqc for elector selection

vihu:


eqc/election_prob_eqc.erl
```diff
@@ -0,0 +1,66 @@
...
+                                                          end, Acc, Selected)
+                                      end,
+                                      #{},
+                                      lists:seq(1, Iterations)),
+
+                Expectations = lists:foldl(fun({Candidate, Prob}, Acc) ->
+                                                   maps:put(Candidate, round(Prob * Iterations), Acc)
+                                           end, #{}, Candidates),
+
+                CheckCounter = Counter == Expectations,
```
vihu, line 32: This is just so that it fails and we can look at the results, I will update it to have an error delta and do the compare with that.

evanmcc (approved):

PR #740, opened 2021-02-22
Allow for testnet keys

madninja:
This replaces all uses of libp2p_crypto:bin_to_pubkey and pubkey_to_bin with a blockchain_utils intermediary which uses the chain network as defined by a chain variable (and cached in a persistent_term for speed)

### DIFF
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -27,6 +27,10 @@
     find_txn/2,
     map_to_bitvector/1,
     bitvector_to_map/2,
+    bin_to_pubkey/1,
+    pubkey_to_bin/1,
+    get_network/1,
+    set_network/1,
     get_pubkeybin_sigfun/1,
     approx_blocks_in_week/1,
     keys_list_to_bin/1,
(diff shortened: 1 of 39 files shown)


src/blockchain_utils.erl
```diff
@@ -547,12 +572,12 @@ bitvector_roundtrip_test() ->
 oracle_keys_test() ->
     #{ public := RawEccPK } = libp2p_crypto:generate_keys(ecc_compact),
     #{ public := RawEdPK } = libp2p_crypto:generate_keys(ed25519),
-    EccPK = libp2p_crypto:pubkey_to_bin(RawEccPK),
-    EdPK = libp2p_crypto:pubkey_to_bin(RawEdPK),
+    EccPK = blockchain_utils:pubkey_to_bin(RawEccPK),
+    EdPK = blockchain_utils:pubkey_to_bin(RawEdPK),
```
evanmcc, line 576: can we either be consistent (you've used both the literal module name and the macro), or better just omit the module name since it's a local call?

PR #742, opened 2021-02-22
Add a relay_swarm_reset command line helper

mrallen1:
Problem to solve: When CG members are connected through relays, sometimes the connection can get bad or slow which kills block times and sometimes completely stops progress altogether. We want to be able to have a quick way to kill a relay swarm and retry connections with a new one.

Solution: Implement a CLI option to do this.

### DIFF
--- a/src/cli/blockchain_cli_peer.erl
+++ b/src/cli/blockchain_cli_peer.erl
@@ -28,6 +28,7 @@ register_all_usage() ->
                   peer_gossip_peers_usage(),
                   peer_gossip_peers_usage(),
                   peer_refresh_usage(),
+                  peer_relay_reset_usage(),
                   peer_usage()
                  ]).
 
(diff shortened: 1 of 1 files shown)


src/cli/blockchain_cli_peer.erl
```diff
@@ -28,6 +28,7 @@ register_all_usage() ->
                   peer_gossip_peers_usage(),
                   peer_gossip_peers_usage(),
                   peer_refresh_usage(),
+                  peer_swarm_reset_usage(),
```
evanmcc, line 31: I'm not sure that the word swarm adds anything to the UI here can we remove it? `relay_reset` or `reset_relay` is probably fine.

evanmcc (approved):
other than the command name change this looks fine.

evanmcc (approved):

PR #743, opened 2021-02-24
Adapt cli commands for rebar 3.14

Vagabond:
Related to helium/miner#567


Vagabond:
One problem here is all the command outputs now ends with the line `ok` which breaks some scripts that expect the output on a single line.

Vagabond:
This depends on erlware/relx#855

Vagabond:
Additionally this branch works with OTP 24

rebar.config
```diff
@@ -29,7 +29,7 @@
 }.
 
 {deps, [
-    {lager, ".*", {git, "https://github.com/erlang-lager/lager.git", {branch, "master"}}},
+    {lager, "3.9.0"},
```
mrallen1, line 32: is there a reason you don't want to use 3.9.1?

src/cli/blockchain_console.erl
```diff
@@ -4,4 +4,22 @@
 
 -spec command([string()]) -> ok.
 command(Cmd) ->
-    clique:run(Cmd).
+    %% this is the contents of clique:run but
+    %% we want to figure out if the command worked
+    %% or not
```
mrallen1, line 9: This is ok for now but seems like it ought to be a new function in clique `run_and_return_status/1` or something...

mrallen1 (approved):

rebar.config
Vagabond: this PR is old, I will bump it

src/cli/blockchain_console.erl
Vagabond: yes, we should make an upstream PR

evanmcc (approved):

PR #744, opened 2021-02-24
bump p2p for enacl

evanmcc:
not just bumping this through because I'm not sure what other changes it might pull in


Vagabond (approved):
It's just one commit past where it was before

PR #745, opened 2021-02-26
send keys of changed items for each CF via commit hook end fun

andymck:
There are two commit hook types, one which is executed for each individual update to a CF and one which is executed when *all* updates for a CF for the current commit context is complete.

This change applies to the second and passes the list of all keys for each CF which were modified for the current commit context to the supplied fun.

This allows the interested service which defined the commit hook funs to be informed of which keys were modified for a CF and to take any necessary steps/actions.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -3518,18 +3518,27 @@ invoke_commit_hooks(Changes, Filters) ->
                 end,
                 Filters),
 
-                %% call the end hook on each group now that all incremental updates are applied
+              %% accumulate a list of all CFs which have updates
+              %% and for each updated CF accumulate a list of keys which have been modified
+              %% call the end funs for each CF with the CF atom and the list of modified keys
               maps:map(
                 fun(CF, HookList) ->
                         CFAtom = maps:get(CF, FiltersMap),
-                        lists:foreach(
-                          fun(#hook{hook_end_fun = HookFun}) ->
-                                  HookFun(CFAtom)
-                          end, HookList)
+                        %% check if any changes for each CF, if true fire the end fun, it false do nothing
+                        case maps:get(CFAtom, Groups, undefined) of
+                            undefined ->
+                                noop;
+                            CFChanges ->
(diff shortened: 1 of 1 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3518,18 +3518,27 @@ invoke_commit_hooks(Changes, Filters) ->
...
-                          fun(#hook{hook_end_fun = HookFun}) ->
-                                  HookFun(CFAtom)
-                          end, HookList)
+                        %% check if any changes for each CF, if true fire the end fun, it false do nothing
+                        case maps:get(CFAtom, Groups, undefined) of
+                            undefined ->
+                                noop;
+                            CFChanges ->
+                                %% iterate over the changes for this group and collect the keys which have been modified
+                                CFChangedKeys = lists:foldl(fun({_CF, Action, Key, _Value}, Acc) -> [{Action, Key} | Acc] end, [], CFChanges),
```
madninja, line 3533: Why not just use a map or list comprehension here?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -4496,7 +4505,7 @@ commit_hooks_test() ->
                         [{active_gateways,
                             undefined,
                             fun(Changes) -> Me ! {hook1, Changes} end,
-                            fun(_CF) -> Me ! {hook1, changes_complete} end
+                            fun(_CF, ChangedKeys) -> Me ! {hook1, changes_complete, ChangedKeys} end
```
madninja, line 4508: Maybe make this a tuple instead of a triplet.. More consistent matching?

src/ledger/v1/blockchain_ledger_v1.erl
andymck: yeah a list comprehension makes better sense here, will change

src/ledger/v1/blockchain_ledger_v1.erl
andymck: really these are just arbitrary payloads for the test.  Dont see much advantage to using a tuple over the triplet here ?

madninja (approved):

PR #746, opened 2021-02-26
adjust how we sync

evanmcc:
now that we can grab fully random peers we should use that capability to sync from non-gossip peers, both to avoid capture and avoid syncing over relays.

This needs some major cleanup, and it's not 100% done, but it indicates direction, and speeds up syncing quite a lot on new nodes.

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -754,6 +754,7 @@ start_sync(#state{blockchain = Chain, swarm = Swarm, swarm_tid = SwarmTID} = Sta
             State#state{sync_pid = Pid, sync_ref = Ref}
     end.
 
+-spec get_random_peer(SwarmTID :: ets:tab()) -> no_peers | string().
 get_random_peer(SwarmTID) ->
     Peerbook = libp2p_swarm:peerbook(SwarmTID),
     %% limit peers to random connections with public addresses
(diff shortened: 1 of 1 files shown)


src/blockchain_worker.erl
```diff
@@ -716,20 +718,23 @@ start_sync(#state{blockchain = Chain, swarm = Swarm, swarm_tid = SwarmTID} = Sta
...
-                         end
-                 end, Peers0).
+    Peerbook = libp2p_swarm:peerbook(SwarmTID),
+    %% limit peers to random connections with public addresses
+    F = fun(Peer) ->
+                lists:any(fun libp2p_transport_tcp:is_public/1,
+                          libp2p_peer:listen_addrs(Peer))
+        end,
+    case libp2p_peerbook:random(Peerbook, [], F, 100) of
+        false -> [];
```
andymck, line 734: i notice get_peer/1 returns an empty list when it fails to get a peer.   There are 4 calls to get_peer in blockchain_worker.  3 of these handle the empty list but this one does not

https://github.com/helium/blockchain-core/blob/ef90cf829c49d9e21dcb6cca2265c94194e00cb0/src/blockchain_worker.erl#L844

the next line will result in a bad arg as far as i can see

https://github.com/helium/blockchain-core/blob/ef90cf829c49d9e21dcb6cca2265c94194e00cb0/src/blockchain_worker.erl#L845

Not sure of the consequences of grab_snapshot/2 errorring out but worth raising

evanmcc: addressed this

Vagabond: grab_snapshot still uses get_random_peer without an empty lists check.

Why not merge get_peer into get_random_peer and have it return a tagged tuple result?

Vagabond:
Will this make sync dialer noiser in the logs?

vihu (approved):

PR #751, opened 2021-03-03
Rewards v2

mrallen1:
Problem to solve: We need to shrink the size and radically improve the efficiency of calculating rewards at the end of an epoch. We also want to make the "rewards metadata" more easily accessible to systems like ETL which power the Helium blockchain API service (so that it eventually can be exposed to end users in explorer and/or the mobile applications).

Solution: Fold over the blocks and their transactions in an epoch a single time, bound map memory size using iterators, allow a callback to be defined to get the rewards data.

### DIFF
--- a/rebar.lock
+++ b/rebar.lock
@@ -41,7 +41,7 @@
   0},
  {<<"helium_proto">>,
   {git,"https://github.com/helium/proto.git",
-       {ref,"5c7be1dda8853636e3fd1f8519bedf6a1911452b"}},
+       {ref,"ec3b6d0f2694434d2c71e91487ae5a1cac9f10d5"}},
   0},
  {<<"inert">>,{pkg,<<"inert">>,<<"1.0.1">>},2},
  {<<"inet_cidr">>,{pkg,<<"erl_cidr">>,<<"1.0.2">>},2},
--- a/rebar.lock
+++ b/rebar.lock
@@ -41,7 +41,7 @@
   0},
  {<<"helium_proto">>,
   {git,"https://github.com/helium/proto.git",
-       {ref,"2c1f62e4bf8013779fdf41da9cae46efdd323a64"}},
+       {ref,"5c7be1dda8853636e3fd1f8519bedf6a1911452b"}},
   0},
  {<<"inert">>,{pkg,<<"inert">>,<<"1.0.1">>},2},
  {<<"inet_cidr">>,{pkg,<<"erl_cidr">>,<<"1.0.2">>},2},
--- a/src/transactions/blockchain_txn.erl
+++ b/src/transactions/blockchain_txn.erl
@@ -37,7 +37,8 @@
              | blockchain_txn_price_oracle_v1:txn_price_oracle()
              | blockchain_txn_gen_price_oracle_v1:txn_genesis_price_oracle()
              | blockchain_txn_transfer_hotspot_v1:txn_transfer_hotspot()
-             | blockchain_txn_state_channel_close_v1:txn_state_channel_close().
+             | blockchain_txn_state_channel_close_v1:txn_state_channel_close()
+             | blockchain_txn_rewards_v2:txn_rewards_v2().
 
 -type before_commit_callback() :: fun((blockchain:blockchain(), blockchain_block:hash()) -> ok | {error, any()}).
 -type txns() :: [txn()].
@@ -113,7 +114,8 @@
     {blockchain_txn_price_oracle_v1, 23},
     {blockchain_txn_state_channel_close_v1, 24},
     {blockchain_txn_token_burn_v1, 25},
-    {blockchain_txn_transfer_hotspot_v1, 26}
+    {blockchain_txn_transfer_hotspot_v1, 26},
+    {blockchain_txn_rewards_v2, 27}
 ]).
 
 block_delay() ->
@@ -192,7 +194,9 @@ wrap_txn(#blockchain_txn_price_oracle_v1_pb{}=Txn) ->
 wrap_txn(#blockchain_txn_gen_price_oracle_v1_pb{}=Txn) ->
     #blockchain_txn_pb{txn={gen_price_oracle, Txn}};
 wrap_txn(#blockchain_txn_transfer_hotspot_v1_pb{}=Txn) ->
-    #blockchain_txn_pb{txn={transfer_hotspot, Txn}}.
+    #blockchain_txn_pb{txn={transfer_hotspot, Txn}};
+wrap_txn(#blockchain_txn_rewards_v2_pb{}=Txn) ->
+    #blockchain_txn_pb{txn={rewards_v2, Txn}}.
 
 -spec unwrap_txn(#blockchain_txn_pb{}) -> blockchain_txn:txn().
 unwrap_txn(#blockchain_txn_pb{txn={bundle, #blockchain_txn_bundle_v1_pb{transactions=Txns} = Bundle}}) ->
@@ -579,8 +583,9 @@ type(#blockchain_txn_price_oracle_v1_pb{}) ->
 type(#blockchain_txn_gen_price_oracle_v1_pb{}) ->
     blockchain_txn_gen_price_oracle_v1;
 type(#blockchain_txn_transfer_hotspot_v1_pb{}) ->
-    blockchain_txn_transfer_hotspot_v1.
-
+    blockchain_txn_transfer_hotspot_v1;
+type(#blockchain_txn_rewards_v2_pb{}) ->
+    blockchain_txn_rewards_v2.
 
 -spec validate_fields([{{atom(), iodata() | undefined},
                         {binary, pos_integer()} |
--- a/src/transactions/v1/blockchain_txn_vars_v1.erl
+++ b/src/transactions/v1/blockchain_txn_vars_v1.erl
@@ -930,6 +930,11 @@ validate_var(?reward_version, Value) ->
         _ ->
             throw({error, {invalid_reward_version, Value}})
     end;
+validate_var(?rewards_txn_version, Value) ->
+    case Value of
+        N when is_integer(N), N >= 1, N =< 2 -> ok;
+        _ -> throw({error, {invalid_rewards_txn_version, Value}})
+    end;
 
 %% bundle vars
 validate_var(?max_bundle_size, Value) ->
--- a/src/transactions/v2/blockchain_txn_rewards_v2.erl
+++ b/src/transactions/v2/blockchain_txn_rewards_v2.erl
@@ -1,0 +1,1615 @@
+%%-------------------------------------------------------------------
+%% @doc
+%% This module implements rewards v2 which track only an account and
+%% an amount. The breakdowns of rewards by type and associated gateway
+%% are not kept here.  This was done to streamline the increasing
+%% size of rewards as the network grows.
+%%
+%% In the future, we will need to work on further ways to streamline
+%% the size of this transaction. One proposal was to use the ledger
+%% as an encoding dictionary and use ledger offsets to mark the
+%% account instead of using the full sized account id.
+%% @end
+%%%-------------------------------------------------------------------
+-module(blockchain_txn_rewards_v2).
+
+-behavior(blockchain_txn).
+
+-behavior(blockchain_json).
+-include("blockchain_json.hrl").
+
+-include("blockchain_vars.hrl").
+-include_lib("helium_proto/include/blockchain_txn_rewards_v2_pb.hrl").
+
+-export([
+    new/3,
+    hash/1,
+    start_epoch/1,
+    end_epoch/1,
+    rewards/1,
+    sign/2,
+    fee/1,
+    is_valid/2,
+    absorb/2,
+    calculate_rewards/3,
+    calculate_rewards_metadata/3,
+    print/1,
+    to_json/2
+]).
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-export([v1_to_v2/1]).
+-endif.
+
+-type txn_rewards_v2() :: #blockchain_txn_rewards_v2_pb{}.
+-type reward_v2() :: #blockchain_txn_reward_v2_pb{}.
+-type reward_vars() :: map().
+-type rewards_share_map() :: #{ libp2p_crypto:pubkey_bin() => non_neg_integer() }.
+-type dc_rewards_share_map() :: #{ libp2p_crypto:pubkey_bin() => non_neg_integer(),
+                                   seen => [ blockchain_state_channel_v1:id() ] }.
+-type rewards_share_metadata() :: #{ poc_challenger => rewards_share_map(),
+                                     poc_challengee => rewards_share_map(),
+                                     poc_witness => rewards_share_map(),
+                                     dc_rewards => dc_rewards_share_map(),
+                                     overages => non_neg_integer() }.
+-type owner_key() :: {owner, securities, libp2p_crypto:pubkey_bin()}.
+-type reward_types() :: poc_challengers | poc_challengees | poc_witnesses | data_credits | consensus.
+-type gateway_key() :: {gateway, reward_types(), libp2p_crypto:pubkey_bin()}.
+-type rewards_map() :: #{ owner_key() | gateway_key() => number() }.
+-type rewards_metadata() :: #{ poc_challenger => rewards_map(),
+                               poc_challengee => rewards_map(),
+                               poc_witness => rewards_map(),
+                               dc_rewards => rewards_map(),
+                               consensus_rewards => rewards_map(),
+                               securities_rewards => rewards_map(),
+                               overages => non_neg_integer() }.
+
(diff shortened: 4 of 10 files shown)


src/transactions/v2/blockchain_txn_reward_v2.erl
```diff
@@ -0,0 +1,147 @@
...
+    Reward#blockchain_txn_reward_v2_pb.amount.
+
+%%--------------------------------------------------------------------
+%% @doc
+%% Determine if a reward_v2 item is valid.
+%% @end
+%%--------------------------------------------------------------------
+-spec is_valid(Reward :: reward_v2()) -> ok | {error, Reason :: term()}.
+is_valid(#blockchain_txn_reward_v2_pb{account=Account, amount=Amount}) ->
+    case erlang:is_binary(Account) andalso Amount > 0 of
```
Vagabond, line 82: You can check account validity better here, the keys have structure.

src/transactions/v2/blockchain_txn_reward_v2.erl
```diff
@@ -0,0 +1,147 @@
+%%%-------------------------------------------------------------------
+%% @doc This module implements reward_v2 items which only record an
```
Vagabond, line 2: I wonder if we need a module for just a two field record?

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -0,0 +1,1444 @@
...
+-spec is_valid(txn_rewards_v2(), blockchain:blockchain()) -> ok | {error, atom()} | {error, {atom(), any()}}.
+is_valid(Txn, Chain) ->
+    Start = ?MODULE:start_epoch(Txn),
+    End = ?MODULE:end_epoch(Txn),
+    case ?MODULE:calculate_rewards(Start, End, Chain) of
+        {error, _Reason}=Error ->
+            Error;
+        {ok, CalRewards} ->
+            TxnRewards = ?MODULE:rewards(Txn),
+            CalRewardsHashes = lists:sort([blockchain_txn_reward_v2:hash(R)|| R <- CalRewards]),
```
Vagabond, line 86: Should reward order not be deterministic, so we can avoid sorting here?

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -0,0 +1,1444 @@
...
+-spec absorb(txn_rewards_v2(), blockchain:blockchain()) -> ok | {error, atom()} | {error, {atom(), any()}}.
+absorb(Txn, Chain) ->
+    Ledger = blockchain:ledger(Chain),
+    Rewards = ?MODULE:rewards(Txn),
+    AccRewards = lists:foldl(
+        fun(Reward, Acc) ->
+            Account = blockchain_txn_reward_v2:account(Reward),
+            Amount = blockchain_txn_reward_v2:amount(Reward),
+            Total = maps:get(Account, Acc, 0),
+            maps:put(Account, Total + Amount, Acc)
```
Vagabond, line 103: I wonder if it's more efficient to just loop once and call credit account each time, all the ledger operations are cached/batched.

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -0,0 +1,1444 @@
...
+                                                                  A, Chain, Ledger, Vars)
+                                 end,
+                                 Acc, Txns),
+            fold_blocks_for_rewards(Current+1, End, Chain, Vars, Ledger, NewAcc)
+    end.
+
+calculate_reward_for_txn(?MODULE, _Txn, _End, _Acc, _Chain,
+                         _Ledger, _Vars) -> throw({error, already_existing_rewards_v2});
+calculate_reward_for_txn(blockchain_txn_rewards_v1, _Txn, _End, _Acc, _Chain,
+                         _Ledger, _Vars) -> throw({error, already_existing_rewards_v1});
```
Vagabond, line 212: :ok_hand:

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -0,0 +1,1444 @@
...
+            Balance = blockchain_ledger_security_entry_v1:balance(Entry),
+            PercentofReward = (Balance*100/TotalSecurities)/100,
+            Amount = erlang:round(PercentofReward*SecuritiesReward),
+            maps:put({owner, securities, Key}, Amount, Acc)
+        end,
+        #{},
+        Securities
+    ).
+
+poc_challenger_reward(Txn, #{poc_challenger := ChallengerRewards} = Acc,
```
Vagabond, line 505: Be nice to have specs on these functions as well

Vagabond:
I haven't reviewed the reward logic in depth (I assume it's mostly taken from the old version) but the general approach looks good.

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -0,0 +1,1444 @@
...
+    %% so we will do that top level work here. If we get a thrown error while
+    %% we are folding, we will abort reward calculation.
+
+    %% Initialize our reward accumulator. We are going to build up a map which
+    %% will be in the shape of
+    %% #{ reward_type => #{ Entry => Amount } }
+    %%
+    %% where Entry is of the the shape
+    %% {owner, reward_type, Owner} or
+    %% {gateway, reward_type, Gateway}
```
Vagabond, line 144: It would be nice if we could optionally omit gateway accounting entirely?

src/transactions/blockchain_txn.erl
```diff
@@ -192,7 +194,9 @@ wrap_txn(#blockchain_txn_price_oracle_v1_pb{}=Txn) ->
 wrap_txn(#blockchain_txn_gen_price_oracle_v1_pb{}=Txn) ->
     #blockchain_txn_pb{txn={gen_price_oracle, Txn}};
 wrap_txn(#blockchain_txn_transfer_hotspot_v1_pb{}=Txn) ->
-    #blockchain_txn_pb{txn={transfer_hotspot, Txn}}.
+    #blockchain_txn_pb{txn={transfer_hotspot, Txn}};
+wrap_txn(#blockchain_txn_rewards_v2_pb{}=Txn) ->
+    #blockchain_txn_pb{txn={rewards_v2, Txn}};
```
vihu, line 199:
```suggestion
    #blockchain_txn_pb{txn={rewards_v2, Txn}}.
```

src/transactions/v2/blockchain_txn_rewards_v2.erl
mrallen1: Yeah I just cleaned this up

src/transactions/v2/blockchain_txn_rewards_v2.erl
mrallen1: agreed

src/transactions/blockchain_txn.erl
mrallen1: yeah fixed, derp sorry

src/transactions/v2/blockchain_txn_rewards_v2.erl
mrallen1: I mean, we could but I was under the impression that we _wanted_ to have the gateway address for rewards broken out. @abhay do you have thoughts here?

src/transactions/v2/blockchain_txn_rewards_v2.erl
mrallen1: Sure, we can sort the list after it's generated in the `prepare` function or we could insert entries into an `ordset` to sort as we insert new items.  I think it's fool's gold to assume the entry traversal order for maps is going to be stable.

src/transactions/v2/blockchain_txn_reward_v2.erl
mrallen1: No? idk, was just following tradition of 1:1 proto -> erlang module - we can nuke it if you want.

src/transactions/v2/blockchain_txn_reward_v2.erl
mrallen1: ok, what should I be testing for? it isn't obvious to me (also it's implemented like this in v1 - not that it is an excuse...)

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -0,0 +1,1448 @@
...
+                    Amount = erlang:round(PercentofReward*ConsensusReward),
+                    maps:put({gateway, consensus, Member}, Amount+OveragePerMember, Acc)
+                end,
+                #{},
+                ConsensusMembers
+            )
+    end.
+
+-spec securities_rewards(blockchain_ledger_v1:ledger(),
+                         map()) -> #{{owner, libp2p_crypto:pubkey_bin()} => non_neg_integer()}.
```
andymck, line 486: spec appears incorrect here for the corresponding maps:put({owner, securities, Key} below

andymck (commented):
lgtm, just a couple of comments.

Overall tho, there is a lot of duped internal functions between v1 and v2....is it possible to break these out into a common module  ?  Not sure if there are subtle deviations between both versions that may prevent this but on an initial pass there does appear to be some scope for this.

src/transactions/blockchain_txn.erl
```diff
@@ -37,7 +37,8 @@
              | blockchain_txn_price_oracle_v1:txn_price_oracle()
              | blockchain_txn_gen_price_oracle_v1:txn_genesis_price_oracle()
              | blockchain_txn_transfer_hotspot_v1:txn_transfer_hotspot()
-             | blockchain_txn_state_channel_close_v1:txn_state_channel_close().
+             | blockchain_txn_state_channel_close_v1:txn_state_channel_close()
```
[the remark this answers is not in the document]
mrallen1: yeah it doesn't

src/transactions/v2/blockchain_txn_rewards_v2.erl
mrallen1: thanks - we are debating removing the reward type from tuple

src/transactions/v2/blockchain_txn_rewards_v2.erl
abhay: I thought I responded here last night. Oops. 

We want to have gateway and validator accounting here so that ETL can write that information to the db.

@madninja also have you thought about having new break out tables since the raw transactions won't have the specific reward types any more?

src/transactions/v2/blockchain_txn_rewards_v2.erl
Vagabond: Maybe we should not use a map then? transactions and blocks have a known order at least..?

src/transactions/v2/blockchain_txn_rewards_v2.erl
Vagabond: only if you're ETL

vihu:
@mrallen1 Not sure if you've gotten to it yet but dialyzer seems pretty sad here...

mrallen1:
Yeah I know - I just finished adding specs to everything late Friday - havent had a chance to fix it up

src/transactions/blockchain_txn.erl
```diff
@@ -579,7 +583,9 @@ type(#blockchain_txn_price_oracle_v1_pb{}) ->
 type(#blockchain_txn_gen_price_oracle_v1_pb{}) ->
     blockchain_txn_gen_price_oracle_v1;
 type(#blockchain_txn_transfer_hotspot_v1_pb{}) ->
-    blockchain_txn_transfer_hotspot_v1.
+    blockchain_txn_transfer_hotspot_v1;
+type(#blockchain_txn_rewards_v2_pb{}) ->
+    blockchain_txn_rewards_v2.
```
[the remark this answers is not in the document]
mrallen1: Where does that definition go @vihu ?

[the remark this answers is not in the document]
vihu: There's an `actor/1` fun in `blockchain_txn` module :)

vihu:
> Yeah I know - I just finished adding specs to everything late Friday - havent had a chance to fix it up

Following up from our discussion on discord, I think that the `finalize_rewards_calculations/3` function is making dialyzer a bit sad. The `AccIn` type is `rewards_share_metadata` and the output is `rewards_metadata`; this won't work as the value of the `metadata_key` differs between `rewards_share_metadata` and `rewards_metadata`. One way would be to have a conversion function and change rewards_share_metadata -> rewards_metadata if possible, I believe we have all the information required to do that conversion. I think that would atleast solve one dialyzer bug here.

rebar.lock
```diff
@@ -114,5 +114,23 @@
  {<<"ranch">>, <<"F04166F456790FEE2AC1AA05A02745CC75783C2BFB26D39FAF6AEFC9A3D3A58A">>},
  {<<"rand_compat">>, <<"011646BC1F0B0C432FE101B816F25B9BBB74A085713CEE1DAFD2D62E9415EAD3">>},
  {<<"small_ints">>, <<"82A824C8794A2DDC73CB5CD00EAD11331DB296521AD16A619C13D668572B868A">>},
- {<<"throttle">>, <<"E881B46D9836AFB70F3E2FA3BE9B0140805BA324ED26AA734FF6C5C1568C6CA7">>}]}
+ {<<"throttle">>, <<"E881B46D9836AFB70F3E2FA3BE9B0140805BA324ED26AA734FF6C5C1568C6CA7">>}]},
```
vihu, line 117: These were edited?

src/transactions/blockchain_txn.erl
```diff
@@ -808,6 +816,8 @@ actor(Txn) ->
         blockchain_txn_state_channel_close_v1 ->
             %% group by owner
             blockchain_txn_state_channel_close_v1:state_channel_owner(Txn);
+        blockchain_txn_reward_v2 ->
```
vihu, line 819:
```suggestion
        blockchain_txn_rewards_v2 ->
```

vihu: The actual txn which appears on change is in the plural right?

mrallen1: yes it is - that's a typo - good catch, thanks

mrallen1: oh wait - no that is correct as written

mrallen1: It's written like that because there's no explicit module that handles the `reward_v2` record type which _does_ exist (because of protobufs, it has to be defined and repeated)

rebar.lock
mrallen1: yeah i don't know :shrug: i didn't edit them by hand or anything, maybe rebar3 rewrote the file or something?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -928,6 +928,11 @@ validate_var(?reward_version, Value) ->
         _ ->
             throw({error, {invalid_reward_version, Value}})
     end;
+validate_var(?rewards_txn_version, Value) ->
+    case Value of
+        N when is_integer(N), N >= 1, N =< 2 -> ok;
```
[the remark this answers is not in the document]
mrallen1: Because they're the only valid values? If we add more rewards transactions, it'll require a new code release anyway - but I can update this if you want

src/transactions/v2/blockchain_txn_rewards_v2.erl
mrallen1: OK, so what I'm hearing @madninja is that I need to resolve the gateway owners for all of the _gateway_ type rewards, right?

madninja: yeah, I believe so.. It'd be nice to understand what the mechanism is for ETL? That callback app env variable question above, basically

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -0,0 +1,1615 @@
...
+    lists:foreach(
+        fun(#blockchain_txn_reward_v2_pb{account=Account, amount=Amount}) ->
+            ok = blockchain_ledger_v1:credit_account(Account, Amount, Ledger)
+        end,
+        ?MODULE:rewards(Txn)
+    ).
+
+-spec calculate_rewards(non_neg_integer(), non_neg_integer(), blockchain:blockchain()) ->
+    {ok, [reward_v2()]} | {error, any()}.
+%% @doc Calculate and return an ordered list of rewards for use in a rewards_v2
```
Vagabond, line 133: Ordered by what?

mrallen1: lists:sort/1

mrallen1: so it would be by account

Vagabond: Updating comment would be good

Vagabond (approved):

PR #753, opened 2021-03-09
Improved validator checking

Vagabond:


src/transactions/v1/blockchain_txn_add_gateway_v1.erl
```diff
@@ -311,27 +311,32 @@ is_valid(Txn, Chain) ->
         {_, _, _, false} ->
             {error, payer_invalid_staking_key};
         {true, true, true, true} ->
-            AreFeesEnabled = blockchain_ledger_v1:txn_fees_active(Ledger),
```
evanmcc, line 314: do we need to check this in transfer hotspot too?  or does that only transfer between owning accounts?

evanmcc (approved):

src/transactions/v1/blockchain_txn_add_gateway_v1.erl
Vagabond: I'm guessing there should be a fee there as well

PR #761, opened 2021-03-24
Fix h3dex type

xandkar:
Looks like this size check has been failing silently.

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -375,7 +375,7 @@ handle_call({install_snapshot, Hash, Snapshot}, _From,
                 %% will have come from the snap.
                 true = erlang:is_map(Snapshot), % fail into the catch if it's an older record
                 H3dex = maps:get(h3dex, Snapshot), % fail into the catch if it's missing
-                case maps:size(H3dex) > 0 of
+                case length(H3dex) > 0 of
                     true -> ok;
                     false -> throw(bootstrap) % fail into the catch it's an empty default value
                 end
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -18,6 +18,10 @@
          diff/2
         ]).
 
+-export_type([
+    snapshot/0
+]).
+
 %% this assumes that everything will have loaded the genesis block
 %% already.  I'm not sure that's totally safe in all cases, but it's
 %% the right thing for the spots and easy to work around elsewhere.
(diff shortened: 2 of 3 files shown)


src/blockchain.app.src
```diff
@@ -1,6 +1,6 @@
 {application, blockchain, [
     {description, "Blockchain OTP Library"},
-    {vsn, "0.1.0"},
+    {vsn, "0.1.1"},
```
vihu, line 3: I am not too certain if this is required here. But I also don't know of a good time to bump this, maybe we do it on OTP bumps?

xandkar: Yeah, that is probably a topic for broader discussion, since the version doesn't seem to be actually used so far. I'd generally favor incrementing it and git-tagging in accordance with the rules of SemVer.

I know that SemVer is somewhat controversial - not all are convinced of its value... Nonetheless, I think it is a useful way to communicate intentions.

mrallen1 (approved):

src/blockchain.app.src
mrallen1: kinda :shrug: on this but w/e

evanmcc: yeah, just remove this for now, we can have a better discussion about this later.

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -28,6 +32,54 @@
 %% more permanent and less flexible format, like protobufs, or
 %% cauterize.
 
+-type snapshot() ::
```
evanmcc, line 35: there's already a hrl file that has this sort of stuff in there, can we do this there?

src/blockchain_worker.erl
```diff
@@ -375,7 +375,7 @@ handle_call({install_snapshot, Hash, Snapshot}, _From,
                 %% will have come from the snap.
                 true = erlang:is_map(Snapshot), % fail into the catch if it's an older record
                 H3dex = maps:get(h3dex, Snapshot), % fail into the catch if it's missing
-                case maps:size(H3dex) > 0 of
```
evanmcc, line 378: great catch

evanmcc (changes requested):

evanmcc:
I think that this is fine, it just needs some cleanup.

src/blockchain.app.src
xandkar: Reverted and rebased.

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
xandkar: The `.hrl` is needed because it is the only way to use records in other modules, while this is a map and can be used directly. Now that the type is exported, other modules can refer to it as `blockchain_ledger_snapshot_v1:snapshot()` and access it directly, or we can make it `-opaque`, but in either case - I don't think it makes sense to have it in a header file. Perhaps another module, if the goal is to keep this one small?

evanmcc (approved):

PR #762, opened 2021-03-24
Fix up to_json for rewards_v2

madninja:
This fixes up to_json for rewards_v2 to include detailed rewards when a chain is passed as one of the options.


Vagabond (approved):

mrallen1 (commented):
lgtm

PR #764, opened 2021-04-01
add follow mode

evanmcc:
allow the chain to be set in follow only mode, which validates that a block is signed correctly and follows the chain, but does not validate transactions.


vihu (approved):
lgtm

PR #765, opened 2021-04-02
don't fetch a million blocks for their height

evanmcc:
this PR both increases the pace of poc gc (trying to make them cheaper), and also fetches only the block hash for the relevant blocks from the disk, which should make the gc dramatically cheaper.

it also removes a large sync write from the state channel client which doesn't seem to be required.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -442,6 +442,8 @@ import(Chain, SHA,
                  blockchain_ledger_v1:commit_context(Ledger)
              end
              || Mode <- [delayed, active]],
+            %% TODO: it might make more sense to do this block by block?  it will at least be
+            %% cheaper to do it that way.
             Ledger2 = blockchain_ledger_v1:new_context(Ledger0),
             Chain1 = blockchain:ledger(Ledger2, Chain),
             {ok, Curr2} = blockchain_ledger_v1:current_height(Ledger2),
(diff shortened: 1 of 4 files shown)


mrallen1 (approved):
lgtm - imo would be worth a couple comments in the PR how this is intended to conserve memory

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1581,11 +1581,22 @@ maybe_gc_pocs(Chain, Ledger) ->
                   {ok, V} -> V;
                   _ -> 1
               end,
-    case Version > 3 andalso Height rem 100 == 0 of
+    case Version > 3 andalso Height rem 51 == 0 of
```
macpie, line 1584: 👍🏻

macpie (approved):

src/ledger/v1/blockchain_ledger_v1.erl
macpie: Could this cause any sync issue?

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -464,7 +464,7 @@ import(Chain, SHA,
                                   end,
 
                               Ht = blockchain_block:height(Block),
-                              case blockchain:get_block(Ht, Chain) of
+                              case blockchain:get_block_hash(Ht, Chain) of
                                   {ok, _Block} ->
```
mrallen1, line 468: `_Hash` here?

mrallen1 (approved):
just the one suggestion

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
evanmcc: will do

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: no, it won't.  the threshold for deletion is still the same, it just happens more often

PR #767, opened 2021-04-05
fix delete conversion typo

evanmcc:
in converting this from a batch write, I forgot to add the option list to the individual deletes, which is causing the miner to crash every time it tries to gc blocks.


vihu (approved):
whoops, nice catch

PR #768, opened 2021-04-07
Add support for payment-v2 memo field

vihu:
Enables support for specifying a uint64 memo for a payment-v2.

Notes:
- Introduces `allow_payment_v2_memos` boolean chain variable
- If the var is not set, existing payment-v2s will have memo=0 as default
- If the var is set, as long as the memo lies between 0 - 2^64 it will be considered valid
- Add some test cases to verify said logic
- Update and refactor payment-v2 txn validation

TODO:
- [x] Switch proto to master once [proto#72](https://github.com/helium/proto/pull/72) lands.

### DIFF
--- a/src/transactions/v2/blockchain_payment_v2.erl
+++ b/src/transactions/v2/blockchain_payment_v2.erl
@@ -15,6 +15,7 @@
          payee/1,
          amount/1,
          memo/1, memo/2,
+         is_valid_memo/1,
          print/1,
          to_json/2
         ]).
@@ -62,6 +63,16 @@ memo(Payment) ->
 -spec memo(Payment :: payment(), Memo :: non_neg_integer()) -> payment().
 memo(Payment, Memo) ->
     Payment#payment_pb{memo=Memo}.
+
+-spec is_valid_memo(Payment :: payment()) -> boolean().
+is_valid_memo(#payment_pb{memo = Memo}) ->
+    try
+        Bin = binary:encode_unsigned(Memo, big),
+        bit_size(Bin) =< 64
+    catch _:_ ->
+            %% we can't do this, invalid
+            false
+    end.
 
 print(undefined) ->
     <<"type=payment undefined">>;
--- a/src/transactions/v2/blockchain_payment_v2.erl
+++ b/src/transactions/v2/blockchain_payment_v2.erl
@@ -72,7 +72,8 @@ print(#payment_pb{payee=Payee, amount=Amount, memo=Memo}) ->
 to_json(Payment, _Opts) ->
     #{
       payee => ?BIN_TO_B58(payee(Payment)),
-      amount => amount(Payment)
+      amount => amount(Payment),
+      memo => ?MAYBE_UNDEFINED(memo(Payment))
      }.
 
 %% ------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


src/transactions/v2/blockchain_payment_v2.erl
```diff
@@ -43,10 +55,18 @@ payee(Payment) ->
...
+
 print(undefined) ->
     <<"type=payment undefined">>;
-print(#payment_pb{payee=Payee, amount=Amount}) ->
-    io_lib:format("type=payment payee: ~p amount: ~p", [?TO_B58(Payee), Amount]).
+print(#payment_pb{payee=Payee, amount=Amount, memo=Memo}) ->
+    io_lib:format("type=payment payee: ~p amount: ~p, memo: ~p", [?TO_B58(Payee), Amount, Memo]).
 
 -spec to_json(payment(), blockchain_json:opts()) -> blockchain_json:json_object().
 to_json(Payment, _Opts) ->
```
madninja, line 72: Missing memo entry in to_json?

vihu: ah yes, nice catch. updated 3d966b2

evanmcc (approved):

test/blockchain_payment_v2_SUITE.erl
```diff
@@ -348,3 +363,220 @@ negative_amount_test(Config) ->
...
+    Payment1 = blockchain_payment_v2:new(Recipient, Amount, ?INVALID_GIANT_MEMO),
+
+    Tx = blockchain_txn_payment_v2:new(Payer, [Payment1], 1),
+    SigFun = libp2p_crypto:mk_sig_fun(PayerPrivKey),
+    SignedTx = blockchain_txn_payment_v2:sign(Tx, SigFun),
+
+    %% Placeholder: update when we land a better version of add_block in test_utils
+    %% which does validations, this will suffice for now
+    %% NOTE: SignedTx being in the second pos implies it's invalid
+    {[], [{SignedTx, InvalidReason}]} = blockchain_txn:validate([SignedTx], Chain),
```
Vagabond, line 564: You should encode/decode the txn before trying to validate it. This will likely close the issue with an out of range memo

src/transactions/v2/blockchain_payment_v2.erl
```diff
@@ -43,16 +56,35 @@ payee(Payment) ->
...
+-spec memo(Payment :: payment()) -> undefined | non_neg_integer().
+memo(Payment) ->
+    Payment#payment_pb.memo.
+
+-spec memo(Payment :: payment(), Memo :: non_neg_integer()) -> payment().
+memo(Payment, Memo) ->
+    Payment#payment_pb{memo=Memo}.
+
+-spec is_valid_memo(Payment :: payment()) -> boolean().
+is_valid_memo(#payment_pb{memo = Memo}) ->
```
Vagabond, line 68: I believe this is unnecessary

vihu:
> Did we want to land this?

I think this is good to land despite differing slightly from HIP26, but [the author](https://github.com/helium/HIP/issues/125#issuecomment-816084955) seems okay with this implementation as well.

Vagabond (approved):

PR #770, opened 2021-04-08
fix snapshots

evanmcc:
for the sake of determinism and snapshots, upgrade all the gateways to be consistent


Vagabond (approved):
This seems to remove the drift

PR #776, opened 2021-04-14
Proof of Coverage - V11

vihu:
Add support for poc_version 11
----

Region related changes:

- Add support for LoraWAN regions
- Add support for LoraWAN regional parameters

The above will be directed via chain variables, refer to `blockchain_vars.hrl`, `blockchain_region.hrl` for the newly introduced variables and `blockchain_region_SUITE.erl` for currently specified usage.

POC related changes:

- Add support for poc_version = 11
- Make POC validation region aware. Src/Dst must be in the same region
- Make POC validation tx/rx gain aware. The receipts/witnesses must follow legal allowed limits for packet transmissions

TODO:
- [x] More testing
- [x] Even more testing with aux ledger
- [x] Finalize poc validations
- [x] Finalize loraWAN region parameters
- [x] Land https://github.com/helium/proto/pull/77

Newly introduced chain variables:
```
%% This will be a list of supported regions
regulatory_regions

%% Each of these would be a binary blob of serialized H3 indices
region_as923_1
region_as923_2
region_as923_3
region_as923_4
region_au915
region_cn470
region_eu433
region_eu868
region_in865
region_kr920
region_ru864
region_us915

%% Each of these would be region specific parameters (binary blob)
region_as923_1_params
region_as923_2_params
region_as923_3_params
region_as923_4_params
region_au915_params
region_cn470_params
region_eu433_params
region_eu868_params
region_in865_params
region_kr920_params
region_ru864_params
region_us915_params


%% poc specific
fspl_loss (float)
poc_distance_limit (int)
check_snr (boolean)
```


JayKickliter:
This PR will need:
- AS923-4 region
  - new polyfill added generated by https://github.com/helium/lorawan-h3
  - sourced from @gradoj https://github.com/gradoj/hplans

include/blockchain_region.hrl
```diff
@@ -0,0 +1,473 @@
...
+-define(REGION_PARAMS_US915, [
+    [
+        % in hz
+        {<<"channel_frequency">>, 903900000},
+        % in hz
+        {<<"bandwidth">>, 125000},
+        % dBi x 10
+        {<<"max_eirp">>, 360},
+        % will be converted to atoms
+        {<<"spreading">>, [<<"sf10">>, <<"sf9">>, <<"sf8">>, <<"sf7">>]}
```
michaeldjeffrey, line 29: SF7 is RFU in US915.
image
I'm assuming that's the table that's being encoded here. from the rp_2-1.0.1.pdf

gradoj:
I'm pretty certain that https://github.com/gradoj/hplans is now out of sync with the csv file that is currently in use: https://github.com/helium/miner/blob/master/priv/countries_reg_domains.csv with the pr's for new regions from Nebra.
I will add a new region 923-4 for the polyfills and try and get synced back up with the latest countries_reg_domains.csv

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1132,20 +1132,30 @@ valid_receipt(PreviousElement, Element, Channel, Ledger) ->
...
+                         {ok, L} ->
+                             D = blockchain_utils:distance(SourceLoc, DestinationLoc),
+                             D > L;
+                         _ ->
+                             false
+                     end,
+
+            SameRegion = is_same_region(Ledger, SourceLoc, DestinationLoc),
+
             try h3:grid_distance(SourceParentIndex, DestinationParentIndex) >= ExclusionCells of
```
michaeldjeffrey, line 1153: Was `>= ExclusionCells` supposed to be removed from this `try`?

shawaj:
@gradoj let me know if you need any help from our side or anything!

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Well removing that from the `try` would end up breaking backwards compatibility (i.e. if `poc-version<10`, we want to do the things we have always been doing).

But if your question is about whether with the introduction of poc-11 we should not do exclusion altogether...then...maybe? I don't see why we shouldn't though. Two hotspots witnessing close to each don't really provide much useful network coverage I suppose.

vihu:
> @gradoj let me know if you need any help from our side or anything!

@shawaj Since you're here :) would you mind taking a look through [the region header](https://github.com/helium/blockchain-core/blob/781c05e2948cb35fab7016377d781c7daab4f445/include/blockchain_region.hrl) please? The intention is to encapsulate all (most) known LoraWAN regional parameters (or atleast the ones which helium blockchain cares about in the first pass) as chain variables.

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
michaeldjeffrey:
```suggestion
            try h3:grid_distance(SourceParentIndex, DestinationParentIndex) of
```
Sorry, I meant this change. 
`>= ExclusionCells`
would caused to never match the first case.

vihu: Wow, good catch :pray: . I r bad at seeing things.

gradoj:
@JayKickliter I've added AS923-4 to the geojson files but it would be great if @shawaj could review/compare the regions with their database before the h3 are generated again. https://github.com/gradoj/hplans/blob/main/regions.csv

include/blockchain_region.hrl
```diff
@@ -0,0 +1,473 @@
+-define(SUPPORTED_REGIONS, [
```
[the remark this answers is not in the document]
vihu: Hmm, they are as `regulatory_regions`. My intention was to use this constant to verify that the var would be set exactly to these values https://github.com/helium/blockchain-core/blob/ac6d47376a30d97470f6d6b99bb9db9f19211a34/src/transactions/v1/blockchain_txn_vars_v1.erl#L1248

But I suppose that makes it troublesome for updating if we need to add/modify these down the line

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1260,15 +1279,39 @@ valid_witnesses(Element, Channel, Ledger) ->
...
 
+-spec is_same_region(
+    Ledger :: blockchain_ledger_v1:ledger(),
+    SourceLoc :: h3:h3_index(),
+    DstLoc :: h3:h3_index()
+) -> boolean().
+is_same_region(Ledger, SourceLoc, DstLoc) ->
+    case blockchain_region_v1:region(SourceLoc, Ledger) of
+        {ok, R1} ->
+            case blockchain_region_v1:region(DstLoc, Ledger) of
```
Vagabond, line 1299: Would suggest *only* searching the region the first one was in, not all regions.

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1409,6 +1478,40 @@ get_channels(Txn, Chain) ->
...
+min_rcv_sig(Receipt, Ledger, SourceLoc, DstPubkeyBin, DestinationLoc, Freq) ->
+    case blockchain:config(?poc_version, Ledger) of
+        {ok, POCVersion} when POCVersion >= 11 ->
+            TxPower = maybe_tx_power_from_receipt(Receipt, SourceLoc, Ledger),
+            SrcPubkeyBin = blockchain_poc_receipt_v1:gateway(Receipt),
+            {ok, DstGW} = blockchain_ledger_v1:find_gateway_info(DstPubkeyBin, Ledger),
+            {ok, SrcGW} = blockchain_ledger_v1:find_gateway_info(SrcPubkeyBin, Ledger),
+            GT = blockchain_ledger_gateway_v2:gain(SrcGW),
+            GR = blockchain_ledger_gateway_v2:gain(DstGW),
+            %% {ok, Loss} = blockchain:config(?fspl_loss, Ledger),
```
Vagabond, line 1490: Where's the loss?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1409,6 +1478,40 @@ get_channels(Txn, Chain) ->
...
+            blockchain_utils:min_rcv_sig(
+                blockchain_utils:free_space_path_loss(SourceLoc, DestinationLoc, Freq)
+            )
+    end.
+
+maybe_tx_power_from_receipt(undefined, SourceLoc, Ledger) ->
+    {ok, Region} = blockchain_region_v1:region(SourceLoc, Ledger),
+    {ok, RegionParams} = blockchain_region_params_v1:params_for_region(Region, Ledger),
+    %% NOTE: all region params have the same max_eirp afaict, just take one
+    %% TODO: maybe look at the freq of the source and match max_eirp if they ever differ?
```
Vagabond, line 1509: Yes, you should match the frequency as some regions may have channels with a higher EIRP

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -404,6 +404,11 @@ get_reward_vars(Start, End, Ledger) ->
                         _ -> undefined
                     end,
 
+    WitnessRewardDecayRate = case blockchain:config(?witness_reward_decay_rate, Ledger) of
+                        {ok, DR} -> DR;
+                        _ -> 0
```
Vagabond, line 409: Should this be 1?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1238,10 +1239,121 @@ validate_var(?penalty_history_limit, Value) ->
     %% also low end cannot be 0
     validate_int(Value, "penalty_history_limit", 10, 100000, false);
 
+validate_var(?regulatory_regions, Value) when is_binary(Value) ->
+    %% The only regulatory_regions value we support must look like this:
+    %% <<"as923_1,as923_2,as923_3,au915,cn470,eu433,eu868,in865,kr920,ru864,us915">>
+    %% The order does not matter in validation
+    case blockchain_region_v1:get_regulatory_regions_var(Value) of
+        {ok, ValueList} ->
+            case ValueList -- ?SUPPORTED_REGIONS == [] of
```
Vagabond, line 1248: I feel like tying a chain var to a header value sort of makes using a chain var sort of pointless if we also need to update the code to support a new region.

vihu: Adding this here for reference https://github.com/helium/blockchain-core/pull/822

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Updated in 30651bc12d7e85bb05db75e7e4f56184b1d41c95

src/region/blockchain_region_v1.erl
```diff
@@ -0,0 +1,75 @@
...
+    case blockchain:config(?regulatory_regions, Ledger) of
+        {ok, Bin} ->
+            {ok, [list_to_atom(I) || I <- string:tokens(binary:bin_to_list(Bin), ",")]};
+        _ ->
+            {error, regulatory_regions_not_set}
+    end.
+
+-spec region(H3 :: h3:h3_index(), Ledger :: blockchain_ledger_v1:ledger()) ->
+    {ok, atom()} | {error, any()}.
+region(H3, Ledger) ->
```
Vagabond, line 30: I'd like a better name for this, like h3_to_region.

Also I think we want a way to check if a h3 is in a specific region (and not search them all).

src/transactions/v1/blockchain_poc_receipt_v1.erl
```diff
@@ -188,13 +195,17 @@ print(#blockchain_poc_receipt_v1_pb{
          gateway=Gateway,
          timestamp=TS,
          signal=Signal,
+         frequency=Freq,
+         snr=SNR,
          origin=Origin
         }) ->
-    io_lib:format("type=receipt gateway: ~s timestamp: ~b signal: ~b origin: ~p",
+    io_lib:format("type=receipt gateway: ~s timestamp: ~b signal: ~b snr: ~p freq: ~p origin: ~p",
```
Vagabond, line 202: Can we add tx power here as well?

src/region/blockchain_region_v1.erl
vihu: Added in 38ae995

src/transactions/v1/blockchain_poc_receipt_v1.erl
vihu: Added in c998c25

src/transactions/v1/blockchain_txn_rewards_v1.erl
vihu: Honestly, I was thinking whether we should even bother to update `rewards_v1` at all...

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Oops, I commented it out earlier for some reason, added back in 299a288

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Added 17ba24b6f565d8d8fe31ecf61f939bb8726f3ad6

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1281,19 +1331,31 @@ tagged_witnesses(Element, Channel, Ledger) ->
...
                          DestinationParentIndex = h3:parent(DestinationLoc, ParentRes),
-                         try h3:grid_distance(SourceParentIndex, DestinationParentIndex) >= ExclusionCells of
-                             true ->
+                         TooFar = case blockchain:config(?poc_distance_limit, Ledger) of
+                                      {ok, L} ->
+                                          D = blockchain_utils:distance(SourceLoc, DestinationLoc),
+                                          D > L;
+                                      _ ->
+                                          false
+                                  end,
```
mkaufmaner, line 1348: Is there an idea of what the `poc_distance_limit` will be? I am afraid this is a naive approach and that an arbitrary number that doesn't account for factors like topology, directional antennas, tropospheric ducting, propagation over water (https://explorer.helium.com/txns/slXAx6YcrKmrbzc4oozZ0-HQbR86gCoCICm8iGZMWxs) etc. With [HIP-15](https://github.com/helium/HIP/blob/master/0015-beaconing-rewards.md) rewarding a majority of the mined HNT onto the witnesses, transmitting at illegal power levels became pointless. In my opinion, this sets a dangerous precedent and MUST NOT be implemented.

vihu: > Is there an idea of what the poc_distance_limit will be?

Maybe some absurdly high value so it _always_ passes, idk like 500 Km or something? I understand your apprehension about this naive approach. We just added this as a placeholder really (it lacks validation). Also note that if we never set the variable, it has 0 effect.

mkaufmaner:
@vihu 

> Is there an idea of what the poc_distance_limit will be? I am afraid this is a naive approach and that an arbitrary number that doesn't account for factors like topology, directional antennas, propagation over water (https://explorer.helium.com/txns/slXAx6YcrKmrbzc4oozZ0-HQbR86gCoCICm8iGZMWxs) etc. With [HIP-15](https://github.com/helium/HIP/blob/master/0015-beaconing-rewards.md) rewarding a majority of the mined HNT onto the witnesses, transmitting at illegal power levels became pointless. In my opinion, this sets a dangerous precedent and MUST NOT be implemented.

> Is there an idea of what the poc_distance_limit will be?
> 
> Maybe some absurdly high value so it always passes, idk like 500 Km or something? I understand your apprehension about this naive approach. We just added this as a placeholder really (it lacks validation). Also note that if we never set the variable, it has 0 effect.

I apologize for deleting my previous comment, didn't realize you had already replied.  I believe there is already a blockchain variable handling this, `h3_max_grid_distance`.  Is this correct?  My concern is that they may be cases where long distances are attainable or the chain variable itself may be used for other purposes besides preventing abuse.

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
mkaufmaner: @vihu 

> Is there an idea of what the poc_distance_limit will be? I am afraid this is a naive approach and that an arbitrary number that doesn't account for factors like topology, directional antennas, propagation over water (https://explorer.helium.com/txns/slXAx6YcrKmrbzc4oozZ0-HQbR86gCoCICm8iGZMWxs) etc. With [HIP-15](https://github.com/helium/HIP/blob/master/0015-beaconing-rewards.md) rewarding a majority of the mined HNT onto the witnesses, transmitting at illegal power levels became pointless. In my opinion, this sets a dangerous precedent and MUST NOT be implemented.

> Is there an idea of what the poc_distance_limit will be?
> 
> Maybe some absurdly high value so it always passes, idk like 500 Km or something? I understand your apprehension about this naive approach. We just added this as a placeholder really (it lacks validation). Also note that if we never set the variable, it has 0 effect.

I apologize for deleting my previous comment, didn't realize you had already replied.  I believe there is already a blockchain variable handling this, `h3_max_grid_distance`.  Is this correct?  My concern is that they may be cases where long distances are attainable or the chain variable itself may be used for other purposes besides preventing abuse.

vihu:
> @vihu
> 
> > Is there an idea of what the poc_distance_limit will be? I am afraid this is a naive approach and that an arbitrary number that doesn't account for factors like topology, directional antennas, propagation over water (https://explorer.helium.com/txns/slXAx6YcrKmrbzc4oozZ0-HQbR86gCoCICm8iGZMWxs) etc. With [HIP-15](https://github.com/helium/HIP/blob/master/0015-beaconing-rewards.md) rewarding a majority of the mined HNT onto the witnesses, transmitting at illegal power levels became pointless. In my opinion, this sets a dangerous precedent and MUST NOT be implemented.
> 
> > Is there an idea of what the poc_distance_limit will be?
> > Maybe some absurdly high value so it always passes, idk like 500 Km or something? I understand your apprehension about this naive approach. We just added this as a placeholder really (it lacks validation). Also note that if we never set the variable, it has 0 effect.
> 
> I apologize for deleting my previous comment, didn't realize you had already replied. I believe there is already a blockchain variable handling this, `h3_max_grid_distance`. Is this correct? My concern is that they may be cases where long distances are attainable or the chain variable itself may be used for other purposes besides preventing abuse.

Well, `h3_max_grid_distance` was primarily used for poc-pathing which is now deprecated. We need a way to clean up old chain vars eventually (but definitely beyond the scope here).

As for long distances being attainable, I don't disagree at all. However, two things come into play, a) earth's curvature when doing line-of-sight transmissions (~160 km over sea or something, I need to check) and b) legally allowed antenna gains varying region-to-region. As I said earlier, just because we have the code does not imply that it's activated, it's just a mechanism IF in case we need to do it in the future. We prefer to pack things when working on a big feature as it's hard to consolidate a big release, then update something minor and do another release cycle.

src/transactions/v1/blockchain_txn_rewards_v1.erl
vihu: I reverted any changes to rewards_v1 here (they are of no consequence). And for rewards_v2 I just made it return `undefined` like all the other lookups and we match accordingly in `witness_decay/2` fun. 6a11c1c6ed354e2b46259e9d1913ce04db1b8d8e

rbrtio:
So the changes mentioned in https://github.com/helium/miner/issues/618 in regards to removing the SNR checks are accomplished by the changes in e76bc0f correct?  I am just trying to validate that this includes that change as I am being impacted by it on a regular basis as it appears are many others.

vihu:
> So the changes mentioned in [helium/miner#618](https://github.com/helium/miner/issues/618) in regards to removing the SNR checks are accomplished by the changes in [e76bc0f](https://github.com/helium/blockchain-core/commit/e76bc0ff4f66e51b13a54ff28b3c3ff8c746db68) correct? I am just trying to validate that this includes that change as I am being impacted by it on a regular basis as it appears are many others.

That commit sha is partially complete, `min_rcv_sig` as it stands [right now](https://github.com/helium/blockchain-core/blob/15754065d1e6fbb377731aa6f0f83a85944d0a27/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl#L1493) in the PR does the right things I believe. Note that it does _not_ include SNR checks there though just the FSPL calculation.

rbrtio:
> > So the changes mentioned in [helium/miner#618](https://github.com/helium/miner/issues/618) in regards to removing the SNR checks are accomplished by the changes in [e76bc0f](https://github.com/helium/blockchain-core/commit/e76bc0ff4f66e51b13a54ff28b3c3ff8c746db68) correct? I am just trying to validate that this includes that change as I am being impacted by it on a regular basis as it appears are many others.
> 
> That commit sha is partially complete, `min_rcv_sig` as it stands [right now](https://github.com/helium/blockchain-core/blob/15754065d1e6fbb377731aa6f0f83a85944d0a27/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl#L1493) in the PR does the right things I believe. Note that it does _not_ include SNR checks there though just the FSPL calculation.

Excellent, thank you for taking the time to reply.  I look forward to the changes.

src/region/blockchain_region_params_v1.erl
```diff
@@ -0,0 +1,72 @@
...
+-spec serialize(region_params_v1()) -> binary().
+serialize(#blockchain_region_params_v1_pb{} = RegionParams) ->
+    blockchain_region_param_v1_pb:encode_msg(RegionParams, blockchain_region_params_v1_pb).
+
+-spec deserialize(binary()) -> region_params_v1().
+deserialize(Bin) ->
+    blockchain_region_param_v1_pb:decode_msg(Bin, blockchain_region_params_v1_pb).
+
+-spec region_params(RegionParams :: region_params_v1()) -> [blockchain_region_param_v1:region_params_v1()].
+region_params(RegionParams) ->
```
Vagabond, line 71: If I call for_region, why do I have to call this on the result?

vihu: I added this here just to access the internal param list, can get rid of it I suppose. But then you'd have to     `RegionParams#blockchain_region_params_v1_pb.region_params` by hand to get to the param list

timcooijmans:
What will be the assumption about the transmitter-power given a provided antenna gain? 

Here in the EU, due to regulatory limits, it's quite common to reduce TX-power when a high(er)-gain antenna is connected. For example, when a device has it's TX-power tuned for a 3 dBi antenna, and a 6 dBi antenna is connected. We reduce the TX-power by 3 dB. However of course RX-gain is actually increased. 

This means that there will be a asymmetry between RX and TX range and RSSI. Can PoC v11 cope with this?

rbrtio:
@vihu Thank you for your continued work on this!  It looks like you are nearing the finish line.  :-)

vihu:
> @vihu Thank you for your continued work on this! It looks like you are nearing the finish line. :-)

Yeah, getting close :). There's still a bit of work on miner side https://github.com/helium/miner/pull/817

timcooijmans:
@vihu how would this handle asymmetric transmission/reception situations like we have here in Europe?

We commonly tune down TX-power to allow for higher-gain antennas to be used while still maintaining EIRP limits.  

Basically this means that a gateway transmits like a 3 dBi antenna but received as a 6 dBi antenna.

vihu:
> @vihu how would this handle asymmetric transmission/reception situations like we have here in Europe?
> 
> We commonly tune down TX-power to allow for higher-gain antennas to be used while still maintaining EIRP limits.
> 
> Basically this means that a gateway transmits like a 3 dBi antenna but received as a 6 dBi antenna.

Thanks for the comment @timcooijmans.

Please refer https://github.com/helium/proto/pull/41#issuecomment-789994410 for the precise reason why we are not supporting asymmetric gain situations. Atleast for the foreseeable future we'll be sticking with **one** gain which would have to be asserted by the hotspot on chain. All FSPL calculations will look that up going forward.

That said, if there are future HIPs on strengthening/reworking the POC algorithm which factors in asymmetric gain in a meaningful way, we will consider it.

timcooijmans:
Thanks for the clarification @vihu 

I'm afraid this would rule out a lot of operator-grade installations in Europe, that commonly use power-reduction.

Vagabond:
Transmission power reduction is allowed and considered. You should be fine.

src/region/blockchain_region_params_v1.erl
```diff
@@ -0,0 +1,141 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain Region Parameters ==
+%% @end
+%%%-------------------------------------------------------------------
+-module(blockchain_region_params_v1).
+
+%% TODO
+%% -behavior(blockchain_json).
```
Vagabond, line 9: TODO?

src/region/blockchain_region_params_v1.erl
```diff
@@ -0,0 +1,141 @@
...
+region_param(?region_cn470) -> ?region_params_cn470;
+region_param(?region_eu433) -> ?region_params_eu433;
+region_param(?region_eu868) -> ?region_params_eu868;
+region_param(?region_in865) -> ?region_params_in865;
+region_param(?region_kr920) -> ?region_params_kr920;
+region_param(?region_ru864) -> ?region_params_ru864;
+region_param(?region_us915) -> ?region_params_us915;
+
+%% NOTE: This _may_ be required for working with existing miner_lora and miner_onion_server
+region_param('AS923_1') -> ?region_params_as923_1;
```
Vagabond, line 80: I don't love these static values as it means we can't add new regions without a bunch of work. Maybe we should use list_to_atom carefully. Since the region names are in a chain var, this should be relatively safe?

src/region/blockchain_region_params_v1.erl
```diff
@@ -0,0 +1,141 @@
...
+region_param('AS923_2') -> ?region_params_as923_2;
```
Vagabond, line 81: When do we resolve regions with the ALL CAPS region name?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1269,40 +1287,88 @@ valid_witnesses(Element, Channel, Ledger) ->
...
+                        {error, _} ->
+                            %% If this errored out:
+                            %% - either the var is not set (improbable because we'll set it)
+                            %% - dst location is somehow not found in the regions we know about but SourceLoc was, sus
+                            false
+                    end;
+                {error, unknown_region} ->
+                    %% We don't know anything about this region
+                    %% Check if this ledger has aux
+                    case blockchain_ledger_v1:has_aux(Ledger) of
```
Vagabond, line 1322: Why do we do this?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1281,10 +1291,95 @@ validate_var(?penalty_history_limit, Value) ->
     %% also low end cannot be 0
     validate_int(Value, "penalty_history_limit", 10, 100000, false);
 
+validate_var(?regulatory_regions, Value) when is_binary(Value) ->
+    %% The regulatory_regions value we support must look like this:
+    %% <<"region_as923_1,region_as923_2,region_as923_3,region_as923_4,region_au915,region_cn470,region_eu433,region_eu868,region_in865,region_kr920,region_ru864,region_us915">>
+    %% The order does not matter in validation
+
+    %% First check is a relatively conservative byte_size check on the value
+    C1 = byte_size(Value) =< 148,
```
Vagabond, line 1300: Where does this number come from?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1281,10 +1291,95 @@ validate_var(?penalty_history_limit, Value) ->
...
+    %% Second check is that we're able to get the regions and it's not some random data
+    %% And it's atleast >= 11 (the number of regions we know about)
+    C2 = length(string:tokens(binary:bin_to_list(Value), ",")) >= 11,
```
Vagabond, line 1303: This seems like a bad idea if regions are removed over time (eg. AU915)

Vagabond: I'd say something like >= 3 or something would be fine, just to assert it is comma separated

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1281,10 +1291,95 @@ validate_var(?penalty_history_limit, Value) ->
...
+    validate_region_params(?region_params_kr920, Value);
+validate_var(?region_params_eu433, Value) ->
+    validate_region_params(?region_params_eu433, Value);
+
 validate_var(Var, Value) ->
     %% something we don't understand, crash
     invalid_var(Var, Value).
 
+validate_region_var(Var, Value) when is_binary(Value) ->
+    case size(Value) rem 8 of
```
Vagabond, line 1369:
```suggestion
    %% This is a list of u64 h3 hex ids, so it will always be a multiple of 8 bytes long
    case size(Value) rem 8 of
```

src/region/blockchain_region_params_v1.erl
vihu: Well, yes it's a bit of work to add new region, however, it's probably a good exercise to do so imho. For instance I added AS923-4 recently and it was...alright. Because the region comes from chain vars it should be relatively safe as well.

src/region/blockchain_region_params_v1.erl
vihu: Hmm, going to test without these fun heads. I remember that there was a period where miner_lora would be booting up and waiting for a chain and would end up calling this? (I could be misremembering and this may not be required at all)

vihu: Removed in the latest commit.

include/blockchain_vars.hrl
```diff
@@ -474,3 +483,27 @@
...
+%% regulatory_region related variables
+
+%% This is a comma separated string like so:
+%% <<"region_as923_1,region_as923_2,region_as923_3,region_as923_4,region_au915,region_cn470,region_eu433,region_eu868,region_in865,region_kr920,region_ru864,region_us915">>
+-define(regulatory_regions, regulatory_regions).
+
+%% Each of the former regions is associated with a dynamic var of the same name which is is a serialized form of an h3_region set determined at h3_res: 7
+%% For more information, check: https://github.com/JayKickliter/lorawan-h3-regions and https://github.com/helium/erlang-h3/pull/29
+
+%% Additionally, each of the former regions is associated with a dynamic var of the form <regionname>_params, eg region_params_us915 which
```
Vagabond, line 497:
```suggestion
%% Additionally, each of the former regions is associated with a dynamic var of the form <regionname>_params, eg region_us915_params which
```

src/region/blockchain_region_spreading_v1.erl
```diff
@@ -0,0 +1,70 @@
...
+-spec region_spreading(TaggedSpreading :: tagged_spreading()) -> undefined | atom().
+region_spreading(TaggedSpreading) ->
+    TaggedSpreading#tagged_spreading_pb.region_spreading.
+
+-spec select_spreading(
+    TaggedSpreading :: [tagged_spreading()],
+    PacketSize :: non_neg_integer()
+) -> {ok, atom()} | {error, any()}.
+select_spreading(TaggedSpreading, PacketSize) ->
+    %% FIXME: This is not quite right...
```
Vagabond, line 50: FIXME

src/region/blockchain_region_v1.erl
```diff
@@ -0,0 +1,96 @@
...
+%% == Blockchain Region API ==
+%% @end
+%%%-------------------------------------------------------------------
+-module(blockchain_region_v1).
+
+-include("blockchain_vars.hrl").
+
+-export([get_all_regions/1, h3_to_region/2, h3_in_region/3]).
+
+-type region_var() ::
```
Vagabond, line 12: Switch this to atom()

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -957,6 +957,16 @@ validate_var(?poc_per_hop_max_witnesses, Value) ->
     validate_int(Value, "poc_per_hop_max_witnesses", 5, 50, false);
 validate_var(?poc_addr_hash_byte_count, Value) ->
     validate_int(Value, "poc_addr_hash_byte_count", 4, 32, false);
+validate_var(?fspl_loss, Value) ->
+    validate_float(Value, "fspl_loss", 0.0, 1.0);
```
Vagabond, line 961: Upper limit is too low?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1418,6 +1493,96 @@ get_channels(Txn, Chain) ->
...
+                _ -> blockchain_utils:min_rcv_sig(FSPL, TxPower) * 1.0
+            end;
+        _ ->
+            blockchain_utils:min_rcv_sig(
+                blockchain_utils:free_space_path_loss(SourceLoc, DestinationLoc, Freq)
+            )
+    end.
+
+calc_fspl(SrcPubkeyBin, DstPubkeyBin, SourceLoc, DestinationLoc, Freq, Ledger) ->
+    %% XXX: Investigate why these blow up with aux ledger...
```
Vagabond, line 1542: XXX?

anthonyra:
While I was working on something I noticed that in `blockchain_ledger_gateway_v2` in the function `create_histogram` it use the `free_space_path_loss` and `min_rcv_sig` to generate the witness histogram. I understand the histogram is an artifact of time long gone but if it's still in the code should it be upgraded to allow the Gt and Gr that's being implemented in PoCv11? Or is it fine to approximate with just Frequency?

https://github.com/helium/blockchain-core/blob/cf5c2e57457ceeca81767a426063de33ab7e8629/src/ledger/v1/blockchain_ledger_gateway_v2.erl#L501

src/region/blockchain_region_spreading_v1.erl
vihu: I think the comment carried over before I fixed it to select based on max_packet_size. I'll remove it.

src/region/blockchain_region_v1.erl
vihu: Alright

src/transactions/v1/blockchain_txn_vars_v1.erl
vihu: Switching upper limit to 5.0

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: This does not happen anymore, will remove

include/blockchain_vars.hrl
```diff
@@ -474,3 +483,27 @@
...
+%% This is a comma separated string like so:
+%% <<"region_as923_1,region_as923_2,region_as923_3,region_as923_4,region_au915,region_cn470,region_eu433,region_eu868,region_in865,region_kr920,region_ru864,region_us915">>
+-define(regulatory_regions, regulatory_regions).
+
+%% Each of the former regions is associated with a dynamic var of the same name which is is a serialized form of an h3_region set determined at h3_res: 7
+%% For more information, check: https://github.com/JayKickliter/lorawan-h3-regions and https://github.com/helium/erlang-h3/pull/29
+
+%% Additionally, each of the former regions is associated with a dynamic var of the form <regionname>_params, eg region_us915_params which
+%% contains the RF parameters for that region.
+%% TODO: Figure out region_parameters for each of the supported regions
```
Vagabond, line 499:
```suggestion
```

src/region/blockchain_region_spreading_v1.erl
```diff
@@ -0,0 +1,70 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain Region Spreading ==
+%% @end
+%%%-------------------------------------------------------------------
+-module(blockchain_region_spreading_v1).
+
+%% TODO
```
Vagabond, line 8: TODO?

src/region/blockchain_region_spreading_v1.erl
```diff
@@ -0,0 +1,70 @@
...
+
+-spec select_spreading(
+    TaggedSpreading :: [tagged_spreading()],
+    PacketSize :: non_neg_integer()
+) -> {ok, atom()} | {error, any()}.
+select_spreading(TaggedSpreading, PacketSize) ->
+    FilterFun = fun(Tagged) ->
+        max_packet_size(Tagged) >= PacketSize
+    end,
+    case lists:filter(FilterFun, TaggedSpreading) of
```
Vagabond, line 53: This assumes the TaggedSpreading are sorted in order of ascending packet size?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1140,20 +1140,30 @@ valid_receipt(PreviousElement, Element, Channel, Ledger) ->
...
+            {ok, DestinationLoc} = blockchain_ledger_v1:find_gateway_location(DstPubkeyBin, Ledger),
             {ok, ExclusionCells} = blockchain_ledger_v1:config(?poc_v4_exclusion_cells, Ledger),
             {ok, ParentRes} = blockchain_ledger_v1:config(?poc_v4_parent_res, Ledger),
             SourceParentIndex = h3:parent(SourceLoc, ParentRes),
             DestinationParentIndex = h3:parent(DestinationLoc, ParentRes),
-            try h3:grid_distance(SourceParentIndex, DestinationParentIndex) >= ExclusionCells of
-                true ->
+            TooFar = case blockchain:config(?poc_distance_limit, Ledger) of
+                         {ok, L} ->
+                             D = blockchain_utils:distance(SourceLoc, DestinationLoc),
```
Vagabond, line 1153: This one won't crash like h3:grid_distance below?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1195,8 +1204,8 @@ valid_receipt(PreviousElement, Element, Channel, Ledger) ->
                                     Receipt
                             end
                     end;
-                false ->
-                    %% too close
+                _ ->
```
Vagabond, line 1207: Is it worth breaking out the 3 error cases here now there's more stuff that can go wrong?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1345,8 +1405,8 @@ tagged_witnesses(Element, Channel, Ledger) ->
                                                  [{true, <<"insufficient_data">>, Witness} | Acc]
                                          end
                                  end;
-                             false ->
-                                 %% too close
+                             _ ->
```
Vagabond, line 1408: *definitely* better error messages here

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1418,6 +1493,95 @@ get_channels(Txn, Chain) ->
...
 
+min_rcv_sig(undefined, Ledger, SrcPubkeyBin, SourceLoc, DstPubkeyBin, DestinationLoc, Freq) ->
+    %% Receipt can be undefined
+    case blockchain:config(?poc_version, Ledger) of
+        {ok, POCVersion} when POCVersion >= 11 ->
+            %% Estimate tx power because there is no receipt with attached tx_power
+            lager:debug("SourceLoc: ~p, Freq: ~p", [SourceLoc, Freq]),
+
+            case estimated_tx_power(SourceLoc, Freq, Ledger) of
+                {ok, not_required} ->
```
Vagabond, line 1504: Arguably we should fail the receipt here for more accurate aux ledger rewards?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1418,6 +1493,95 @@ get_channels(Txn, Chain) ->
...
+min_rcv_sig(Receipt, Ledger, SrcPubkeyBin, SourceLoc, DstPubkeyBin, DestinationLoc, Freq) ->
+    %% We do have a receipt
+    case blockchain:config(?poc_version, Ledger) of
+        {ok, POCVersion} when POCVersion >= 11 ->
+            %% Get tx_power from attached receipt and use it to calculate min_rcv_sig
+            TxPower = blockchain_poc_receipt_v1:tx_power(Receipt),
+            FSPL = calc_fspl(SrcPubkeyBin, DstPubkeyBin, SourceLoc, DestinationLoc, Freq, Ledger),
+            case blockchain:config(?fspl_loss, Ledger) of
+                {ok, Loss} -> blockchain_utils:min_rcv_sig(FSPL, TxPower) * Loss;
+                _ -> blockchain_utils:min_rcv_sig(FSPL, TxPower) * 1.0
```
Vagabond, line 1533: Multipltying by 1.0 seems redundant?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1418,6 +1493,95 @@ get_channels(Txn, Chain) ->
...
+                {ok, not_required} ->
+                    %% Just do the old behavior
+                    blockchain_utils:min_rcv_sig(
+                      blockchain_utils:free_space_path_loss(SourceLoc, DestinationLoc, Freq)
+                     );
+                {ok, TxPower} ->
+                    FSPL = calc_fspl(SrcPubkeyBin, DstPubkeyBin, SourceLoc, DestinationLoc, Freq, Ledger),
+                    case blockchain:config(?fspl_loss, Ledger) of
+                        {ok, Loss} -> blockchain_utils:min_rcv_sig(FSPL, TxPower) * Loss;
+                        _ -> blockchain_utils:min_rcv_sig(FSPL, TxPower) * 1.0
```
Vagabond, line 1513: another multiply by 1

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -1140,20 +1140,30 @@ valid_receipt(PreviousElement, Element, Channel, Ledger) ->
...
+                             false
+                     end,
+
+            SameRegion = is_same_region(Ledger, SourceLoc, DestinationLoc),
+
+            try h3:grid_distance(SourceParentIndex, DestinationParentIndex) of
+                Dist when Dist >= ExclusionCells andalso SameRegion andalso not TooFar ->
                     RSSI = blockchain_poc_receipt_v1:signal(Receipt),
                     SNR = blockchain_poc_receipt_v1:snr(Receipt),
                     Freq = blockchain_poc_receipt_v1:frequency(Receipt),
```
Vagabond, line 1165: We should check the frequency is correct for this region, right?

test/blockchain_region_SUITE.erl
```diff
@@ -0,0 +1,708 @@
...
+download_regions(RegionURLs) ->
+    blockchain_ct_utils:pmap(
+        fun({Region, URL}) ->
+            Ser = blockchain_ct_utils:download_serialized_region(URL),
+            {Region, Ser}
+        end,
+        RegionURLs
+    ).
+
+region_params_us915() ->
```
Vagabond, line 540: I *feel* like it'd be better to store these as the proto record and serialize them when inserting to the chain?

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: No this appears returns reasonable default values using vincenty distance

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Yes, I have it in https://github.com/helium/blockchain-core/pull/949 will be testing it shortly

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Addressed in #949

src/region/blockchain_region_spreading_v1.erl
vihu: Updated in #949 to sort the list on insertion

src/region/blockchain_region_spreading_v1.erl
vihu: Removed in #949

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Removed in #949

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Removed in #949

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
vihu: Added in https://github.com/helium/blockchain-core/pull/949

src/blockchain_utils.erl
```diff
@@ -296,7 +297,7 @@ free_space_path_loss(Loc1, Loc2, Frequency, Gt, Gl) ->
     (10*math:log10(math:pow((4*math:pi()*(Frequency*1000000)*(Distance*1000))/(299792458), 2)))-Gt-Gl.
 
 %% Subtract FSPL from our transmit power to get the expected minimum received signal.
--spec min_rcv_sig(float(), float()) -> float().
+-spec min_rcv_sig(float(), number()) -> float().
 min_rcv_sig(Fspl, TxGain) ->
```
anthonyra, line 301: A discussion on discord lead me to look into how EIRP and transmit power will be changing with v11 and just for clarity shouldn't the `-spec min_rcv_sig` take (Fspl, TxPower) to make the call and what it's doing since TxGain is different than TxPower? `blockchain_utils:min_rcv_sig(FSPL, TxPower)`

timcooijmans:
Following the discussion on Discord I want to voice my opinion here.

Yesterday I learned that the plan is to have the miner software adjust the requested tx power according to the antenna asserted. I think this is not the right direction and instead any antenna-gain compensation should be handled in the LoRa-packet-forwarder code as it’s usually done for other LoRa networks for several reasons:

- The power-field in the interface is always used to request the EIRP power, not the power excluding antenna at the connector. This change forces a breaking definition for this field making it non-compatible with any other LoRa network
- All (Helium) code related to transmission should take into account the antenna gain, instead of just handling it in the LoRa-packet-forwarder using an existing already used configuration option. 
- Our LongAP One has a 3 dBi stock antenna and currently when the miner requests a 14 dBm transmission we make sure 14 dBm EIRP is transmitted in accordance with ETSI standards applicable to CE certification. However due to bugs in the Helium app and old v1 assertions many of our hotspots still have the default 1.2 dBi antenna asserted. When we do nothing these hotspots will start transmitting at illegal levels due to this software change, this is unacceptable behaviour. We (legally) can’t rely on users to update their antennas (which is also a paid, but relatively cheap operation) or require them to do so. The only option for us is to modify the LoRa-packet-forwarder from the “stock” code to disallow transmissions above 12.2 dBm. This is a non-standard change and disrupts any possible transmission of a higher-power-level when allowed for the channel.

So our proposal is to expose the asserted antenna gain on the miner via CLI and/or JSON-RPC and let manufacturers handle the tx power adjustment in the LoRa packet forwarder like we already do for the region on a very similar fashion. This allows us to very easy catch any <3 dBi antennas by making sure the minimal allowed antenna gain is 3 dBi in the config of the LoRa packet forwarder and avoids us from having to inspect each and every transmitted packet.

Vagabond:
Ah, yes, there is a `antenna_gain` field in the packet forwarder.

Vagabond:
> > Ah, yes, there is a `antenna_gain` field in the packet forwarder.
> 
> Exactly, that gain simply gets subtracted from the transmission power requested. It's only a int but the transmission levels aren't more fine grained anyway.

Yeah, I think this is workable, so we can ask the packet forwarder for effective tx powers and simply assume the antenna gain is set correctly.

timcooijmans:
We had some analysis and discussions internally about this change. It seems that there's another error here, CE certification allows for 14 dBm ERP that's 16.15 dBm EIRP. So I think values should be changed if EIRP is used.

vihu:
> We had some analysis and discussions internally about this change. It seems that there's an error here, CE certification allows for 14 dBm ERP that's 16.15 dBm EIRP. So I think values should be changed if EIRP is used.

Which region specifically? I mostly picked the EIRP values from https://stackforce.github.io/LoRaMac-doc/LoRaMac-doc-v4.4.7/group___r_e_g_i_o_n_e_u868.html#ga66750d8b2f9ba9f6b8466f190207d53f

timcooijmans:
> > We had some analysis and discussions internally about this change. It seems that there's an error here, CE certification allows for 14 dBm ERP that's 16.15 dBm EIRP. So I think values should be changed if EIRP is used.
> 
> 
> 
> Which region specifically? I mostly picked the EIRP values from https://stackforce.github.io/LoRaMac-doc/LoRaMac-doc-v4.4.7/group___r_e_g_i_o_n_e_u868.html#ga66750d8b2f9ba9f6b8466f190207d53f

The max EIRP for EU868 is 16(.15) dBm, also see; https://github.com/Lora-net/LoRaMac-node/blob/05353ff70bec5349a26b23a02168accb0ea03ad9/src/mac/region/RegionEU868.h#L115

I haven't checked any other regions for correctness.

vihu:
> > > We had some analysis and discussions internally about this change. It seems that there's an error here, CE certification allows for 14 dBm ERP that's 16.15 dBm EIRP. So I think values should be changed if EIRP is used.
> > 
> > 
> > Which region specifically? I mostly picked the EIRP values from https://stackforce.github.io/LoRaMac-doc/LoRaMac-doc-v4.4.7/group___r_e_g_i_o_n_e_u868.html#ga66750d8b2f9ba9f6b8466f190207d53f
> 
> The max EIRP for EU868 is 16(.15) dBm, also see; https://github.com/Lora-net/LoRaMac-node/blob/05353ff70bec5349a26b23a02168accb0ea03ad9/src/mac/region/RegionEU868.h#L115
> 
> I haven't checked any other regions for correctness.

Ah yes, I see it. Thanks for catching it, I believe the current value was ported from miner's configuration (which set it to 14, arguably incorrectly). I'll fix and update it here.

Vagabond (approved):

PR #778, opened 2021-04-16
Wire up stake_release_height in txn and ledger

mrallen1:
Problem to solve: We want owners to explicitly agree to a defined stake release height when unstaking, since stakes in a cooldown state are unable to earn new HNT or transfer HNT.  This PR implements that.

Addresses helium/miner#736

See also helium/proto#74

### DIFF
--- a/src/transactions/v1/blockchain_txn_unstake_validator_v1.erl
+++ b/src/transactions/v1/blockchain_txn_unstake_validator_v1.erl
@@ -15,11 +15,12 @@
 -include_lib("helium_proto/include/blockchain_txn_unstake_validator_v1_pb.hrl").
 
 -export([
-         new/3,
+         new/4,
          hash/1,
          address/1,
          owner/1,
          stake_amount/1,
+         stake_release_height/1,
          owner_signature/1,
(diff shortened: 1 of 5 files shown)


evanmcc (approved):
update this for merged proto and then go ahead and merge it.  I'll handle the fixing the already broken miner tests.

src/transactions/v1/blockchain_txn_unstake_validator_v1.erl
```diff
@@ -144,6 +151,14 @@ is_valid(Txn, Chain) ->
                         case blockchain_ledger_validator_v1:owner_address(V) of
                             Owner -> ok;
                             _ -> throw(bad_owner)
+                        end,
+                        Cooldown = blockchain:config(?stake_withdrawl_cooldown, Ledger),
+                        CooldownMax = blockchain:config(?stake_withdrawl_max, Ledger),
```
abhay, line 156: Surprised a test didn't catch this but these return `{ok, V}` and not the bare value so they need to be wrapped in a case statement.

abhay: Looks like you caught this in #779

PR #781, opened 2021-04-19
fix erroneous log message when loading snapshot

BenoitDuffez:
`Curr2` is actually the height _before_ absorbing the snapshot.


vihu (approved):
Good catch, thanks!

PR #783, opened 2021-04-20
generalize historical penalties

evanmcc:
changes penalties to an on-ledger thing


Vagabond (approved):
I think this will help a lot as long as we can keep the phi tangent from tending towards infinity as the group order increases past the Pitman limit.

PR #786, opened 2021-04-22
Incrementalize snapshot (de)serialization to avoid memory spikes

xandkar:
TODO:
- [ ] merge of https://gitlab.com/barrel-db/erlang-rocksdb/-/merge_requests/142
- [ ] merge above into https://gitlab.com/Vagabond1/erlang-rocksdb


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -4080,26 +4083,17 @@ load_pocs(PoCs, Ledger) ->
...
-        CF,
-        fun(KV, Acc) ->
-                [KV | Acc]
-        end,
-        []
-       )).
+-spec snapshot_raw(rocksdb:cf_handle(), ledger()) -> [{binary(), binary()}].
+snapshot_raw(CF, L) ->
+    %% We can just reverse this since rocks folds are lexicographic.
+    %% TODO Why reverse at all? Why do we care about order?
```
evanmcc, line 4089: we care about order because these need to hash deterministically.  We don't need to reverse I guess if it's an efficiency concern, but it will lead to some broken snaps around the release of this version.

xandkar: Gotcha. Makes sense. Reversal; shouldn't use any extra memory, so for this purpose I guess it doesn't matter - will leave it be.

evanmcc (approved):

PR #789, opened 2021-04-30
Macpie/ch9765/router instruct hotspot over p2p to transmit

macpie:


macpie:
I will update here https://github.com/helium/blockchain-core/pull/794

PR #795, opened 2021-05-03
Make tenure penalty amt chain var

mrallen1:
Problem to solve: Previously, tenure penalties were hard coded to `1.0`. We may need to adjust them in the future.

Solution: Make the tenure penalty amount a chain var float


evanmcc (changes requested):
This looks great so far, but is missing a var for the election filter (the magic 5.0 above which we no longer consider things for election).  We'll need to hold off merging this until we're ready for a restart.

evanmcc (approved):

PR #797, opened 2021-05-04
Only do ledger_at if necessary

vihu:
When calculating rewards_v2, only invoke `ledger_at` if necessary. Expectation is that this would also fix aux ledger invocations.

TODO:
- [ ] Manual testing


vihu:
This is wrong.

PR #808, opened 2021-05-20
Implement blockchain GC for older snapshots and blocks

Vagabond:
Because disks are not infinite, we want to occasionally remove some
older blocks and snapshots so we don't fill the disk entirely. This
change adds an environment variable, BLOCKCHAIN_GC_BYTES that allows the
system to signal to the blockchain it needs to reclaim some space.

If this environment variable is set, the blockchain will kick off a
background process that, starting at the oldest blocks, will delete
blocks and snapshots at that height until it has deleted at least as
many bytes as was requested. Note that this will temporarily increase
disk usage because compaction will need to happen before rocksdb can
actually reclaim space.

When using this feature it's probably important that you remove *more*
than is your threshold for triggering GC. For example if you want to GC
when the disk is 80% full, you should probably tell the blockchain to
delete enough bytes to get you back to 75% so you have some hystersis.

The last 100k of blocks and snapshots are protected from GC.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -25,7 +25,7 @@
     find_first_block_after/2,
 
     add_blocks/2, add_blocks/3, add_block/2, add_block/3,
-    delete_block/2,
+    delete_block/2, gc/2,
     config/2,
     fees_since/2,
     build/3,
@@ -1942,6 +1942,41 @@ add_bin_snapshot(BinSnap, Height, Hash, #blockchain{db=DB, dir=Dir, snapshots=Sn
             {error, Why}
     end.
 
+gc(BytesToDrop, #blockchain{db=DB, heights=HeightsCF}=Blockchain) ->
+    {ok, Height} = blockchain:height(Blockchain),
+    CutoffHeight = max(2, Height - 100000),
+    %% start at 2 here so we don't GC the genesis block
+    {ok, Itr} = rocksdb:iterator(DB, HeightsCF, [{iterate_lower_bound, <<2:64/integer-unsigned-big>>}, {iterate_upper_bound, <<CutoffHeight:64/integer-unsigned-big>>}]),
+    gc(BytesToDrop, Itr, Blockchain,  rocksdb:iterator_move(Itr, first)).
+
+gc(_Bytes, _Itr, _Blockchain, {error, _}) ->
+    ok;
+gc(Bytes, _Itr, _Blockchain, _Res) when Bytes < 1 ->
+    ok;
+gc(Bytes, Itr, #blockchain{db=DB, heights=HeightsCF, blocks=BlocksCF, snapshots=SnapshotsCF}=Blockchain, {ok, <<IntHeight:64/integer-unsigned-big>>=Height, Hash}) ->
+    lager:info("GCing block at height ~p", [IntHeight]),
+    BytesRemoved0 = case rocksdb:get(DB, BlocksCF, Hash, []) of
+                        {ok, Block} -> byte_size(Block);
+                        _ -> 0
+                    end,
+    {ok, Batch} = rocksdb:batch(),
+    rocksdb:batch_delete(Batch, HeightsCF, Height),
+    rocksdb:batch_delete(Batch, BlocksCF, Hash),
+    BytesRemoved1 = case rocksdb:get(DB, SnapshotsCF, Height, []) of
+        {ok, SnapHash} ->
+            lager:info("GCing snapshot at height ~p", [IntHeight]),
+            rocksdb:batch_delete(Batch, SnapshotsCF, Height),
+            rocksdb:batch_delete(Batch, SnapshotsCF, SnapHash),
+            case rocksdb:get(DB, SnapshotsCF, SnapHash, []) of
+                {ok, Snap} -> byte_size(Snap);
+                _ -> 0
+            end;
+        _ ->
+            0
+    end,
+    ok = rocksdb:write_batch(DB, Batch, [{sync, true}]),
+    gc(Bytes - BytesRemoved0 - BytesRemoved1, Itr, Blockchain, rocksdb:iterator_move(Itr, next)).
+
 
 -spec get_snapshot(blockchain_block:hash() | integer(), blockchain()) ->
                           {ok, binary()} | {error, any()}.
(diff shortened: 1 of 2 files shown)


src/blockchain.erl
```diff
@@ -20,7 +20,7 @@
     find_first_block_after/2,
 
     add_blocks/2, add_block/2, add_block/3,
-    delete_block/2,
+    delete_block/2, gc/2,
```
mrallen1, line 23: I wonder if this function should be named `disk_gc` or `rocksdb_gc` just to make its intention more clear

src/blockchain.erl
```diff
@@ -1782,6 +1817,19 @@ load(Dir, Mode) ->
...
             },
+            %% check if we need to GC some blocks to keep disk space free
+            %% note that this will temporarily increase disk usage
+            case os:getenv("BLOCKCHAIN_GC_BYTES") of
+                false -> ok;
+                ValString ->
+                    try list_to_integer(ValString, 10) of
+                        BytesToGC ->
+                            lager:notice("System requested we free ~b bytes of disk space"),
+                            spawn(fun() -> gc(BytesToGC, Blockchain) end)
```
mrallen1, line 1828: If this spawn dies from an error do we want to monitor its exit and log it?

mrallen1 (commented):
Generally lgtm - couple of questions to consider

src/blockchain.erl
Vagabond: I'd be inclined to say it could retry on the next startup if the failure was transient. Permanant failure... well maybe it would be good to log?

src/blockchain.erl
```diff
@@ -1650,6 +1650,41 @@ add_bin_snapshot(BinSnap, Height, Hash, #blockchain{db=DB, snapshots=SnapshotsCF
             {error, Why}
     end.
 
+rocksdb_gc(BytesToDrop, #blockchain{db=DB, heights=HeightsCF}=Blockchain) ->
+    {ok, Height} = blockchain:height(Blockchain),
+    CutoffHeight = max(2, Height - 100000),
```
vihu, line 1655: Maybe put `100000` to util.hrl or chain.hrl

src/blockchain.erl
```diff
@@ -1650,6 +1650,41 @@ add_bin_snapshot(BinSnap, Height, Hash, #blockchain{db=DB, snapshots=SnapshotsCF
...
+    BytesRemoved0 = case rocksdb:get(DB, BlocksCF, Hash, []) of
+                        {ok, Block} -> byte_size(Block);
+                        _ -> 0
+                    end,
+    {ok, Batch} = rocksdb:batch(),
+    rocksdb:batch_delete(Batch, HeightsCF, Height),
+    rocksdb:batch_delete(Batch, BlocksCF, Hash),
+    BytesRemoved1 = case rocksdb:get(DB, SnapshotsCF, Height, []) of
+        {ok, SnapHash} ->
+            lager:info("GCing snapshot at height ~p", [IntHeight]),
```
vihu, line 1675:
```suggestion
            lager:debug("GCing snapshot at height ~p", [IntHeight]),
```

shawaj:
@vihu @mrallen1 @evanmcc @Vagabond is there any progress with this at all?

Vagabond:
> @vihu @mrallen1 @evanmcc @Vagabond is there any progress with this at all?

I think we just need to address the review comments, sorry it fell through the cracks.

src/blockchain.erl
Vagabond: I'd rather log this explicitly

Vagabond:
Ok, this should have been modernized and had all the review comments handled now.

jadeallenx (approved):

src/blockchain.erl
```diff
@@ -2119,6 +2170,21 @@ load(Dir, Mode) ->
...
                 ledger=Ledger
             },
+            %% check if we need to GC some blocks to keep disk space free
+            %% note that this will temporarily increase disk usage
+            case os:getenv("BLOCKCHAIN_ROCKSDB_GC_BYTES") of
+                false -> ok;
+                ValString ->
+                    try list_to_integer(ValString, 10) of
+                        BytesToGC ->
+                            lager:notice("System requested we free ~b bytes of disk space"),
```
Vagabond, line 2180:
```suggestion
                            lager:notice("System requested we free ~b bytes of disk space", [BytesToGC]),
```

src/blockchain.erl
```diff
@@ -1942,6 +1944,55 @@ add_bin_snapshot(BinSnap, Height, Hash, #blockchain{db=DB, dir=Dir, snapshots=Sn
...
+    %% start at 2 here so we don't GC the genesis block
+    {ok, Itr} = rocksdb:iterator(DB, HeightsCF, [{iterate_lower_bound, <<2:64/integer-unsigned-big>>}, {iterate_upper_bound, <<CutoffHeight:64/integer-unsigned-big>>}]),
+    do_rocksdb_gc(BytesToDrop, Itr, Blockchain,  rocksdb:iterator_move(Itr, first)).
+
+do_rocksdb_gc(_Bytes, _Itr, _Blockchain, {error, _}) ->
+    ok;
+do_rocksdb_gc(Bytes, _Itr, _Blockchain, _Res) when Bytes < 1 ->
+    ok;
+do_rocksdb_gc(Bytes, Itr, #blockchain{dir=Dir, db=DB, heights=HeightsCF, blocks=BlocksCF, snapshots=SnapshotsCF}=Blockchain, {ok, <<IntHeight:64/integer-unsigned-big>>=Height, Hash}) ->
+    lager:info("GCing block at height ~p", [IntHeight]),
```
Vagabond, line 1959:
```suggestion
    lager:info("GCing block at height ~p, ~b bytes remain", [IntHeight, Bytes]),
```

evanmcc (approved):

shawaj:
@Vagabond just a quick question...

It says here that the default is to retain 100k blocks.

Is there a minimum we should observe for this?

abhay:
> It says here that the default is to retain 100k blocks. Is this supposed to actually say 10k as per:

Yes it should be 10k

> https://github.com/helium/blockchain-core/blob/3cd6bca6c5595a1363a9bbd625ef254383a4141b/src/blockchain.erl#L1988
> 
> Also, is there a minimum we should observe for this?

You can go fairly low here on a Hotspot. They only need about 2K blocks to be safe (even less since all a Hotspot needs, today, is to look back at active PoCs and State Channels.

shawaj:
Ah ok we were thinking 5k or so, so that's great, thanks @abhay 

One last question - is it safe/ok to just set the env variable for `BLOCKCHAIN_ROCKSDB_GC_BYTES` permanently to like 8GB or something like that and let it constantly prune down the storage? 

Or does it need to be set and removed each time you want to do a prune?

abhay:
> One last question - is it safe/ok to just set the env variable for `BLOCKCHAIN_ROCKSDB_GC_BYTES` permanently to like 8GB or something like that and let it constantly prune down the storage? 
> 
> 
> 
> Or does it need to be set and removed each time you want to do a prune?

We only run it when we see we're about to hit 70% used on the OG hotspots when we start the miner.

@Vagabond may have more context here on why.

PR #811, opened 2021-05-23
Fix snapshot grab CLI

vihu:


vihu:
I'll post a different PR shortly which _only_ updates the spec

PR #821, opened 2021-05-25
Make handle purchase async & do not use sync=true for rocksdb

macpie:


macpie:
We are making the calls async for a few reasons: 
- the `call` to `cast`: as we handle the purchase the sc handler is getting locked until sc client answers which is bringing us to our second issue.
- We send the packet ([here](https://github.com/helium/blockchain-core/blob/1e680e7fc00dedb663b5029d24dccba17b86eeb7/src/state_channel/blockchain_state_channels_client.erl#L442)) and then write to db right after but issue is that if that write takes too long (calls to rocks being  `sync=true` can be sometimes very long, I have seen up to 12s...) the `gen_server:call` is also blocked which blocks the sc handler and its ability to receive the packet mentioned before.

Therefore, the solution is to make sure that no calls to rocks are synchronous and that we do not block the handler just in case.

Note: it should not mess up message order as they are queued by streams.

Vagabond (approved):

PR #823, opened 2021-05-26
add some eqc tests to the guts of the election mechanism

evanmcc:
- [x] sizes checking
- [ ] validator filtering
- [x] deduplication
- [ ] adjustment if possible
- [ ] transition if possible


Vagabond:
Can you add some notes on what the remaining tasks entail? Maybe I can take a tilt at those windmills.

evanmcc:
the two eqc tests here hit the deselection code and the resize quantity code.

other things we could look at are: generating quality and then running some statistical test against the generated new group to see if it "looks" like what we'd expect, testing new node selection directly for the same effect, adding some more properties to the removal test to test the quality of the non-offline nodes, etc.

Vagabond (approved):

PR #838, opened 2021-06-01
[DNM] Assert res 9 testing

vihu:
Testing what it will take to switch all existing gateways and subsequently added gateways to have gateway location asserted at h3 resolution 9.

Notes:

- Add `switch_assert_res_9` chain var
- Add `blockchain_change_assert_res_SUITE`
- This depends on var hook support
- Sync is broken, obviously
- From limited testing, it appears that wherever `h3:parent(SomeLoc, Res)` gets called and Res > 9, :boom: 
- So at the very minimum we need to set the following variables, although I don't yet fully know if there are others which could be borked.

```
    #{
        ?switch_assert_res_9 => true,
        ?min_assert_h3_res => 9,
        ?poc_v4_parent_res => 8
    }.
```

TODO:

- [ ] Resyncing
- [ ] Model POC behavior change
- [ ] Model reward behavior change

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -456,3 +456,9 @@
 -define(penalty_history_limit, penalty_history_limit). % blocks
 -define(dkg_penalty, dkg_penalty). % float
 -define(tenure_penalty, tenure_penalty). % float
+
+
+%%%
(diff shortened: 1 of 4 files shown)


vihu:
Stale, closing.

PR #840, opened 2021-06-02
Add fee_payer/2 callback

madninja:
This determines the fee_payer given a txn and a potentially needed ledger for context.

### DIFF
--- a/src/transactions/v1/blockchain_txn_add_gateway_v1.erl
+++ b/src/transactions/v1/blockchain_txn_add_gateway_v1.erl
@@ -25,6 +25,7 @@
     payer_signature/1,
     staking_fee/1, staking_fee/2,
     fee/1, fee/2,
+    fee_payer/2,
     sign/2,
     sign_request/2,
     sign_payer/2,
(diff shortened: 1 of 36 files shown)


src/transactions/v1/blockchain_txn_add_gateway_v1.erl
```diff
@@ -355,13 +364,9 @@ absorb(Txn, Chain) ->
     AreFeesEnabled = blockchain_ledger_v1:txn_fees_active(Ledger),
     Owner = ?MODULE:owner(Txn),
     Gateway = ?MODULE:gateway(Txn),
-    Payer = ?MODULE:payer(Txn),
+    ActualPayer = ?MODULE:fee_payer(Txn, Ledger),
```
mrallen1, line 367: When you have code like this, do you want to _also_ put it in the `print` and `to_json` function o/p?

src/transactions/v1/blockchain_txn_stake_validator_v1.erl
```diff
@@ -79,6 +80,10 @@ stake(Txn) ->
 fee(Txn) ->
     Txn#blockchain_txn_stake_validator_v1_pb.fee.
 
+-spec fee_payer(txn_stake_validator(), blockchain_ledger_v1:ledger()) -> libp2p_crypto:pubkey_bin() | undefined.
```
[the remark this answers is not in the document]
madninja: I was just matching the callback spec

vihu (approved):

src/transactions/v1/blockchain_txn_add_gateway_v1.erl
madninja: huh.. you mean like adjust _all_ the to_json function to add a fee_payer field if the payer is actually used? I'd not considered that, but it would seem like it'd be null/absent for most of the transactions?

mrallen1: I mean it's fine if you want to omit it - just curious if we should surface it in print o/p or json o/p? I guess it is undefined most of the time tho

PR #841, opened 2021-06-02
Store snapshots we downloaded from S3, handle snapshot initialization better

mrallen1:
Problem to solve: We don't store S3 downloads and we don't check if we already have a snapshot before attempting to pull another one. This can lead to a lot of S3 downloads.

Solution: Store S3 snapshot downloads and if we have a snapshot at start up, directly use it

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -2129,8 +2129,6 @@ init_blessed_snapshot(Blockchain, _HashAndHeight={Hash, Height0}) when is_binary
                   lager:info("Got snapshot for height ~p - attempting install", [Height0]),
                   blockchain_worker:install_snapshot(Hash, Snap);
                {error, not_found} ->
-                  lager:warning("snapshot at height: ~p not found - attempt to sync",
-                                [Height0, Hash]),
                   blockchain_worker:snapshot_sync(Hash, Height);
                Other ->
                   lager:error("Got ~p trying to get snapshot at height: ~p hash ~p - attempt to sync",
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -1646,8 +1646,12 @@ add_snapshot(Snapshot, #blockchain{db=DB, snapshots=SnapshotsCF}) ->
             {error, Why}
     end.
 
--spec add_bin_snapshot(blockchain_ledger_snapshot:snapshot(), integer(), binary(), blockchain()) ->
+-spec add_bin_snapshot(blockchain_ledger_snapshot:snapshot(), integer(),
+                       none | {some, binary()} |binary(), blockchain()) ->
                               ok | {error, any()}.
+add_bin_snapshot(_BinSnap, _Height, none, _Chain) -> {error, no_snapshot_hash};
+add_bin_snapshot(BinSnap, Height, {some, Hash}, Chain) ->
+   add_bin_snapshot(BinSnap, Height, Hash, Chain);
 add_bin_snapshot(BinSnap, Height, Hash, #blockchain{db=DB, snapshots=SnapshotsCF}) ->
     try
         {ok, Batch} = rocksdb:batch(),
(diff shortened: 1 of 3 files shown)


src/blockchain.erl
```diff
@@ -2114,13 +2114,21 @@ init_blessed_snapshot(Blockchain, _HashAndHeight={Hash, Height0}) when is_binary
...
         {ok, CurrHeight} ->
-            lager:info("ch ~p h ~p: snap sync", [CurrHeight, Height]),
-            blockchain_worker:snapshot_sync(Hash, Height),
+            lager:debug("ch ~p h ~p: snap sync", [CurrHeight, Height]),
+            case get_snapshot(Hash, Blockchain) of
+               {ok, Snap} ->
+                  lager:info("Got snapshot for height ~p - attempting install", [Height0]),
+                  blockchain_worker:install_snapshot(Hash, Snap);
+               Other ->
+                  lager:warning("Got ~p trying to get snapshot at height: ~p hash ~p - attempt to sync",
```
evanmcc, line 2128: maybe just log something if this isn't `{error, not_found}`?

src/blockchain_worker.erl
```diff
@@ -917,11 +917,28 @@ start_snapshot_sync(Hash, Height, Swarm, Chain, Peer) ->
                                       %% adjusted either +1 or -1
                                       {ok, ConfigHeight} = application:get_env(blockchain,
                                                                 blessed_snapshot_block_height),
-                                      Url = build_url(BaseUrl, ConfigHeight),
-                                      %% this return snapshot is already deserialized
-                                      {ok, Snap} = attempt_fetch_s3_snapshot(Url),
-                                      lager:info("Successfully downloaded snap from ~p", [Url]),
-                                      blockchain_worker:install_snapshot(Hash, Snap)
+                                      case blockchain:have_snapshot(ConfigHeight, Chain) of
```
evanmcc, line 920: maybe we should just revert this?  there are two issues with it:
1. we shouldn't get here nearly as often with the change at startup
2. we don't do anything with the information that we have the snap and we fall back to getting it over p2p

So, I think that we can either revert this code, or handle the new error by catching it specifically and then installing the snapshot from there.

mrallen1: Yeah, agree it needs to be modified - hopefully this flow makes more sense

src/blockchain.erl
```diff
@@ -2114,13 +2114,25 @@ init_blessed_snapshot(Blockchain, _HashAndHeight={Hash, Height0}) when is_binary
...
         {ok, CurrHeight} ->
-            lager:info("ch ~p h ~p: snap sync", [CurrHeight, Height]),
-            blockchain_worker:snapshot_sync(Hash, Height),
+            lager:debug("ch ~p h ~p: snap sync", [CurrHeight, Height]),
+            case get_snapshot(Hash, Blockchain) of
+               {ok, Snap} ->
+                  lager:info("Got snapshot for height ~p - attempting install", [Height0]),
+                  blockchain_worker:install_snapshot(Hash, Snap);
+               {error, not_found} ->
+                  lager:warning("snapshot at height: ~p not found - attempt to sync",
```
evanmcc, line 2128: this is the common case, not worth logging, imo

mrallen1: ok

evanmcc (approved):

src/blockchain_worker.erl
```diff
@@ -918,10 +918,24 @@ start_snapshot_sync(Hash, Height, Swarm, Chain, Peer) ->
...
                                                                 blessed_snapshot_block_height),
                                       Url = build_url(BaseUrl, ConfigHeight),
-                                      %% this return snapshot is already deserialized
-                                      {ok, Snap} = attempt_fetch_s3_snapshot(Url),
+                                      {ok, BinSnap} = attempt_fetch_s3_snapshot(Url),
                                       lager:info("Successfully downloaded snap from ~p", [Url]),
-                                      blockchain_worker:install_snapshot(Hash, Snap)
+                                      {ok, Snap} = blockchain_ledger_snapshot_v1:deserialize(BinSnap),
+                                      %% assert we got hash we expected before we store it
+                                      case blockchain_ledger_snapshot_v1:hash(Snap) of
```
vihu, line 925: nit: worth moving this case statement to `maybe_install_snapshot` or something similarly named?

mrallen1: Yeah that's a good idea, but since we're checking SHAs at deserialize, if the hash doesn't match what we downloaded, we'll crash out of this anyway before we hit the store and install thing

evanmcc (approved):

PR #843, opened 2021-06-03
Update condition for next plausible block

vihu:
Problem: `is_block_plausible` will incorrectly return false when the next block can indeed be considered plausible, for example getting a block after an election happens. This happens because of a sublist + shuffle when invoking `verify_signatures` for a block, and is exacerbated by the fact that `verify_signatures` terminates early on a single signature verification failure.

Solution:

- Expose `verified_signees/1` from block, this just accumulates those members who verifiably signed the block
- Change plausibility condition to a sets:size + intersection call
- Add test to verify said behavior
- Also added `is_election_block` optional callback (intended to use it for something, but decided against it, lmk if it needs to go away)

### DIFF
--- a/src/blockchain_block_v1.erl
+++ b/src/blockchain_block_v1.erl
@@ -31,10 +31,11 @@
     snapshot_hash/1,
     verify_signatures/4, verify_signatures/5,
     is_rescue_block/1,
-    to_json/2
+    is_election_block/1,
+    to_json/2,
+    verified_signees/1
 ]).
 
--include("blockchain.hrl").
 -include_lib("helium_proto/include/blockchain_block_v1_pb.hrl").
 
 -ifdef(TEST).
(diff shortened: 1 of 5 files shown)


Vagabond:
I believe because we only look for f+1 signatures from the current group even election blocks should be plausible here without this patch.

vihu:
> I believe because we only look for f+1 signatures from the current group even election blocks should be plausible here without this patch.

Ah right, that makes sense. I'll close this unless we want `is_election_block` optional callback only from here.

src/blockchain_block_v1.erl
```diff
@@ -347,6 +348,14 @@ is_rescue_block(Block) ->
         (Block#blockchain_block_v1_pb.rescue_signature /= <<>> orelse
          Block#blockchain_block_v1_pb.rescue_signatures /= []).
 
+-spec is_election_block(block()) -> boolean().
+is_election_block(Block) ->
```
evanmcc, line 352: see blockchain_election for a similar function

vihu: Ah I see, could call `has_new_group/1` from here and continue exposing this or rip this one out altogether.

evanmcc (approved):

PR #846, opened 2021-06-03
readd snapshot back compat

evanmcc:
this re-adds the ability to load older snapshots, hopefully in a cheaper way than before, and changes the deserialize API around a little bit.


src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -510,7 +560,20 @@ height(#{current_height := H}) ->
     H.
 
 -spec hash(snapshot()) -> binary().
-hash(#{version := v6}=Snap) ->
+hash(Snap) ->
+    hash(version(Snap), Snap).
+
+hash(V, Snap) when V == 1 orelse V == v1 ->
```
xandkar, line 566: Which code would call this with the atom version of version?

xandkar (approved):

PR #852, opened 2021-06-04
Fix state channel server to switch state channel when it becomes too big

macpie:


macpie:
Fix for https://github.com/helium/router/issues/335

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -255,13 +255,42 @@ handle_cast({offer, SCOffer, HandlerPid},
...
+                                                                  SignedPurchaseSC,
+                                                                  Skewed),
+                            NewState =
+                                maybe_add_stream(Hotspot,
+                                                 HandlerPid,
+                                                 State#state{state_channels=maps:put(ActiveSCID,
+                                                                                     {SignedPurchaseSC, Skewed},
+                                                                                     SCs)}),
+                            {noreply, NewState};
+                        error ->
```
vihu, line 280: Can you please tag this error with a msg?

macpie: 👍🏻

macpie:
@vihu I have added a test, was useful as I found out that `blockchain_state_channel_v1:update_summary_for/4` was kind of broken

test/blockchain_state_channel_SUITE.erl
```diff
@@ -468,6 +471,108 @@ expired_test(Config) ->
...
+            Offer0 = blockchain_state_channel_offer_v1:from_packet(Packet, PubKeyBin, 'US915'),
+            Offer1 = blockchain_state_channel_offer_v1:sign(Offer0, SigFun),
+            ok = ct_rpc:call(RouterNode, gen_server, cast, [blockchain_state_channels_server, {offer, Offer1, Self}])
+        end,
+        lists:seq(1, ?MAX_UNIQ_CLIENTS + 10)
+    ),
+
+    %% Checking that new SC ID is not old SC ID
+    ok = blockchain_ct_utils:wait_until(fun() ->
+        State1 = ct_rpc:call(RouterNode, sys, get_state, [blockchain_state_channels_server]),
```
vihu, line 570: Same as above

test/blockchain_state_channel_SUITE.erl
```diff
@@ -468,6 +471,108 @@ expired_test(Config) ->
...
+                    915.2,
+                    "SF8BW125",
+                    -12.0,
+                    {devaddr, 12}
+                ),
+            Offer0 = blockchain_state_channel_offer_v1:from_packet(Packet, PubKeyBin, 'US915'),
+            Offer1 = blockchain_state_channel_offer_v1:sign(Offer0, SigFun),
+            ok = ct_rpc:call(RouterNode, gen_server, cast, [blockchain_state_channels_server, {offer, Offer1, Self}])
+        end,
+        lists:seq(1, ?MAX_UNIQ_CLIENTS + 10)
```
vihu, line 565: Does it have to be 10? what happens if it goes 1 up? It would still switch correct?

macpie: yes 10 was random want me to switch to 1?

test/blockchain_state_channel_SUITE.erl
```diff
@@ -468,6 +471,106 @@ expired_test(Config) ->
...
+            Offer0 = blockchain_state_channel_offer_v1:from_packet(Packet, PubKeyBin, 'US915'),
+            Offer1 = blockchain_state_channel_offer_v1:sign(Offer0, SigFun),
+            ok = ct_rpc:call(RouterNode, gen_server, cast, [blockchain_state_channels_server, {offer, Offer1, Self}])
+        end,
+        lists:seq(1, ?MAX_UNIQ_CLIENTS + 1)
+    ),
+
+    %% Checking that new SC ID is not old SC ID
+    ok = blockchain_ct_utils:wait_until(fun() ->
+        ActiveSCID0 =/= ct_rpc:call(RouterNode, blockchain_state_channels_server, active_sc_id, [])
```
michaeldjeffrey, line 569: Is there any use in checking here that the `active_sc_id` is not `undefined`?

vihu (approved):
lgtm

test/blockchain_state_channel_SUITE.erl
macpie: we dont have another SC opened here so it will go to `undefined` but if there was another one it should switch

macpie:
This si not ready and needs to be reworked to spill over instead of just switching SC

src/state_channel/blockchain_state_channel_handler.erl
```diff
@@ -109,21 +109,22 @@ init(server, _Conn, [_Path, Blockchain]) ->
...
-                undefined ->
+            case blockchain_state_channels_server:active_scs() of
+                [] ->
                     %% Send empty banner
                     SCBanner = blockchain_state_channel_banner_v1:new(),
                     lager:info("sc_handler, empty banner: ~p", [SCBanner]),
                     {ok, #state{ledger=Ledger, handler_mod=HandlerMod, pending_offer_limit=OfferLimit},
                      blockchain_state_channel_message_v1:encode(SCBanner)};
-                ActiveSC ->
+                [ActiveSC|_] ->
```
vihu, line 119: Is the first one the real active one always?

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -357,55 +287,153 @@ terminate(_Reason, _State) ->
...
+    {ok, blockchain_state_channel_v1:state_channel()} | {error, any()}.
+select_best_active_sc(PubKeyBin, #state{active_sc_ids=ActiveSCIDs, state_channels=SCs}) ->
+    select_best_active_sc(PubKeyBin, ActiveSCIDs, SCs).
+
+-spec select_best_active_sc(libp2p_crypto:pubkey_bin(), [blockchain_state_channel_v1:id()], state_channels()) ->
+    {ok, blockchain_state_channel_v1:state_channel()} | {error, any()}.
+select_best_active_sc(_PubKeyBin, [], _SCs) ->
+    {error, no_active_state_channels};
+select_best_active_sc(PubKeyBin, [ActiveSCID|Others], StateChannels) ->
+    {SC, _} = maps:get(ActiveSCID, StateChannels),
```
vihu, line 301: This will crash if ActiveSCID is not in StateChannels, it that's intentional plz add a comment here

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -357,55 +287,153 @@ terminate(_Reason, _State) ->
...
-process_packet(ClientPubkeyBin, Packet, SC, Skewed, HandlerPid,
-               #state{db=DB, sc_version=SCVer, active_sc_id=ActiveSCID, state_channels=SCs,
-                      blooms=Blooms, owner={_, OwnerSigFun}}=State) ->
-
+handle_packet(ClientPubKeyBin, Packet, HandlerPid,
+              #state{db=DB, sc_version=SCVer,state_channels=SCs,
+                     blooms=Blooms, owner={_, OwnerSigFun}}=State) ->
     Payload = blockchain_helium_packet_v1:payload(Packet),
-
+    {ok, SC} = select_best_active_sc(ClientPubKeyBin, State),
```
vihu, line 317: This may return `{error, no_active_state_channels}`

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -357,55 +287,153 @@ terminate(_Reason, _State) ->
...
+        false ->
+            lager:error("payload size (~p) exceeds maximum (~p). Sending rejection of offer ~p from ~p",
+                        [PayloadSize, ?MAX_PAYLOAD_SIZE, SCOffer, HandlerPid]),
+            ok = send_rejection(HandlerPid),
+            {noreply, State0};
+        true ->
+            Hotspot = blockchain_state_channel_offer_v1:hotspot(SCOffer),
+            case select_best_active_sc(Hotspot, State0) of
+                {error, _Reason} ->
+                    lager:info("could not get a good active SC (~p)", [_Reason]),
```
vihu, line 360: may want to change it to debug/warning

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -357,55 +287,153 @@ terminate(_Reason, _State) ->
...
+                    ActiveSCID = blockchain_state_channel_v1:id(ActiveSC),
+                    lager:debug("selected state channel ~p", [libp2p_crypto:bin_to_b58(ActiveSCID)]),
+                    NumDCs = blockchain_utils:do_calculate_dc_amount(PayloadSize, State0#state.dc_payload_size),
+                    TotalDCs = blockchain_state_channel_v1:total_dcs(ActiveSC),
+                    DCAmount = blockchain_state_channel_v1:amount(ActiveSC),
+                    PreventOverSpend = application:get_env(blockchain, prevent_sc_overspend, true),
+                    case (TotalDCs + NumDCs) > DCAmount andalso PreventOverSpend of
+                        true ->
+                            % QUESTION
+                            % Do we need to sign and save on rejection?
```
vihu, line 380: Don't think so, atleast not in this PR. If there's a reason to, we can do so later.

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -128,15 +129,15 @@ update_summary_for(ClientPubkeyBin,
...
-    case get_summary(ClientPubkeyBin, SC) of
-        {error, not_found} ->
-            {SC#blockchain_state_channel_v1_pb{summaries=[NewSummary | Summaries]}, true};
-        {ok, _Summary} ->
-            case WillFit orelse can_fit(ClientPubkeyBin, Summaries) of
-                false ->
-                    %% Cannot fit this into summaries
-                    {SC, false};
-                true ->
+    case WillFit orelse ?MODULE:can_fit(ClientPubkeyBin, SC) of
```
vihu, line 132: Might be better to handle these individually?

```
case {WillFit, Canfit} of
   {true, true} -> %% yes
   _ -> %% no
end
```

xandkar: Even as a pattern-matching maximalist, I'd say this one is better as an `orelse` that it already is.

vihu: Heh, fair enough. I was also wondering from a functional standpoint, i.e. if will_fit=true, can_fit=false should it try and update summary? If not, then maybe it should be an `andalso` instead of `orelse`?

src/state_channel/blockchain_state_channel_handler.erl
Vagabond: Don't we need to check which one we can banner from, if we're spilling?

macpie: The first one should be the oldest one as I always add to the back of the list

macpie: We should send banner for all of them I guess?

src/state_channel/v1/blockchain_state_channel_v1.erl
macpie: Will fit at that point should already be exact if I am not mistaken this is why I went with orelse

src/state_channel/blockchain_state_channels_server.erl
macpie: yes this is "bad" so I want to crash here

include/blockchain_utils.hrl
```diff
@@ -48,7 +48,7 @@
                      -13 => {-125,-125},
                      11 => {-90,-35}}).
 
--define(MAX_UNIQ_CLIENTS, 1100).
+-define(MAX_UNIQ_CLIENTS, 2000).
```
macpie, line 51: @Vagabond is that ok?

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -55,7 +55,8 @@
     streams = #{} :: streams(),
     dc_payload_size :: undefined | pos_integer(),
     sc_version = 0 :: non_neg_integer(), %% defaulting to 0 instead of undefined
-    blooms = #{} :: blooms()
+    blooms = #{} :: blooms(),
+    max_actors_allowed = 1100 :: non_neg_integer()
```
vihu, line 59:
```suggestion
    max_actors_allowed = 1100 :: pos_integer()
```

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -516,6 +522,20 @@ can_fit(ClientPubkeyBin, Summaries) ->
             true
     end.
 
+-spec max_actors_allowed(Ledger :: blockchain_ledger_v1:ledger()) -> non_neg_integer().
```
vihu, line 525:
```suggestion
-spec max_actors_allowed(Ledger :: blockchain_ledger_v1:ledger()) -> pos_integer().
```

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -472,38 +477,39 @@ quick_compare_causality(OlderSC, CurrentSC, PubkeyBin) ->
     end.
 
 -spec merge(SCA :: state_channel(),
-            SCB :: state_channel()) -> state_channel().
-merge(SCA, SCB) ->
+            SCB :: state_channel(),
+            MaxActorsAllowed :: non_neg_integer()) -> state_channel().
```
vihu, line 481:
```suggestion
            MaxActorsAllowed :: pos_integer()) -> state_channel().
```

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -123,20 +126,22 @@ summaries(Summaries, SC) ->
 -spec update_summary_for(ClientPubkeyBin :: libp2p_crypto:pubkey_bin(),
                          NewSummary :: blockchain_state_channel_summary_v1:summary(),
                          SC :: state_channel(),
-                         WillFit :: boolean()) -> {state_channel(), boolean()}.
+                         WillFit :: boolean(),
+                         MaxActorsAllowed :: non_neg_integer()) -> {state_channel(), boolean()}.
```
vihu, line 130:
```suggestion
                         MaxActorsAllowed :: pos_integer()) -> {state_channel(), boolean()}.
```

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -357,55 +289,174 @@ terminate(_Reason, _State) ->
...
+                                    SC3
+                            end,
 
-    {ClientBloom, PacketBloom} = maps:get(ActiveSCID, Blooms),
+                    SignedSC = blockchain_state_channel_v1:sign(NewSC, OwnerSigFun),
+                    %% save it
+                    ok = blockchain_state_channel_v1:save(DB, SignedSC, Skewed1),
+                    %% Put new state_channel in our map
+                    TempState = State#state{state_channels=maps:update(ActiveSCID, {SignedSC, Skewed1}, SCs)},
+                    maybe_add_stream(ClientPubKeyBin, HandlerPid, TempState)
```
vihu, line 334: There should also be a `maybe_broadcast_banner` here since the active SC has changed I think

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -357,55 +289,174 @@ terminate(_Reason, _State) ->
...
+                {error, _Reason} ->
+                    lager:debug("could not get a good active SC (~p)", [_Reason]),
+                    case maybe_get_new_active(State0) of
+                        undefined ->
+                            lager:warning("could not get a new active SC (~p), rejecting", [_Reason]),
+                            ok = send_rejection(HandlerPid),
+                            State0;
+                        NewActiveID ->
+                            lager:info("adding new active SC ~p", [libp2p_crypto:bin_to_b58(NewActiveID)]),
+                            handle_offer(SCOffer, HandlerPid, State0#state{active_sc_ids=ActiveSCIDs ++ [NewActiveID]})
```
vihu, line 360: Hmm, this seems like infinite retries for handling offer? Should there be limited number of retries or are we expecting that a previous condition fails and a rejection gets sent?

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -357,55 +289,174 @@ terminate(_Reason, _State) ->
...
+                    PreventOverSpend = application:get_env(blockchain, prevent_sc_overspend, true),
+                    case (TotalDCs + NumDCs) > DCAmount andalso PreventOverSpend of
+                        true ->
+                            lager:warning("dropping this packet because it will overspend DC ~p, (cost: ~p, total_dcs: ~p)",
+                                          [DCAmount, NumDCs, TotalDCs]),
+                            lager:warning("overspend, SC1: ~p", [ActiveSC]),
+                            ok = send_rejection(HandlerPid),
+                            case maybe_get_new_active([ActiveSCID], State0) of
+                                undefined ->
+                                    State0#state{active_sc_ids=lists:delete(ActiveSCID, ActiveSCIDs)};
```
vihu, line 377: I believe `state_channels` map in the state also needs to be updated here?

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -357,55 +289,174 @@ terminate(_Reason, _State) ->
...
+                                NewActiveSCID ->
+                                    State1 = State0#state{active_sc_ids=lists:delete(ActiveSCID, ActiveSCIDs) ++ [NewActiveSCID]},
```
vihu, line 379: Same, update `state_channels` map also?

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -357,55 +289,174 @@ terminate(_Reason, _State) ->
...
+                                    ok = maybe_broadcast_banner(active_scs(State1), State1),
```
vihu, line 380: I _think_ this call should happen outside this case statement

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -465,54 +514,38 @@ update_state_sc_open(Txn,
...
-                           active_sc_id=NewActiveSCID},
-
-    case NewActiveSCID /= ActiveSCID of
+update_state_sc_close(Txn, #state{db=DB, cf=SCF, blooms=Blooms, state_channels=SCs0, active_sc_ids=ActiveSCIDs0}=State0) ->
+    ClosedSC = blockchain_txn_state_channel_close_v1:state_channel(Txn),
+    ClosedID = blockchain_state_channel_v1:id(ClosedSC),
+    ok = delete_closed_sc(DB, SCF, ClosedID),
+    ActiveSCIDs1 = lists:delete(ClosedID, ActiveSCIDs0),
+    SCs1 = maps:remove(ClosedID, SCs0),
+    NewActiveSCIDs =
```
vihu, line 526: Move `526 - 537` to a helper fun? `select_new_active_sc_ids` or something?

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -472,38 +477,39 @@ quick_compare_causality(OlderSC, CurrentSC, PubkeyBin) ->
...
 
-
-%% ------------------------------------------------------------------
-%% Internal Function Definitions
-%% ------------------------------------------------------------------
 -spec can_fit(ClientPubkeyBin :: libp2p_crypto:pubkey_bin(),
-              Summaries :: summaries()) -> boolean().
-can_fit(ClientPubkeyBin, Summaries) ->
+              SC :: state_channel(),
+              Max :: non_neg_integer()) -> boolean().
```
vihu, line 508:
```suggestion
              Max :: pos_integer()) -> boolean().
```

vihu:
While we're here maybe we should update `validate_int(Value, "max_open_sc", 1, 10, false);` to `validate_int(Value, "max_open_sc", 1, 100, false);` or something as well to allow us more headroom in the future

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -148,122 +149,49 @@ init(Args) ->
...
-    {noreply, NewState};
-handle_cast({offer, SCOffer, HandlerPid}, #state{active_sc_id=undefined}=State) ->
-    erlang:spawn(
-        fun() ->
-            lager:warning("Got offer: ~p when no sc is active", [SCOffer]),
-            %% Reject any offer if we don't have an active_sc as well, as a courtesy for the router
-            ok = send_rejection(HandlerPid)
-        end),
+handle_cast({packet, ClientPubKeyBin, Packet, HandlerPid}, State0) ->
+    State1 = handle_packet(ClientPubKeyBin, Packet, HandlerPid, State0),
```
michaeldjeffrey, line 185: Debug message here? Since we were previously seeing offers but no packets.

src/state_channel/blockchain_state_channels_server.erl
macpie: if we got a new active it should then succeed otherwise the `maybe_get_new_active` will fail

src/state_channel/blockchain_state_channels_server.erl
macpie: It was not previously, I think the idea here is that the sc is not considered active anymore and will get remove fully when the close txn comes in

src/state_channel/blockchain_state_channels_server.erl
macpie: see above

src/state_channel/blockchain_state_channels_server.erl
macpie: 👍🏻

src/state_channel/blockchain_state_channels_server.erl
vihu: Ah, I suppose this should be ok then as is

src/state_channel/blockchain_state_channels_server.erl
macpie: 👍🏻

src/state_channel/blockchain_state_channels_server.erl
macpie: It's only used there tho

src/state_channel/blockchain_state_channel_handler.erl
Vagabond: no, banner for the one that we'd spill this peer to

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1056,6 +1056,9 @@ validate_var(?sc_causality_fix, Value) ->
     validate_int(Value, "sc_causality_fix", 1, 1, false);
 validate_var(?sc_gc_interval, Value) ->
     validate_int(Value, "sc_gc_interval", 10, 100, false);
+validate_var(?sc_max_actors, Value) ->
+    %% Minimum is the old MACRO value
```
Vagabond, line 1060: I would use a lower minimum here, like 500

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -516,6 +522,20 @@ can_fit(ClientPubkeyBin, Summaries) ->
...
     end.
 
+-spec max_actors_allowed(Ledger :: blockchain_ledger_v1:ledger()) -> pos_integer().
+max_actors_allowed(Ledger) ->
+    case blockchain_ledger_v1:config(?sc_max_actors, Ledger) of
+        {ok, I} ->
+            I;
+        _ ->
+            %% 1100 was the previously hardcoded value
+            1100
```
Vagabond, line 532: I would have left this as the macro defined before.

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -36,26 +36,27 @@
...
     owner = undefined :: {libp2p_crypto:pubkey_bin(), libp2p_crypto:sig_fun()} | undefined,
     state_channels = #{} :: state_channels(),
-    active_sc_id = undefined :: undefined | blockchain_state_channel_v1:id(),
+    active_sc_ids = [] :: [blockchain_state_channel_v1:id()],
     streams = #{} :: streams(),
     dc_payload_size :: undefined | pos_integer(),
     sc_version = 0 :: non_neg_integer(), %% defaulting to 0 instead of undefined
-    blooms = #{} :: blooms()
+    blooms = #{} :: blooms(),
+    max_actors_allowed = ?SC_MAX_ACTORS :: pos_integer()
```
Vagabond, line 59: Caching this is a bit risky unless you're watching for the chain var to change somehow?

macpie: I watch the chain so I ll update on add block just in case

vihu (approved):

src/state_channel/v1/blockchain_state_channel_v1.erl
```diff
@@ -516,6 +522,17 @@ can_fit(ClientPubkeyBin, Summaries) ->
             true
     end.
 
+-spec max_actors_allowed(Ledger :: blockchain_ledger_v1:ledger()) -> pos_integer().
+max_actors_allowed(Ledger) ->
```
vihu, line 526: I think this function belongs in the ledger tbh

vihu: Meh, maybe it's fine

src/state_channel/blockchain_state_channel_handler.erl
macpie: Fixed

src/state_channel/v1/blockchain_state_channel_v1.erl
macpie: Done

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -4976,6 +4977,32 @@ get_sc_mod(Channel, Ledger) ->
         _ -> blockchain_ledger_state_channel_v1
     end.
 
+    %%--------------------------------------------------------------------
```
vihu, line 4980: nit: indentation

src/state_channel/blockchain_state_channel_handler.erl
```diff
@@ -103,25 +103,52 @@ send_response(Pid, Resp) ->
...
+                                    SCBanner = blockchain_state_channel_banner_v1:new(),
+                                    lager:info("sending empty banner", []),
+                                    {ok, #state{ledger=Ledger, handler_mod=HandlerMod, pending_offer_limit=OfferLimit},
+                                    blockchain_state_channel_message_v1:encode(SCBanner)};
+                                {ok, ActiveSC} ->
+                                    lager:debug("found sc ~p for hotspot ~p, sending banner",
+                                                [blockchain_utils:addr2name(blockchain_state_channel_v1:id(ActiveSC)),
+                                                 blockchain_utils:addr2name(PubKeyBin)]),
+                                    SCBanner = blockchain_state_channel_banner_v1:new(ActiveSC),
+                                    {ok,#state{ledger=Ledger, handler_mod=HandlerMod, pending_offer_limit=OfferLimit},
```
vihu, line 148: not a huge deal but looks like the return val is the same for all the cases here? maybe we should pull it out further? just remove some dupes, your call

Vagabond (approved):

PR #856, opened 2021-06-07
Make taking ledger checkpoints more atomic

Vagabond:


src/blockchain.erl
```diff
@@ -526,7 +526,6 @@ fold_blocks(Chain0, DelayedHeight, DelayedLedger, Height, ForceRecalc) ->
                                       %% make things faster in the future
                                       Ledger1 = ?MODULE:ledger(Chain1),
                                       {ok, NewLedger} = blockchain_ledger_v1:context_snapshot(Ledger1),
-                                      delayed = blockchain_ledger_v1:mode(NewLedger),
```
Vagabond, line 529: I'm not sure this is wise to remove, can we ever get a non delayed ledger here??

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -671,6 +671,8 @@ new_snapshot(#ledger_v1{db=DB,
                     ets:delete(Cache, DeleteHeight),
                     1 = ets:select_replace(Cache, [{Old, [], [{const, {Height, {snapshot, SnapshotHandle}}}]}]),
                     %% take a checkpoint as well for use after a restart
+                    %% This is treated at atomic and there are no further updates required, unlike
```
evanmcc, line 674:
```suggestion
                    %% This is treated as atomic and there are no further updates required, unlike
```

src/blockchain.erl
evanmcc: I think this is ok?  if it's possible it's a very rare path.   I think that the delayed ledger stuff is also slated for demolition now that we have checkpoints, so hopefully this won't matter for very long.

evanmcc (approved):

vihu (approved):

PR #857, opened 2021-06-08
Fix some bugs with creating atomic checkpoints

Vagabond:


evanmcc (approved):
I don't understand this super well, but it seems cleaner and safer.

vihu (approved):

PR #860, opened 2021-06-09
Dont download S3 snap if honor_quick_sync is false

vihu:
Problem: miner will ignore user specified `honor_quick_sync` app:env and subsequent snapshot load command and try to re-ingest a snapshot from S3

Solution: Honor `honor_quick_sync` as false when trying to load snapshot.

TODO: I'm not sure if this is _all_ that's needed

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -912,9 +912,15 @@ grab_snapshot(Height, Hash) ->
 start_snapshot_sync(Hash, Height, Swarm, Chain, Peer) ->
     spawn_monitor(fun() ->
                           try
-                              case application:get_env(blockchain, s3_base_url, undefined) of
-                                  undefined -> throw({error, no_s3_base_url});
-                                  BaseUrl ->
+                              BaseUrl = application:get_env(blockchain, s3_base_url, undefined),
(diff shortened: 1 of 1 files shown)


mrallen1 (approved):
:shipit:

src/blockchain_worker.erl
```diff
@@ -912,9 +912,15 @@ grab_snapshot(Height, Hash) ->
...
                           try
-                              case application:get_env(blockchain, s3_base_url, undefined) of
-                                  undefined -> throw({error, no_s3_base_url});
-                                  BaseUrl ->
+                              BaseUrl = application:get_env(blockchain, s3_base_url, undefined),
+                              HonorQS = application:get_env(blockchain, honor_quick_sync, true),
+
+                              case {HonorQS, BaseUrl} of
+                                  {_, undefined} -> throw({error, no_s3_base_url});
+                                  {false, _} ->
```
evanmcc, line 920: these are ordered incorrectly, honor_quick_sync = false but no s3 base url will still try to load a snap

vihu: Good catch, I updated slightly 18c7399

mrallen1 (approved):

PR #864, opened 2021-06-14
Fix deadlock when continuing to load snapshot

Vagabond:
Additionally we avoid re-loading the ledger if we detect an interrupted
snapshot load.

Finally, we try to increase the file size for column families storing
blocks so that we do compactions less often because we are storing large
values.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -1669,7 +1669,7 @@ add_bin_snapshot(BinSnap, Height, Hash, #blockchain{db=DB, snapshots=SnapshotsCF
 
 
 -spec get_snapshot(blockchain_block:hash() | integer(), blockchain()) ->
-                          {ok, blockchain_ledger_snapshot:snapshot()} | {error, any()}.
+                          {ok, binary()} | {error, any()}.
 get_snapshot(<<Hash/binary>>, #blockchain{db=DB, snapshots=SnapshotsCF}) ->
     case rocksdb:get(DB, SnapshotsCF, Hash, []) of
         {ok, <<"__sentinel__">>} ->
@@ -1786,16 +1786,11 @@ load(Dir, Mode) ->
             Ledger =
                 case Mode of
                     blessed_snapshot when HonorQuickSync == true ->
-                        %% use 1 as a noop, but this combo is poorly defined
-                        Height = application:get_env(blockchain, blessed_snapshot_block_height, 1),
                         L = blockchain_ledger_v1:new(Dir, DB, BlocksCF, HeightsCF),
                         case blockchain_ledger_v1:current_height(L) of
-                            {ok, 1} ->
-                                L;
-                            {ok, CHt} when CHt < Height ->
-                                blockchain_ledger_v1:clean(L),
-                                blockchain_ledger_v1:new(Dir, DB, BlocksCF, HeightsCF);
                             {ok, _} ->
+                                %% no longer check height here, we will check the height elsewhere
+                                %% where we have better information
                                 L;
                             %% if we can't open the ledger and we can
                             %% load a snapshot, does the error matter?
@@ -1985,7 +1980,7 @@ open_db(Dir) ->
         end,
 
     CFOpts = GlobalOpts,
-    case rocksdb:open_with_cf(DBDir, DBOptions,  [{CF, CFOpts} || CF <- ExistingCFs]) of
+    case rocksdb:open_with_cf(DBDir, DBOptions,  [adjust_cfopts({CF, CFOpts}) || CF <- ExistingCFs]) of
         {error, _Reason}=Error ->
             Error;
         {ok, DB, OpenedCFs} ->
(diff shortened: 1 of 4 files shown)


src/blockchain.erl
```diff
@@ -2128,10 +2129,41 @@ init_blessed_snapshot(Blockchain, _HashAndHeight={Hash, Height0}) when is_binary
...
+                          {ok, LaggingHeight} = blockchain_ledger_v1:current_height(blockchain_ledger_v1:mode(delayed, OldLedger)),
+                          {ok, LeadingHeight} = blockchain_ledger_v1:current_height(OldLedger),
+                          SnapHeight = blockchain_ledger_snapshot_v1:height(Snap),
+                          lager:info("Snapshot height ~p, lagging ledger height ~p, leading ledger height ~p", [SnapHeight, LaggingHeight, LeadingHeight]),
+                          case LaggingHeight == SnapHeight andalso LeadingHeight >= LaggingHeight of
+                              true when HasAux == false ->
+                                  lager:info("resuming interrupted snapshot block load"),
+                                  %% ledger height is correct, but blockchain height is not
+                                  %% so lets just try to resume loading blocks as that was likely
+                                  %% interripted
```
evanmcc, line 2149:
```suggestion
                                  %% interrupted
```

src/blockchain.erl
```diff
@@ -2002,6 +1997,12 @@ open_db(Dir) ->
     end.
 
 
+adjust_cfopts({CF, CFOpts}) when CF == "blocks"; CF == "temp_blocks"; CF == "plausible_blocks" ->
+    %% use 8mb file sizes to avoid too many compactions when storing blocks
+    {CF, lists:keystore(target_file_size_base, 1, CFOpts, {target_file_size_base, 8388608})};
```
evanmcc, line 2002: maybe we just want this as a global option? regardless, this is good

src/blockchain.erl
```diff
@@ -2128,10 +2129,41 @@ init_blessed_snapshot(Blockchain, _HashAndHeight={Hash, Height0}) when is_binary
...
+                                  %% so lets just try to resume loading blocks as that was likely
+                                  %% interripted
+                                  %%
+                                  %% snapshots are loaded atomically into the ledger, so if we have
+                                  %% the right height we can be quite confident it was an interrupted
+                                  %% block load and bypass re-creating the ledger
+                                  blockchain_ledger_snapshot_v1:load_blocks(OldLedger, Blockchain, Snap),
+                                  Blockchain;
+                              _ ->
+                                  blockchain_ledger_v1:clean(OldLedger),
```
evanmcc, line 2157: if `HasAux` is `true`, do we need to clean the aux here too?

evanmcc (approved):

src/blockchain.erl
```diff
@@ -2128,10 +2129,41 @@ init_blessed_snapshot(Blockchain, _HashAndHeight={Hash, Height0}) when is_binary
...
+                          lager:info("Got snapshot for height ~p - attempting install", [Height0]),
+                          %% do the install in-line here vs making a blocking call to
+                          %% blockchain_worker
+                          OldLedger = blockchain:ledger(Blockchain),
+                          HasAux = blockchain_ledger_v1:has_aux(OldLedger),
+                          {ok, LaggingHeight} = blockchain_ledger_v1:current_height(blockchain_ledger_v1:mode(delayed, OldLedger)),
+                          {ok, LeadingHeight} = blockchain_ledger_v1:current_height(OldLedger),
+                          SnapHeight = blockchain_ledger_snapshot_v1:height(Snap),
+                          lager:info("Snapshot height ~p, lagging ledger height ~p, leading ledger height ~p", [SnapHeight, LaggingHeight, LeadingHeight]),
+                          case LaggingHeight == SnapHeight andalso LeadingHeight >= LaggingHeight of
```
vihu, line 2144: Why do we check `LaggingHeight == SnapHeight` here, is it because we're continuing the snap load or...?

vihu (approved):

src/blockchain.erl
Vagabond: I believe that's handled? The reason I check for it was it's a bit complicated to recover the snapshot load with an aux ledger, so I decided to let it just blow it away and start fresh.

src/blockchain.erl
Vagabond: yes, the snapshot's ledger height is actually the height of the lagging ledger, so if we have a lagging ledger at the snapshot height, but the chain is not up to the *other* snapshot height, we know that block loading got interrupted, and we can continue it.

src/blockchain_worker.erl
```diff
@@ -1104,8 +1104,7 @@ get_sync_mode(Blockchain) ->
                             {snapshot, {Hash, Height}};
                         _Chain ->
                             {ok, CurrHeight} = blockchain:height(Blockchain),
-                            %% this should be height - 1, but it crashes badly if you set it
```
Vagabond, line 1107: Someone should figure out why this was bad, and if it's better now?

evanmcc: fairly sure that it's safe to remove this comment as this PR should be addressing the issues I was seeing here.

PR #866, opened 2021-06-14
misc txn fixes

andymck:
Addresses two issues

1. Replaces the "not_found" response from blockchain_ledger_v1:find_gateway/2 when a gateway entry is not found with the more descriptive msg "address_entry_not_found".   The updated response is then handled within blockchain_ledger_v1:check_dc_or_hnt_balance/4 and translated into a meaningful error msg which will be propagated to txn_mgr and from there out to ETL via the txn callback mechanism.

2.  Handles the scenario of a gateway not being found as part of assert location v2's is_new_location/2 function.  Prevents this crashing out and polluting ETL logs

### DIFF
--- a/src/transactions/v2/blockchain_txn_assert_location_v2.erl
+++ b/src/transactions/v2/blockchain_txn_assert_location_v2.erl
@@ -343,9 +343,14 @@ do_is_valid_checks(Txn, Chain) ->
 is_new_location(Txn, Ledger) ->
     GwPubkeyBin = ?MODULE:gateway(Txn),
     NewLoc = ?MODULE:location(Txn),
-    {ok, Gw} = blockchain_ledger_v1:find_gateway_info(GwPubkeyBin, Ledger),
-    ExistingLoc = blockchain_ledger_gateway_v2:location(Gw),
-    NewLoc /= ExistingLoc.
+    case blockchain_ledger_v1:find_gateway_info(GwPubkeyBin, Ledger) of
+        {ok, Gw} ->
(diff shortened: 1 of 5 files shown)


src/transactions/v2/blockchain_txn_assert_location_v2.erl
```diff
@@ -343,9 +343,14 @@ do_is_valid_checks(Txn, Chain) ->
...
-    {ok, Gw} = blockchain_ledger_v1:find_gateway_info(GwPubkeyBin, Ledger),
-    ExistingLoc = blockchain_ledger_gateway_v2:location(Gw),
-    NewLoc /= ExistingLoc.
+    case blockchain_ledger_v1:find_gateway_info(GwPubkeyBin, Ledger) of
+        {ok, Gw} ->
+            ExistingLoc = blockchain_ledger_gateway_v2:location(Gw),
+            NewLoc /= ExistingLoc;
+        {error, _Reason} ->
+            %% if GW doesnt exist, default to true
+            true
```
vihu, line 352:
```suggestion
            throw({error, gateway_not_found})
```

vihu: Just crash this here and have the catch all in blockchain_txn pick it up

madninja: or throw the same error that do_remaining_checks returns when it can't find the gateway

vihu (approved):

PR #867, opened 2021-06-14
Store state channels in protobuf format

JayKickliter:
This PR modified the way we store state channels in RocksDB in an effort to reduce disk space.

- store SCs as protobuf serialized binaries
- handle older database entries serialized with `term_to_binary`

## Remaining work

I am not familiar with this codebase and there are quite a few un-abstracted RocksDB gets/puts involving state channels, so I'm leaving this as draft PR until at least I figure out:

- Is there a danger of a get/put in another module expecting the older `erlang_to_binary` format when in actuality it is a protobuf binary?
- How to test?


evanmcc:
closing to clean the list, reopen if we need to move on this

PR #873, opened 2021-06-21
Add accessor to remove var txns from block

xandkar:
Needed for var autoskip in https://github.com/helium/miner/pull/840

### DIFF
--- a/src/blockchain_block_v1.erl
+++ b/src/blockchain_block_v1.erl
@@ -33,7 +33,8 @@
     is_rescue_block/1,
     is_election_block/1,
     to_json/2,
-    verified_signees/1
+    verified_signees/1,
+    remove_var_txns/1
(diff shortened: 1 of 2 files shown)


src/blockchain_block_v1.erl
```diff
@@ -59,6 +60,12 @@
 
 -export_type([block/0, block_map/0]).
 
+-spec remove_var_txns(block()) -> block().
+remove_var_txns(#blockchain_block_v1_pb{transactions=Txns0}=Block) ->
```
vihu, line 64: nit: worth having a little eunit for this?

xandkar: agreed

evanmcc (approved):

evanmcc (approved):

PR #876, opened 2021-06-22
Ensure that condense_votes properly accounts for all nodes

Vagabond:


Vagabond:
This is not useful as we only look for explicit *false* values.

PR #877, opened 2021-06-22
Don't run ledger upgrades before checking if we need a snapshot sync

Vagabond:
Ledger upgrades can be expensive, so expensive in fact that they can OOM
older hardware. This change defers the ledger upgrade check until after
we've decided we don't need to do a snapshot sync so we don't try to do
a doomed or irrelevant upgrade before downloading the snapshot we need.


vihu (approved):

mrallen1 (approved):
Yeah should help bootstrap a chain before we do any upgrades (which may be in the snapshot)

PR #880, opened 2021-06-24
Only count SC active if it's open

mrallen1:
Problem to solve: Currently the state channels server returns the literal length of the Active ID list regardless of the state of the underlying state channel itself. This causes the router to come under state channel pressure because channels often hit a limit and are closed before a new one can be opened and garbage collected.

Solution: We will check the open or closed state of a state channel, and only count channels that are in the `open` state.


Vagabond (commented):
I think this is correct, but refactoring AND changing the logic makes it a bit trickier to review.

Is there a test we can do to compare the behaviour with/without this var?

macpie:
Love this PR would be great improvement!

> I think this is correct, but refactoring AND changing the logic makes it a bit trickier to review.
> 
> Is there a test we can do to compare the behaviour with/without this var?

Yes I like the refactoring but harder to read, are current test passing?

vihu:
@mrallen1 I have rebased and reverted the last commit here as we discussed internally. The changes to validation composability far too many and we'd address those separately when we can.

Vagabond:
There was a change to is_valid in the reverted commit

vihu:
> There was a change to is_valid in the reverted commit

Right, yes. I updated validation [here](https://github.com/helium/blockchain-core/pull/880/commits/b9e4bd2feda6c6b484f7a0b1925376b8796f421a). I _think_ this is the behavior we want, would be good to double-check.

cc @mrallen1

mrallen1:
@vihu This looks correct to me - thanks

macpie (approved):

Vagabond (approved):

Vagabond (approved):

PR #882, opened 2021-06-25
Flatten deep validation

xandkar:
2 initial versions to choose from in the review.


xandkar:
Closing, because bitrot and this was just an experiment anyhow. Similar improvements will be bundled in the 4th, final step of transaction-validation-breakup series of PRs.

PR #887, opened 2021-06-29
Make rewards v2 great again

jadeallenx:
Problem to solve: Rewards v2 was great for a while, but now we need to make it more memory efficient since network growth and state channel utilization continues to increase. This increase in memory use while computing rewards can lead to OOM and other problems on lightly provisioned gateway equipment.

Solution: The main approach is to discard every possible intermediate data structure (for example, we immediately attempt to resolve a gateway/validator to its owner) before we can generate the actual transaction list. Future work may include periodic rewards transactions (say we generate rewards every N blocks instead of only after an epoch ends, for example) and/or rewards which are segmented into smaller chunks instead of one giant transaction every epoch.

### DIFF
--- a/src/transactions/v2/blockchain_txn_rewards_v2.erl
+++ b/src/transactions/v2/blockchain_txn_rewards_v2.erl
@@ -441,14 +441,19 @@ finalize_reward_calculations(#{shares_acc := Shares} = RewardsMD, Ledger, Vars) 
                               dc_reward => DCReward},
                         [poc_challengers, poc_challengees, poc_witnesses]),
 
+    normalize_all_reward_shares(RewardsMD1, Vars0).
 
+-spec normalize_all_reward_shares(
+        RewardsMD :: rewards_metadata(),
+        Vars :: reward_vars() ) -> NewRewardsMD :: rewards_metadata().
+normalize_all_reward_shares(#{ shares_acc := Shares} = RewardsMD, Vars) ->
     %% we need to map over gateway owners and then fold over the list
     %% of rewards shares and turn those shares into HNT payouts
     maps:map(fun(shares_acc, V) -> V;
                 (_K, #rewards_meta{ shares = S } = RMD) ->
                       HNT = lists:foldl(fun(#rewards_shares{type = T,
                                                             shares = A}, Acc) ->
-                                                Acc + normalize_shares(T, A, Shares, Vars0)
+                                                Acc + normalize_shares(T, A, Shares, Vars)
                                         end,
                                         0, S),
                       update_hnt(RMD, HNT)
@@ -791,9 +796,9 @@ new_rewards_meta(Type, Shares) ->
 update_rewards_shares(Type, Shares, SharesList) ->
     case lists:keyfind(Type, #rewards_shares.type, SharesList) of
         false -> [ new_rewards_share(Type, Shares) ];
-        #rewards_shares{ shares = Current } = Rec ->
+        #rewards_shares{} = Rec ->
             lists:keyreplace(Type, #rewards_shares.type, SharesList,
-                             Rec#rewards_shares{ shares = Current + Shares })
+                             Rec#rewards_shares{ shares = Shares })
     end.
 
 new_rewards_share(Type, Shares) -> #rewards_shares{ type = Type, shares = Shares }.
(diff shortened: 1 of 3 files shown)


src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -724,61 +694,104 @@ securities_rewards(Ledger, #{epoch_reward := EpochReward,
...
+    #rewards_meta{ shares = [ new_rewards_share(Type, Shares) ] }.
+
+-spec update_rewards_shares( Type :: reward_types(),
+                             Shares :: number(),
+                             SharesList :: [ rewards_shares() ] ) -> [ rewards_shares() ].
+update_rewards_shares(Type, Shares, SharesList) ->
+    case lists:keyfind(Type, #rewards_shares.type, SharesList) of
+        false -> [ new_rewards_share(Type, Shares) ];
+        #rewards_shares{ shares = Current } = Rec ->
+            lists:keyreplace(Type, #rewards_shares.type, SharesList,
```
vihu, line 774: second arg should be an integer not a `rewards_shares.type` which can be undefined | some_reward_type

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -1231,15 +1250,99 @@ dc_reward(Txn, End, AccIn, Ledger, #{ sc_grace_blocks := GraceBlocks,
...
+                          MDRecord :: rewards_meta() ) -> NewMDRecord :: rewards_meta().
+update_shares_list(Type, Amount, #rewards_meta{shares = SharesList} = MDRecord) ->
+    NewSharesList = case lists:keyfind(Type, #rewards_shares.type, SharesList) of
+                        false ->
+                            %% no record of this type exists, add it to our list
+                            [ #rewards_shares{type = Type, shares = Amount} | SharesList ];
+                        #rewards_shares{shares = S} = Rec ->
+                            %% update existing record with new amount and put it back
+                            %% into our list
+                            lists:keyreplace(Type, #rewards_shares.type,
```
vihu, line 1319: same, second arg needs to be an integer

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -724,61 +694,104 @@ securities_rewards(Ledger, #{epoch_reward := EpochReward,
...
+get_shares(Type, Gateway, RewardsMD, Ledger, Default) ->
+    case blockchain_ledger_v1:find_gateway_owner(Gateway, Ledger) of
+        {ok, GwOwner} ->
+            case maps:get(GwOwner, RewardsMD, Default) of
+                %% this gateway owner is NOT in the rewards metadata (yet)
+                Default -> Default;
+                #rewards_meta{ shares = S } ->
+                    %% ok, this owner IS in the metadata, so let's find
+                    %% the share data for this rewards type
+                    case lists:keyfind(Type, #rewards_shares.type, S) of
```
vihu, line 720: same issue, second arg here also should be an integer

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -1231,15 +1250,99 @@ dc_reward(Txn, End, AccIn, Ledger, #{ sc_grace_blocks := GraceBlocks,
...
+                                  update_shares_list(Type, Shares, MD)
+                          end,
+                     #rewards_meta{shares = [new_rewards_shares(Type, Shares)]},
+                     RewardsMeta).
+
+-spec update_shares_list( Type :: reward_types(),
+                          Amount :: number(),
+                          MDRecord :: rewards_meta() ) -> NewMDRecord :: rewards_meta().
+update_shares_list(Type, Amount, #rewards_meta{shares = SharesList} = MDRecord) ->
+    NewSharesList = case lists:keyfind(Type, #rewards_shares.type, SharesList) of
```
vihu, line 1312: another one...

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -701,14 +671,19 @@ consensus_members_rewards(Ledger, #{consensus_epoch_reward := EpochReward,
       fun(Member, Acc) ->
               PercentofReward = 100/Count/100,
               Amount = erlang:round(PercentofReward*ConsensusReward),
-              maps:put({GwOrVal, consensus, Member}, Amount+OveragePerMember, Acc)
+              case get_owner_address(GwOrVal, Member, Ledger) of
+                  {ok, Owner} ->
+                      update_hnt(Owner, Amount + OveragePerMember, Acc);
```
vihu, line 676: `update_hnt/3` fun signature expects the first arg to be `rewards_metadata()` => :boom:

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -724,61 +699,104 @@ securities_rewards(Ledger, #{epoch_reward := EpochReward,
             Balance = blockchain_ledger_security_entry_v1:balance(Entry),
             PercentofReward = (Balance*100/TotalSecurities)/100,
             Amount = erlang:round(PercentofReward*SecuritiesReward),
-            maps:put({owner, securities, Key}, Amount, Acc)
+            update_hnt(Key, Amount, Acc)
```
vihu, line 702: another one, `update_hnt/3` first arg is `rewards_metadata()`

evanmcc:
superseded by later work.

PR #889, opened 2021-06-29
Update debit_fee to include potential implicit burn storage

syuan100:
Updating a few new txns so debit_fee can track implicit burns.


abhay:
@syuan100 raised a good question in DMs. How do we ensure this method is used for all new transactions (that do implicit burn) going forward?

syuan100:
Gentle bump on this, hoping to get this merged in soon! Would like to post a v1 of the rosetta implementation on their forums at the end of this week 🤞

evanmcc (approved):

evanmcc:
@syuan100 thanks for this.  I think that the best way is probably to deprecate the version of debit fee that doesn't take the hash and the chain?

PR #891, opened 2021-06-30
change error to exit, filter dial issues

evanmcc:
#746  has made sync logging very noisy, because error was used instead of exit.  This PR fixes that and further filters out any dial error printing (which can be enabled with debug if needed).

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -596,7 +596,7 @@ handle_info({'DOWN', SyncRef, process, _SyncPid, Reason},
             {Hash, Height} = State#state.snapshot_info,
             {noreply, snapshot_sync(Hash, Height, State)};
         _ ->
-            lager:info("block sync down reason ~p", [Reason]),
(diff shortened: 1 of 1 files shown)


src/blockchain_worker.erl
```diff
@@ -596,7 +596,7 @@ handle_info({'DOWN', SyncRef, process, _SyncPid, Reason},
             {Hash, Height} = State#state.snapshot_info,
             {noreply, snapshot_sync(Hash, Height, State)};
         _ ->
-            lager:info("block sync down reason ~p", [Reason]),
+            case Reason of dial -> ok; _ -> lager:info("block sync down: ~p", [Reason]) end,
```
madninja, line 599: the noise I was was for the error(dial).. is this to just silence untying to do with dialing?

Vagabond: yes, this just quiets failed dials (as opposed to timeouts and other failures)

madninja (approved):

PR #892, opened 2021-06-30
Delete corrupt ledger on open

Vagabond:


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -451,6 +451,8 @@ sweep_old_checkpoints(Ledger) ->
     ok.
 
 clean_checkpoints(#ledger_v1{dir = RecordDir}) ->
+    clean_checkpoints(RecordDir);
```
madninja, line 454: this doesn't hurt, but doesn't really do anything useful either right?

madninja (approved):

PR #895, opened 2021-07-01
Macpie/bloom fix

macpie:
I think I have identified why we see so many late packets it seems that the `ClientBloom` filter is miss configured.
I have a CT that can replicate potential issue. Not sure if this is just due to the algorithm or if the implementation is broken.

I went ahead and removed the client bloom anyway as it is mostly useless (I think, correct me if I am wrong). I seemed to be in place to avoid lookup but it is only a `maps:get` so it seems unnecessary.

I will try to test this asap on dev so I can confirm improvement.

### DIFF
--- a/src/state_channel/blockchain_state_channels_server.erl
+++ b/src/state_channel/blockchain_state_channels_server.erl
@@ -64,8 +64,7 @@
 -type sc_key() :: blockchain_state_channel_v1:id().
 -type sc_value() :: {blockchain_state_channel_v1:state_channel(), skewed:skewed()}.
 -type state_channels() :: #{sc_key() => sc_value()}.
--type bloom_value() :: {ClientBloom :: bloom_nif:bloom(), PacketBloom :: bloom_nif:bloom()}.
--type blooms() :: #{sc_key() => bloom_value()}.
+-type blooms() :: #{sc_key() => bloom_nif:bloom()}.
 -type streams() :: #{libp2p_crypto:pubkey_bin() => {pid(), reference()}}.
 
 -ifdef(TEST).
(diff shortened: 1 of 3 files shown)


src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -915,23 +917,17 @@ update_sc_summary(ClientPubKeyBin, PayloadSize, DCPayloadSize, SC, ClientBloom,
...
+            lager:debug("update_sc_summary ~p", [blockchain_utils:addr2name(ClientPubKeyBin)]),
             %% Update summaries
             {NewSC, DidFit} = blockchain_state_channel_v1:update_summary_for(ClientPubKeyBin,
                                                                              NewSummary,
                                                                              SC,
-                                                                             bloom:check(ClientBloom, ClientPubKeyBin),
                                                                              MaxActorsAllowed),
-            ok = maybe_set_client_bloom(ClientPubKeyBin, ClientBloom, DidFit),
+            lager:debug("update_sc_summary ~p", [blockchain_utils:addr2name(ClientPubKeyBin)]),
+            lager:debug("update_sc_summary ~p", [blockchain_utils:addr2name(ClientPubKeyBin)]),
```
michaeldjeffrey, line 927: Double log here. 
Want to add some different text to the ones above?

macpie: i forgot my testing logs I need to remove them

test/blockchain_bloom_SUITE.erl
```diff
@@ -0,0 +1,46 @@
...
+
+all() ->
+    [
+        bloom_test
+    ].
+
+%% Right now we put about 1100 actors per SC with the current 15000000 size it seems to be causing issues
+%% I have here an exemple of a better tailored bloom. If you repeat this test it will pass 99% of the time
+%% if you decide to increase m (Number of bits in the filter)  it will start having issue and taking too long
+%% https://hur.st/bloomfilter/?n=1100&p=1.0E-6&m=&k=20
```
JayKickliter, line 25: For our purposes, how big of a deal is a false positive? We're calculating the size of the filter for a really small false-positive rate (0.0001 % ?). If size/speed is an issue, maybe we can increase the allowable FP rate?

JayKickliter:
> I think I have identified why we see so many late packets it seems that the ClientBloom filter is miss configured.

Looking at the test you added, I still don't understand what's wrong or how it was misconfigured. Can you elaborate please?

test/blockchain_bloom_SUITE.erl
macpie: we could it would not hurt I guess, but once I was able to tailor numbers better it was faster

PR #896, opened 2021-07-07
Write a ledger upgrade hook to fix witness last location nonces of 0

Vagabond:

PR #903, opened 2021-07-08
Trigger witness location nonces updates via ledger chainvar hooks

andymck:
The current defaulting to a value of zero for witness related nonce vars ( WitnessGW#gateway_v2.last_location_nonce and Witness#witness.added_location_nonce )  makes it difficult to converge on agreement for a new snap as the application of the default value is non deterministic ( for example a witness GW could reassert its location at any time and that update may not be seen at the same time across the spots thus the spots vary in the value of these vars )

The fix here is to default the mentioned two witness vars to the current nonce value of the witness GW and  to do so at a consistent point in time.  This is achieved by linking the updating of the vars to a chain var with a specific value.  The provides the necessary determinism to ensure the spots all have a consistent view and thus thereafter be able to converge on a new snap.

This PR pulls in and amends the following two previous PRs:

https://github.com/helium/blockchain-core/pull/352
https://github.com/helium/blockchain-core/pull/896

TODO:
------
There are a couple of TODO's noted in the comments, which should be reviewed

### DIFF
--- a/include/blockchain.hrl
+++ b/include/blockchain.hrl
@@ -27,6 +27,6 @@
 
 -define(BC_UPGRADE_NAMES, [<<"gateway_v2">>, <<"hex_targets">>, <<"gateway_oui">>,
                            <<"h3dex">>, <<"h3dex2">>,
-                           <<"gateway_lg3">>]).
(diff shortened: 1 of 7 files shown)


andymck:
superseded by https://github.com/helium/blockchain-core/pull/920

PR #909, opened 2021-07-10
Make snap download and load into discrete steps

mrallen1:
Problem to solve: When a spot fails to deserialize or load a snapshot from an S3 download, it can download the same file over and over and over.

Solution: Separate snapshot download and load into discrete steps. Also need to deal with disk space management, partial downloads, snapshot deserialization failures and invalid snapshot hashes. We check for a "plausible" looking file before we download a new copy and if we encounter errors, we delete the now implausible file and eventually we'll try to download the snapshot again. A function cleans up snapshots older than 1 week when we start a download attempt. This should keep the spot storage reasonably small and avoid loops where a spot will download a snapshot and get into an OOM loop.

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -8,6 +8,7 @@
 -behavior(gen_server).
 
 -include("blockchain_vars.hrl").
+-include_lib("kernel/include/file.hrl").
 
 %% ------------------------------------------------------------------
 %% API Function Exports
@@ -69,6 +70,8 @@
 -include("blockchain.hrl").
 
 -define(SERVER, ?MODULE).
+-define(READ_SIZE, 16 * 1024). % read 16 kb chunks
+-define(WEEK_OLD_SECONDS, 7*24*60*60). %% a week's worth of seconds
 
 -ifdef(TEST).
 -define(SYNC_TIME, 1000).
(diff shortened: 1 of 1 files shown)


src/blockchain_worker.erl
```diff
@@ -973,34 +969,105 @@ attempt_fetch_p2p_snapshot(Hash, Height, Swarm, Chain, Peer) ->
...
               ],
 
-    lager:info("Attempting snapshot download from ~p", [Url]),
+    lager:info("Attempting snapshot download from ~p, writing to scratch file ~p",
+               [Url, ScratchFile]),
     case httpc:request(get, {Url, Headers}, HTTPOptions, Options) of
-        {ok, {200, Response}} -> {ok, Response};
+        {ok, saved_to_file} ->
+            lager:info("snap written to scratch file ~p", [ScratchFile]),
+            {ok, _BytesCopied} = file:copy(ScratchFile, Filepath),
```
Vagabond, line 1023: Suggest a file rename as that's atomic.

src/blockchain_worker.erl
```diff
@@ -973,34 +972,87 @@ attempt_fetch_p2p_snapshot(Hash, Height, Swarm, Chain, Peer) ->
...
+    Filepath = filename:join([BaseDir, "snap", Filename]),
+    ok = filelib:ensure_dir(Filepath),
+
+    %% clean_dir will remove any snapshots older than 1 week (should help prevent
+    %% filling up the SSD card with old snapshots)
+    ok = clean_dir(filename:dirname(Filepath)),
+
+    case filelib:is_regular(Filepath) of
+        true ->
+            lager:info("Already have snapshot for height ~p - hash matches", [Height]),
```
evanmcc, line 995:
```suggestion
            lager:info("Already have snapshot file for height ~p", [Height]),
```

evanmcc (approved):

PR #910, opened 2021-07-11
lazy snapshot field deserialization

evanmcc:
I got this to the point where I could load a node that would sync using it, but I'm tapped out.

this attempts to use a lot less memory, or at least use it in a less all-at-once way, by leaving fields in binary format until the last moment before use.

This needs:
- better abstraction/cleanup, it was written in haste
- diff testing
- roundtrip testing (I suspect that the new_test is passing vacuously and the basic test seems broken)
- close review

### DIFF
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -26,14 +26,20 @@
          diff/2
         ]).
 
+-ifdef(TEST).
+-export([
+         deserialize_field/2
+        ]).
+-endif.
+
 -export_type([
     snapshot/0,
     snapshot_v5/0
 ]).
 
--type state_channel() ::
-      {blockchain_ledger_state_channel_v1, blockchain_ledger_state_channel_v1:state_channel()}
-    | {blockchain_ledger_state_channel_v2, blockchain_ledger_state_channel_v2:state_channel_v2()}.
+%% -type state_channel() ::
+%%       {blockchain_ledger_state_channel_v1, blockchain_ledger_state_channel_v1:state_channel()}
+%%     | {blockchain_ledger_state_channel_v2, blockchain_ledger_state_channel_v2:state_channel_v2()}.
 
 %% this assumes that everything will have loaded the genesis block
 %% already.  I'm not sure that's totally safe in all cases, but it's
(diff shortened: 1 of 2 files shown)


src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -417,7 +421,10 @@ import(Chain, SHA, #{version := v6}=Snapshot) ->
     L :: blockchain_ledger_v1:ledger(),
     M :: blockchain_ledger_v1:mode().
 load_into_ledger(Snapshot, L0, Mode) ->
-    Get = fun (K) -> maps:get(K, Snapshot) end,
+    Get = fun (K) ->
+                  {value, {K, V}} = lists:keysearch(K, 1, Snapshot),
```
mrallen1, line 425: Would prefer `lists:keyfind` here

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -282,22 +282,26 @@ serialize(Snapshot, BlocksOrNoBlocks) ->
     Serialize(Snapshot, BlocksOrNoBlocks).
 
 -spec serialize_v6(snapshot_v6(), blocks | noblocks) -> iolist().
-serialize_v6(#{version := v6}=Snapshot, BlocksOrNoBlocks) ->
+serialize_v6(Snapshot, BlocksOrNoBlocks) when is_list(Snapshot) ->
     Key = blocks,
     Blocks =
         case BlocksOrNoBlocks of
```
mrallen1, line 288: Could eliminate this case statement by matching `blocks` or `noblocks` in the function head.

mrallen1 (changes requested):
Overall this looks like a small change that could bring big benefits. I have a couple of changes to request.

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -282,22 +282,26 @@ serialize(Snapshot, BlocksOrNoBlocks) ->
...
+                          blockchain_block:serialize(B);
+                      (B) -> B
+                  end,
+                  Bs
+                 );
             noblocks ->
                 []
         end,
-    Pairs = lists:keysort(1, maps:to_list(maps:put(Key, Blocks, Snapshot))),
+    Pairs = lists:keysort(1, lists:keyreplace(Key, 1, Snapshot, {Key, Blocks})),
```
xandkar, line 304: `lists:keyreplace` behaves differently from `maps:put`. We probably want `lists:keystore` instead:

```
1> lists:keyreplace(k, 1, [], {k, v}).
[]
2> lists:keystore(k, 1, [], {k, v}).
[{k,v}]
3> maps:put(k, v, #{}).
#{k => v}
4>
```

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -870,6 +897,7 @@ v4_to_v5(#blockchain_snapshot_v4{
            }) ->
     #{
       version => v5,
+      %% TODO these need to be reserialized to please dialyzer (and probably to work?)
```
xandkar, line 900: v5 and v6 just need different types to show their difference.

vihu (approved):

vihu:
I can confirm that the tests pass and one v5 and two v6 snapshots load with the changes here.

`Testing lib.blockchain.blockchain_snapshot_SUITE: TEST COMPLETE, 3 ok, 0 failed of 3 test cases`

PR #914, opened 2021-07-12
Force clean witnesses, don't allow them to be added, disable gc

evanmcc:
this is not the nicest PR, but it's faster than the alternative, which is to add a new var and wait.


Vagabond (approved):

vihu (approved):
This will be updated further in #919

PR #922, opened 2021-07-15
HIP-28 consensus reward calc grace period

PaulVMo:
As described in [HIP-28](https://github.com/helium/HIP/blob/master/0028-consensus-reward-adjustments.md), this update adds a grace period in the consensus epoch reward calculation. Instead of always basing the calculation on election_interval number of block epochs, allow for up to election_interval + election_retry_interval blocks to be used in an epoch for calculating consensus rewards in order to provide time for the election itself. Change is protected behind reward_version chain var and is active when the value is >=6.

This PR does not address the other suggested changes in the HIP, specifically reallocating unearned consensus rewards to PoC as evanmcc and others have some alternative suggestions.

Lastly, this fixes the epoch reward calc for other reward types as well which was off by one block. Start and End are inclusive but epoch length was calculated as End - Start. Changed to End - Start + 1.

To-Do: add some test cases. I have verified it compiles and all existing EUnit tests execute successfully. However, I have not added any tests specific to the new code.

### DIFF
--- a/src/transactions/v2/blockchain_txn_rewards_v2.erl
+++ b/src/transactions/v2/blockchain_txn_rewards_v2.erl
@@ -277,10 +277,18 @@ calculate_rewards_metadata(Start, End, Chain) ->
         Results0 = fold_blocks_for_rewards(Start, End, Chain,
                                           Vars, Ledger, AccInit),
 
-        %% Forcing calculation of the EpochReward amount for the CG to always
+        %% Prior to HIP 28 (reward_version <6), force EpochReward amount for the CG to always
         %% be around ElectionInterval (30 blocks) so that there is less incentive
-        %% to stay in the consensus group
-        ConsensusEpochReward = calculate_epoch_reward(1, Start, End, Ledger),
+        %% to stay in the consensus group. With HIP 28, relax that to be up to election_interval +
+        %% election_retry_interval to allow for time for election to complete.
+        ConsensusEpochReward = 
+            case maps:get(reward_version,Vars) of
+               RewardVersion when RewardVersion >= 6 ->
+                    calculate_consensus_epoch_reward(Start, End, Vars);
+                _ ->
+                    calculate_epoch_reward(1, Start, End, Ledger)
+            end,
+
         Vars1 = Vars#{ consensus_epoch_reward => ConsensusEpochReward },
 
         Results = finalize_reward_calculations(Results0, Ledger, Vars1),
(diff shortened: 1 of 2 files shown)


src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -276,10 +276,18 @@ calculate_rewards_metadata(Start, End, Chain) ->
...
 
-        %% Forcing calculation of the EpochReward amount for the CG to always
+        %% Prior to HIP 28 (reward_version <6), force EpochReward amount for the CG to always
         %% be around ElectionInterval (30 blocks) so that there is less incentive
-        %% to stay in the consensus group
-        ConsensusEpochReward = calculate_epoch_reward(1, Start, End, Ledger),
+        %% to stay in the consensus group. With HIP 28, relax that to be up to election_interval +
+        %% election_retry_interval to allow for time for election to complete.
+        ConsensusEpochReward = 
+            case maps:get(reward_version,Vars) of
```
evanmcc, line 284: this should maybe be 3?  this also needs to be changed in the vars txn validation, otherwise we can't enable it.

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -683,6 +701,24 @@ calculate_epoch_reward(_Version, _Start, _End, BlockTime0, ElectionInterval, Mon
...
+                                                election_restart_interval := ElectionRestartInterval, 
+                                                monthly_reward := MonthlyReward }) ->
+
+    BlockTime1 = (BlockTime0/1000),
+    % Convert to blocks per min
+    BlockPerMin = 60/BlockTime1,
+    % Convert to blocks per month
+    BlockPerMonth = BlockPerMin*60*24*30,
+    % Calculate epoch length in blocks, cap at election interval + grace period
+    EpochLength = erlang:min(End - Start + 1, ElectionInterval + ElectionRestartInterval),
```
evanmcc, line 718: I'm not sure that this is adequate.  We probably need access to the transaction to judge this correctly at this point.

src/transactions/v2/blockchain_txn_rewards_v2.erl
PaulVMo: The current reward_version is 5 which is where 6 came from. Thanks for the point out of the vars transaction validation. I will incremented check on reward_version there as well.

src/transactions/v2/blockchain_txn_rewards_v2.erl
PaulVMo: I modeled this on calculate_epoch_reward just above this which does not have access to the transaction either. If you recommend a different approach here, happy to work it in. I just may need a little more guidance.

PaulVMo:
@evanmcc any chance you would be able to review my responses to your comments? I would like to have this included in the agenda for the community call tomorrow July 28 if possible.

evanmcc:
in the future please rebase on master instead of merging master into the branch.

Vagabond (approved):
Would be nice to clean up the commit history here.

PaulVMo:
I think i was able to finally clean up the commit history. let me know how it looks.

evanmcc (approved):

PR #926, opened 2021-07-26
add net emissions code

evanmcc:
closes https://github.com/helium/miner/issues/863, please see that issue for more details

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -474,3 +474,13 @@
 -define(penalty_history_limit, penalty_history_limit). % blocks
 -define(dkg_penalty, dkg_penalty). % float
 -define(tenure_penalty, tenure_penalty). % float
+
+%%%
+%%% net emissions vars
+%%%
+
+%% use net emissions code paths
+-define(net_emissions_enabled, net_emissions_enabled). % boolean
+%% maximum amount of burnt hnt per epoch to return to the rewards pool
+%% initial proposed max 34.24 HNT
+-define(net_emissions_max_rate, net_emissions_max_rate). % bones
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -91,6 +91,12 @@
     check_dc_balance/3,
     check_dc_or_hnt_balance/4,
 
+    hnt_burned/1,
+    add_hnt_burned/2,
+    clear_hnt_burned/1,
+
+    net_overage/1,  net_overage/2,
+
     token_burn_exchange_rate/1,
     token_burn_exchange_rate/2,
 
@@ -319,6 +325,8 @@
 -define(MULTI_KEYS, <<"multi_keys">>).
 -define(VARS_NONCE, <<"vars_nonce">>).
 -define(BURN_RATE, <<"token_burn_exchange_rate">>).
+-define(HNT_BURNED, <<"hnt_burned">>).
+-define(NET_OVERAGE, <<"net_overage">>).
 -define(CURRENT_ORACLE_PRICE, <<"current_oracle_price">>). %% stores the current calculated price
 -define(ORACLE_PRICES, <<"oracle_prices">>). %% stores a rolling window of prices
 -define(hex_list, <<"$hex_list">>).
@@ -2774,6 +2782,7 @@ debit_fee(Address, Fee, Ledger, MaybeTryImplicitBurn, TxnHash, Chain) ->
     case ?MODULE:find_dc_entry(Address, Ledger) of
         {error, dc_entry_not_found} when MaybeTryImplicitBurn == true ->
             {ok, FeeInHNT} = ?MODULE:dc_to_hnt(Fee, Ledger),
+            ok = add_hnt_burned(FeeInHNT, Ledger),
             ?MODULE:debit_fee_from_account(Address, FeeInHNT, Ledger, TxnHash, Chain);
         {error, _}=Error ->
             Error;
@@ -2791,6 +2800,7 @@ debit_fee(Address, Fee, Ledger, MaybeTryImplicitBurn, TxnHash, Chain) ->
                 {false, true} ->
                     %% user does not have sufficient DC balance, try to do an implicit hnt burn instead
                     {ok, FeeInHNT} = ?MODULE:dc_to_hnt(Fee, Ledger),
+                    ok = add_hnt_burned(FeeInHNT, Ledger),
                     ?MODULE:debit_fee_from_account(Address, FeeInHNT, Ledger, TxnHash, Chain);
                 {false, false} ->
                     {error, {insufficient_dc_balance, {Fee, Balance}}}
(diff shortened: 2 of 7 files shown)


src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -147,6 +147,39 @@ is_valid(Txn, Chain) ->
 absorb(Txn, Chain) ->
     Ledger = blockchain:ledger(Chain),
 
+    case blockchain:config(?net_emissions_enabled, Ledger) of
+        {ok, true} ->
+            %% initial proposed max 34.24
+            {ok, Max} = blockchain:config(?net_emissions_max_rate, Ledger),
+            {ok, Burned} = blockchain_ledger_v1:hnt_burned(Ledger),
```
vihu, line 154: Do we want to :boom: if this and the next line return `{error, whatever}` (because they can)?

src/transactions/v1/blockchain_txn_rewards_v1.erl
```diff
@@ -136,6 +136,39 @@ is_valid(Txn, Chain) ->
 absorb(Txn, Chain) ->
     Ledger = blockchain:ledger(Chain),
 
+    case blockchain:config(?net_emissions_enabled, Ledger) of
```
vihu, line 139: I don't _think_ `rewards_v1` needed to change tbh...

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -147,6 +147,39 @@ is_valid(Txn, Chain) ->
...
+            {ok, Overage} = blockchain_ledger_v1:net_overage(Ledger),
+
+            %% clear this since we have it already
+            ok = blockchain_ledger_v1:clear_hnt_burned(Ledger),
+
+            case Burned > Max of
+                %% if burned > max, then add to overage
```
vihu, line 161:
```suggestion
                %% if burned > max, then add (burned - max) to overage
```

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -147,6 +147,39 @@ is_valid(Txn, Chain) ->
...
+                true ->
+                    Overage1 = Overage + (Burned - Max),
+                    ok = blockchain_ledger_v1:net_overage(Overage1, Ledger);
+                %% else we may have pulled from overage to the tune of
+                %% max - burned
+                 _ ->
+                    %% here we pulled from overage up to max
```
vihu, line 168: Off chance that burned = max? Should we handle each clause more explicitly here?

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -659,7 +692,18 @@ calculate_epoch_reward(Version, Start, End, Ledger) ->
...
     {ok, BlockTime0} = blockchain:config(?block_time, Ledger),
     {ok, MonthlyReward} = blockchain:config(?monthly_reward, Ledger),
-    calculate_epoch_reward(Version, Start, End, BlockTime0, ElectionInterval, MonthlyReward).
+    Reward = calculate_epoch_reward(Version, Start, End, BlockTime0,
+                                    ElectionInterval, MonthlyReward),
+    case blockchain:config(?net_emissions_enabled, Ledger) of
+        {ok, true} ->
+            %% initial proposed max 34.24
+            {ok, Max} = blockchain:config(?net_emissions_max_rate, Ledger),
+            {ok, Burned} = blockchain_ledger_v1:hnt_burned(Ledger),
```
vihu, line 701: Same as above, these calls _may_ return `{error, _}`

src/transactions/v2/blockchain_txn_rewards_v2.erl
evanmcc: that was intentional, yes.  if these are returning errors, we probably cannot sensibly proceed.

src/transactions/v1/blockchain_txn_rewards_v1.erl
evanmcc: I'm ok with removing this code if desired.

src/transactions/v2/blockchain_txn_rewards_v2.erl
evanmcc: I think that this code is safe with 0, but maybe we should test it explicitly?

src/transactions/v1/blockchain_txn_rewards_v1.erl
vihu: I'd vote yes :slightly_smiling_face:

evanmcc: will do.

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -659,7 +692,18 @@ calculate_epoch_reward(Version, Start, End, Ledger) ->
...
+            {ok, Overage} = blockchain_ledger_v1:net_overage(Ledger),
+            Reward + max(Max, Burned + Overage);
```
evanmcc, line 703: this should actually be min here, I think I was thrown off by the variable name.

include/blockchain_vars.hrl
```diff
@@ -474,3 +474,13 @@
...
 -define(tenure_penalty, tenure_penalty). % float
+
+%%%
+%%% net emissions vars
+%%%
+
+%% use net emissions code paths
+-define(net_emissions_enabled, net_emissions_enabled). % boolean
+%% maximum amount of burnt hnt per epoch to return to the rewards pool
+%% initial proposed max 34.24 HNT
```
Vagabond, line 485: Is there a rationale for this value?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2826,6 +2835,67 @@ check_dc_or_hnt_balance(Address, Amount, Ledger, IsFeesEnabled) ->
...
+                end,
+            New = Prev + Burned,
+            cache_put(Ledger, DefaultCF, ?HNT_BURNED, <<New:64/integer-unsigned-big>>);
+        _ -> ok
+    end.
+
+-spec clear_hnt_burned(ledger()) -> ok.
+clear_hnt_burned(Ledger) ->
+    DefaultCF = default_cf(Ledger),
+    cache_put(Ledger, DefaultCF, ?HNT_BURNED, <<0:64/integer-unsigned-big>>).
```
Vagabond, line 2872: Could also delete here?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2826,6 +2835,67 @@ check_dc_or_hnt_balance(Address, Amount, Ledger, IsFeesEnabled) ->
             end
     end.
 
+-spec hnt_burned(ledger()) -> {ok, non_neg_integer()} | {error, any()}.
+hnt_burned(Ledger) ->
+    case blockchain:config(?net_emissions_enabled, Ledger) of
+        {ok, true} ->
+            DefaultCF = default_cf(Ledger),
+            case cache_get(Ledger, DefaultCF, ?HNT_BURNED, []) of
+                {ok, <<Burned:64/integer-unsigned-big>>} ->
```
Vagabond, line 2844: There's no need to store this big endian as the value is not subject to lexiographic order. Little endian machines are far more prevalent these days so this is sort of a pointless byte swap on 99% of machines.

Vagabond:
Should we make the halvening automatic based on block height as well so we don't need to adjust it anymore?

include/blockchain_vars.hrl
evanmcc: it's what is in the HIP, you'd need to ask @tjain-mcc

src/transactions/v2/blockchain_txn_rewards_v2.erl
evanmcc: added a test here, seems to work as expected

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: we just rewrite soon enough, but sure.

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: will fix

evanmcc:
@Vagabond this code isn't effected by halvening.  We do need to move certain configs into some height-keyed config, though.

include/blockchain_vars.hrl
tjain-mcc: yes it was equal to 50k HNT per month, or 1% of the initial level of emissions

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -663,7 +696,18 @@ calculate_epoch_reward(Version, Start, End, Ledger) ->
     {ok, ElectionInterval} = blockchain:config(?election_interval, Ledger),
     {ok, BlockTime0} = blockchain:config(?block_time, Ledger),
     {ok, MonthlyReward} = blockchain:config(?monthly_reward, Ledger),
-    calculate_epoch_reward(Version, Start, End, BlockTime0, ElectionInterval, MonthlyReward).
+    Reward = calculate_epoch_reward(Version, Start, End, BlockTime0,
```
mrallen1, line 699: shouldn't this whole block live inside of the `calculate_epoch_reward` function? What about when we calculate rewards for consensus? They won't include this amount then.

mrallen1 (commented):
Concerned about the position of the net emission calculation

mrallen1:
One more thing - if we're going to expose the `net emissons` reward value via API, we will need to report it in the rewards metadata map ETL uses

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1281,6 +1281,15 @@ validate_var(?penalty_history_limit, Value) ->
...
     validate_int(Value, "penalty_history_limit", 10, 100000, false);
 
+validate_var(?net_emissions_enabled, Value) ->
+    case Value of
+        true -> ok;
+        false -> ok;
+        _ -> throw({error, {invalid_net_emissions_boolean, Value}})
+    end;
+validate_var(?net_emissions_max_rate, Value) ->
+    validate_int(Value, "net_emissions_max_rate", 0, ?bones(200), false);
```
Vagabond, line 1291: How does this relate to the 34.24 HNT above?

Vagabond:
Snapshots need to cover this value, right?

src/transactions/v1/blockchain_txn_vars_v1.erl
Vagabond: Is the limit really 200 bones?

evanmcc: the ?bones macro converts an HNT amount into bones.  the names is maybe bad.

Vagabond (approved):

PR #931, opened 2021-08-01
Deprecate debit_fee/4

syuan100:
Deprecates outdated `debit_fee` to ensure that implicit burns are recorded when necessary. Will prevent PRs like [this](https://github.com/helium/blockchain-core/pull/889) in the future.


syuan100:
@evanmcc addresses deprecation that was mentioned here: https://github.com/helium/blockchain-core/pull/889#issuecomment-884463696

PR #934, opened 2021-08-03
Use direct rocksdb operations when loading snap; fix snap sync check

Vagabond:


src/blockchain.erl
```diff
@@ -2153,7 +2153,7 @@ init_blessed_snapshot(Blockchain, _HashAndHeight={Hash, Height0}) when is_binary
     %% loaded.  before this fix, if we crashed after loading, even if
     %% we succeeded loading, we'd redo that work.
     Height = Height0 - 1,
-    case blockchain:height(Blockchain) of
+    case blockchain_ledger_v1:current_height(blockchain:ledger(Blockchain)) of
```
xandkar, line 2156: What is the _relevant_ difference between these two sources of height? It looks the new one looks it up it a cache instead of directly in rocks. Why is that needed here?

Vagabond: I believe the old code here was buggy. The *ledger* height is what dictates if we need to load a snapshot, not the blockchain height.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3796,14 +3811,18 @@ validators_cf(Ledger) ->
...
+        {Cache, _GwCache} ->
+            true = ets:insert(Cache, {{Name, Key}, Value})
+    end,
     ok.
 
 -spec cache_get(ledger(), rocksdb:cf_handle(), any(), [any()]) -> {ok, any()} | {error, any()} | not_found.
 cache_get(Ledger, {Name, DB, CF}, Key, Options) ->
     case context_cache(Ledger) of
-        {undefined, undefined} ->
+        {C, _GwCache} when C == undefined; C == direct ->
```
xandkar, line 3825: I generally dislike this way of working around our lack of or-patterns and would instead call the same thunk from two patterns.

But, that's just like my opinion, man...

src/blockchain.erl
xandkar: Ah-ha! Thanks.

xandkar (approved):

mrallen1 (approved):

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -575,6 +575,11 @@ new_context(Ledger) ->
     GwCache = ets:new(gw_cache, [set, protected, {keypos, 1}]),
     context_cache(Cache, GwCache, Ledger).
 
+new_direct_context(Ledger) ->
```
vihu, line 578:
```suggestion
-spec new_direct_context(ledger()) -> ok.
new_direct_context(Ledger) ->
```

Vagabond: I'm pretty sure this returns a new ledger?

vihu: oh my bad, let me edit the suggestion, sorry

PR #935, opened 2021-08-04
support txn submitter to provide timestamp key to be used in txn mgr cache

andymck:
Allow the key by which txn mgr will cache txns to be provided by the submitter ( such as ETL )

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -206,6 +206,7 @@ integrate_genesis_block(Block) ->
 
 %%--------------------------------------------------------------------
 %% @doc
+%% submit_txn/1 is deprecated.  use blockchain_txn_mgr:submit_txn/2 instead
 %% @end
 %%--------------------------------------------------------------------
 -spec submit_txn(blockchain_txn:txn()) -> ok.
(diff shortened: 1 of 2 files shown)


src/blockchain_worker.erl
```diff
@@ -19,7 +19,7 @@
     num_consensus_members/0,
     consensus_addrs/0,
     integrate_genesis_block/1,
-    submit_txn/1, submit_txn/2,
+    submit_txn/1, submit_txn/2, submit_txn/3,
```
Vagabond, line 22: I suggest we not extend the blockchain_worker submit txn apis because they predate the txn manager and should be deprecated.

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -81,6 +81,11 @@ start_link(Args) when is_map(Args) ->
 submit(Txn, Callback) ->
```
Vagabond, line 81: You could generate the key here, and then call into a single submit path.

Vagabond (approved):

madninja (approved):

PR #939, opened 2021-08-05
Be more robust in the face of stale state channel banners

Vagabond:
Router sometimes sends us a stale state channel banner, this was
improperly being detected as a state channel conflict and leading the
hotspot to stop sending packets for this state channel. This PR improves
some logging, avoids writing duplicate copies of the SC to disk and
fixes the causality check for when a stale banner is received.


vihu (approved):
lgtm

michaeldjeffrey (approved):

PR #940, opened 2021-08-05
Be more selective about bannering state channels

michaeldjeffrey:
- env var to not banner all updates

- first time we encounter a client, it will be bannered the sc it was
  placed into

- when a sc closes, banners will be sent with replacement sc to clients
  in the closed sc

### DIFF
--- a/src/state_channel/blockchain_state_channels_server.erl
+++ b/src/state_channel/blockchain_state_channels_server.erl
@@ -65,7 +65,8 @@
 -type sc_value() :: {blockchain_state_channel_v1:state_channel(), skewed:skewed()}.
 -type state_channels() :: #{sc_key() => sc_value()}.
 -type blooms() :: #{sc_key() => bloom_nif:bloom()}.
--type streams() :: #{libp2p_crypto:pubkey_bin() => {pid(), reference()}}.
+-type stream() :: {pid(), reference()}.
+-type streams() :: #{libp2p_crypto:pubkey_bin() => stream()}.
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
@@ -344,7 +345,7 @@ handle_packet(ClientPubKeyBin, Packet, HandlerPid,
             case SCVer > 1 andalso bloom:check_and_set(PacketBloom, Payload) of
                 true ->
                     %% Don't add payload
-                    maybe_add_stream(ClientPubKeyBin, HandlerPid, State);
+                    maybe_add_stream(SC, ClientPubKeyBin, HandlerPid, State);
                 false ->
                     {SC1, Skewed1} = blockchain_state_channel_v1:add_payload(Payload, SC, Skewed),
                     ExistingSCNonce = blockchain_state_channel_v1:nonce(SC1),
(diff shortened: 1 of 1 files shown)


src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -458,21 +459,29 @@ handle_offer(SCOffer, HandlerPid, #state{db=DB, dc_payload_size=DCPayloadSize, a
...
+maybe_add_stream(ChosenSC, ClientPubKeyBin, Stream, #state{streams=Streams}=State) ->
+    ClientUpdate =
+        case maps:get(ClientPubKeyBin, Streams, undefined) of
+            undefined ->
+                %% first time we've communicated with a client. let them know about
+                %% their first state channel, should be updated down the road with
+                %% replacement SC when ChosenSC closes
+                Client = {Stream, erlang:monitor(process, Stream)},
+                ok = broadcast_banner(ChosenSC, [Client]),
+                Client;
```
Vagabond, line 475: You need to track the SCID here as well, so you know which streams need an update.

michaeldjeffrey: Something different than what's in `broadcast_replacement_banner/2`?

```
PubKeyBins =
   lists:map(
    fun(Summary) -> blockchain_state_channel_summary_v1:client_pubkeybin(Summary) end,
    blockchain_state_channel_v1:summaries(ClosedSC)
  ),
StreamsNeedingReplacementSC = maps:with(PubKeyBins, Streams),
```

Vagabond: What happens to clients who never (successfully) used the SC they were bannered?

Vagabond:
This looks ok, although I'd still be inclined to favor tracking the SC association with the stream directly.

src/state_channel/blockchain_state_channels_server.erl
michaeldjeffrey: `gateway_bannered_sc_skip_test` tests for this.
They move into the next available state channel, and receive that banner.

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -955,39 +965,67 @@ update_sc_summary(ClientPubKeyBin, PayloadSize, DCPayloadSize, SC, MaxActorsAllo
...
+            ok;
+        ReplacementSCID ->
+            PubKeyBins =
+                lists:map(
+                  fun(Summary) -> blockchain_state_channel_summary_v1:client_pubkeybin(Summary) end,
+                  blockchain_state_channel_v1:summaries(ClosedSC)
+                 ),
+            StreamsNeedingReplacementSC = maps:with(PubKeyBins, Streams),
+
+            {SC, _Skewed} = maps:get(ReplacementSCID, SCs),
```
vihu, line 986: Any chance this might return `{badkey, _}` ?

michaeldjeffrey: There is not. `ReplacementSCID` comes from the call to `maybe_get_new_active` that only knows how to pull from existing state channels in `#state{}`.

src/state_channel/blockchain_state_channels_server.erl
Vagabond: I don't see how it's tested (from a getting a banner perspective) and I don't see how it would be possible.

Vagabond: Because a client that never successfully used the state channel will not be listed in the summaries

Issue #941, opened 2021-08-08
fail to download snapshot from s3


KillingJacky:
It is observed on some hotspots that the miner fails to download the blessed snapshot from S3, the temporary scratch file `snap-*.scratch` will keep growing, up to a super large size than the actual (~130MB). This issue makes the hotspot stuck on chain syncing. 

```
2021-08-07 07:44:26.391 160 [error] <0.9550.0>@blockchain_worker:start_snapshot_sync:949 snapshot download or loading failed because {error,timeout}: [{blockchain_worker,do_s3_download,2,[{file,"blockchain_worker.erl"},{line,1034}]},{blockchain_worker,'-start_snapshot_sync/4-fun-0-',6,[{file,"blockchain_worker.erl"},{line,936}]}]
2021-08-07 07:44:26.392 160 [info] <0.9550.0>@blockchain_worker:attempt_fetch_p2p_snapshot:955 attempting snapshot sync with "/p2p/11uvSUGnmwtQ6uB1grfrdgrDZnFugrX64oEqJUvWxjAQC9WDJtb"
2021-08-07 07:44:31.468 160 [info] <0.1481.0>@blockchain_worker:handle_info:601 snapshot sync down reason normal
2021-08-07 07:44:31.473 160 [info] <0.1481.0>@blockchain_worker:snapshot_sync:741 snapshot_sync starting <0.11069.0> #Ref<0.169456592.1972371457.214293>
2021-08-07 07:44:31.476 160 [info] <0.11069.0>@blockchain_worker:do_s3_download:1024 Attempting snapshot download from "https://snapshots.helium.wtf/mainnet/snap-953281", writing to scratch file "/var/data/snap/snap-953281.scratch"
...


/opt/miner # ls -lh /var/data/snap
total 601M   
-rw-r--r--    1 root     root      600.6M Aug  7 07:59 snap-953281.scratch

```

For this example, the size of the scratch file already gets 600MB+.

Attached is the result of `grep  start_snapshot_sync` from the recent log of one hotspot
[miner-fail-to-download-snapshot.log](https://github.com/helium/blockchain-core/files/6950230/miner-fail-to-download-snapshot.log)

KillingJacky:
Tried downloading with `wget https://snapshots.helium.wtf/mainnet/snap-953281`, it can succeed. After put the snapshot file to its location and restart miner, the snapshot can be loaded flawlessly.

evanmcc:
should be resolved.

PR #943, opened 2021-08-10
Clean up scratch files when cleaning snap dir

mrallen1:
Problem to solve: Files are only removed from the snapshot directory when they're older than a week. That means largish scratch files can linger for a while and take up space.

Solution: Explicitly remove scratch files which may exist when we clean out the snapshot download folder.


evanmcc:
closes #941

evanmcc (approved):

PR #947, opened 2021-08-16
Introduce listen_addresses parameter to specify listen IP

PaulVMo:
This proposed change introduces the blockchain configuration parameter listen_addresses which replaces 'port' and 'ports' parameters and allows the listen IP as well as port to be specified. This is a request from community members such as @denseone to better control listening IP of multi-homed hosts. Currently, the libp2p swarm always listens on IPv4 address 0.0.0.0 with the tcp port(s) specified in 'port' or 'ports'. 

'listen_addresses' is formatted as a list of multiaddrs (e.g., /ip4/10.0.1.15/tcp/2154) for greater flexibility down the line (e.g., adding transport types, enabling IPv6).

The change deprecates port and ports. A corresponding change is required in miner to change the default config to use listen_addresses and not port or ports in the config file. If port or ports is in the config, they take precedent over listen_addresses. This approach was deliberately chosen to maintain backwards compatibility with any operators that are currently overriding the default port(s). If the listen_addresses parameter took precedent over port and ports, then anyone intentionally overriding ports would see their deployment break.


src/blockchain_worker.erl
```diff
@@ -323,13 +323,24 @@ init(Args) ->
...
+                              undefined ->
+                                  case application:get_env(blockchain, listen_addresses, undefined) of
+                                      undefined -> ["/ip4/0.0.0.0/tcp/0"];
+                                      AddrList when is_list(AddrList) -> AddrList
+                                  end;
+                              Port ->
+                                  lager:warning("Using deprecated port configuration parameter. Swtich to listen_addresses"),
+                                  ["/ip4/0.0.0.0/tcp/" ++ integer_to_list(Port)]
+                          end;
+                      PortList when is_list(PortList) ->
```
madninja, line 340: You're missing the legacy case for when a single port is given (vs a ports list)

madninja (changes requested):
Looks good, almost there

src/blockchain_worker.erl
PaulVMo: Thanks for the review. The single port case is there. Line 330 has the case and the line 342 the condition.

madninja: Ah yes.. I see he ladder case now. Sorry for the misread

madninja (approved):

src/blockchain_worker.erl
```diff
@@ -323,13 +323,24 @@ init(Args) ->
...
+    ListenAddrs = case application:get_env(blockchain, ports, undefined) of
+                      undefined -> 
+                          case application:get_env(blockchain, port, undefined) of
+                              undefined ->
+                                  case application:get_env(blockchain, listen_addresses, undefined) of
+                                      undefined -> ["/ip4/0.0.0.0/tcp/0"];
+                                      AddrList when is_list(AddrList) -> AddrList
+                                  end;
+                              Port ->
+                                  lager:warning("Using deprecated port configuration parameter. Swtich to listen_addresses"),
```
Vagabond, line 337:
```suggestion
                                  lager:warning("Using deprecated port configuration parameter. Switch to listen_addresses"),
```

src/blockchain_worker.erl
```diff
@@ -323,13 +323,24 @@ init(Args) ->
...
-            end,
+    %% Get list of listening addresses. If deprecated 'ports' or 'port' variable used,
+    %% assume listening IP of 0.0.0.0. otherwise use listen_addresses parameter.
+    ListenAddrs = case application:get_env(blockchain, ports, undefined) of
+                      undefined -> 
+                          case application:get_env(blockchain, port, undefined) of
+                              undefined ->
+                                  case application:get_env(blockchain, listen_addresses, undefined) of
+                                      undefined -> ["/ip4/0.0.0.0/tcp/0"];
+                                      AddrList when is_list(AddrList) -> AddrList
```
Vagabond, line 334: one thing to watch out for here; if the user supplies a "bare" multiaddr that is not a list, the multi addr string will count as a list.

A good way to differentiate is to check if the head of the list is itself a list.

PaulVMo: Thanks. I added a case to cover this scenario of a multiaddr that is not in a list by checking if the head is a list, it not assume it a a bare string/multiaddr.

PR #952, opened 2021-08-19
Add `has_witnesses` check to `filter_interactive_gws` 

anthonyra:
# The Issue

With the current system it's possible to adversely affect another hotspot by remotely asserting one hotspot directly ontop of a target hotspot. This results in the ability to lower the target hotspots transmit reward scale by a factor equal to the number of hotspots used to target the area.

This PR will not permenantly fix the issue if someone was determined enough to attempt this while being close to the target hotspots location. However, this will help protect against the scenario where the hotspot is located in a different state/country and would be still able to adversely affect the transmit reward scale of the target.

This could allow "bad actors" the ability to reduce the top performing hotspots on the network by 50% while not doing anything aside from remotely asserting cannon fodder to do it.

# The Proof

After reviewing the code, I noticed that it could be possible for a hotspot that is physically located in a different geographical location than it's `assert_location` could be considered interactive due to the fact that the `last_poc_challenge` variable was used to validate this state. The issue is that `last_poc_challenge` is updated on the creation of a `poc_request` which is all done network side as long as the hotspot meets 3 requirements. 1) Has an asserted location, 2) is synced to the blockchain, 3) and is outside it's `poc_interval` number of blocks from their last submission.

To put this theory to practice;
1) I was able to take 2 hotspots one with an antenna attached and one without an antenna and asserted them next to my lone wolf hotspot. These two hotspots were physically located in a completely different state, the one with the antenna was still collecting witnesses (zero of these witnesses being valid), but both still were considered interactive in regards to HIP17.
2) After the hotspots `assert_locations` were accepted and they met the above 3 conditions they successfully scaled my lone wolf hotspot from a transmit reward scale of 1.0 => 0.33.

# The Solution

Upon further review of the code and how the blockchain handles current witnesses and PoC receipts, I decided the best bandaid for now would be to add an additional check to ensure that to be considered an interactive hotspot, it has to have valid witnesses. The current system wipes the witnesses upon `assert_locations` and they are only added to the witness list if they are valid in regards to RSSI/SNR. This will ensure that a hotspot wouldn't be considered interactive unless they've recently created a PoC request and have valid witnesses for that location.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -1549,7 +1549,7 @@ find_gateway_has_witnesses(Address, Ledger) ->
                 {ok, BinGw} ->
                     Gw = blockchain_ledger_gateway_v2:deserialize(BinGw),
                     Witnesses = blockchain_ledger_gateway_v2:witnesses(Gw),
-                    HasWitnesses = map_size(Witnesses) > 0, %% Not sure if this "waterfall" is the right way todo this
+                    HasWitnesses = map_size(Witnesses) > 1, %% Not sure if this "waterfall" is the right way todo this
                     {ok, HasWitnesses};
                 not_found ->
                     {error, not_found};
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -55,6 +55,7 @@
     find_gateway_location/2,
     find_gateway_owner/2,
     find_gateway_last_challenge/2,
+    find_gateway_has_witnesses/2, %% I'm assuming this is needed since blockchain_ledger_gateway_v2:witnesses/1 takes GW not GWAddr
     %% todo add more here
 
     add_gateway/3, add_gateway/4, add_gateway/6,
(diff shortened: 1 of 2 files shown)


evanmcc:
this will need a chain variable as a feature flag.

anthonyra:
> this will need a chain variable as a feature flag.

Like a brand new chain variable or would a version bump work? I thought it was going to need a chain variable but since the only effects, at least to my knowledge, the table that is create each an every time when creating the reward block than is deleted on the next that it might not need it? Unless it's the `blockchain_ledger_v1` bit that's causing the new chain var...

anthonyra:
@evanmcc I took a look at the PR that you recently submitted for HIP17 improvements and I think they fall under the same feature so I used the same chain var feature flag you did to select between `blockchain_hex_v2` and `blockchain_hex_v1`... let me know if this is a good approach or if I need to go a little bit higher up with it. Thank  you for any and all help!

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1533,6 +1538,26 @@ find_gateway_last_challenge(Address, Ledger) ->
             end
     end.
 
+find_gateway_has_witnesses(Address, Ledger) ->
```
mattstabeler, line 1541: Does this account for invalid witnesses also? 
I can't quite tell if it needs to do something similar to this function: 
https://github.com/helium/blockchain-core/blob/61acf888b87c9c067c0b490bbdf858911bada02f/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl#L1439

anthonyra: This is based on the witness list that is built on the miner. That list already is filtered for invalid witnesses and is reset on `assert_location` which means that only valid witnesses should be counted atleast that's how I intended it...

mattstabeler:
Is this something that's ready for review @Vagabond  @evanmcc - or does it need more work? :)

mattstabeler:
Any thoughts on this @Vagabond @evanmcc @andymck  ? :) Perhaps some review comments, or notes on how to change it, to get it adopted? 
It seems to ✅ all the boxes in the code contribution guide

PR #955, opened 2021-08-19
HIP 17 performance and correctness update

evanmcc:
HIP17 as currently implemented is very, very slow.  It's also dramatically sensitive to starting location.  I noticed that if one is to calculate the `unclipped density` of a level from the `unclipped density` of the level below it, instead of the `clipped density` as is done currently, the start location sensitivity goes away and makes the entire process amenable to bulk precalcuation of the entire set of clipped and unclipped densities, which ends up being quite a lot faster and produced identical results to the existing implementation with unclipped densities fixed.


evanmcc:
After more HIP close reads, my take here about the bug is wrong.  it does seem like I can adapt this to the correct algorithm, so my take about initial location sensitivity was likely wrong as well.  New PR soon.

PR #962, opened 2021-08-24
Log for absorb_and_commit height

syuan100:
Quick log edit for `absorb_and_commit` to log block height. Tested on blockchain-node and logs come through as expected.

Example:
```
2021-08-24 10:24:33.491 [info] <0.1286.0>@blockchain_txn:absorb_and_commit:436 took 4999 ms for block height 975934
```


syuan100:
@evanmcc fwiw decided to do a quick check on the code path for `absorb_and_commit` to see if we could nab a block height after reading the chat in the validator discord channel. Seemed like a quick fix so just went ahead and did it.

evanmcc (approved):

PR #963, opened 2021-08-24
try and speed up syncing

evanmcc:
Requests and Receipts are slower than they need to be, because they're not taking advantage of information that we already have.  This code attempts to use that information.  Two bits:

1. For requests, which are fetching an entire block to get the height, we add a reverse height->hash lookup to the blocks store, with a fallback for older blocks.
2. For receipts, although we already have the block fetched for other reasons, we don't re-hash it, since as blocks get larger that gets more and more expensive, but instead fetch it from the block store.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -1,3 +1,4 @@
+
 %%%-------------------------------------------------------------------
 %% @doc
 %% == Blockchain ==
@@ -15,7 +16,9 @@
     ledger/0, ledger/1, ledger/2, ledger_at/2, ledger_at/3,
     dir/1,
 
-    blocks/1, get_block/2, get_block_hash/2, get_raw_block/2, save_block/2,
+    blocks/1,
+    get_block/2, get_block_hash/2, get_block_height/2, get_raw_block/2,
+    save_block/2,
     has_block/2,
     find_first_block_after/2,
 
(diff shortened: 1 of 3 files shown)


src/blockchain.erl
```diff
@@ -636,6 +639,23 @@ get_block_hash(Height, #blockchain{db=DB, heights=HeightsCF}) ->
             Error
     end.
 
+get_block_height(Hash, #blockchain{db=DB, heights=HeightsCF, blocks=BlocksCF}) ->
```
vihu, line 642:
```suggestion
-spec get_block_height(Hash :: blockchain_block:hash(), Blockchain :: blockchain()) -> {ok, non_neg_integer()} | {error, any()}.
get_block_height(Hash, #blockchain{db=DB, heights=HeightsCF, blocks=BlocksCF}) ->
```

src/blockchain.erl
```diff
@@ -636,6 +639,23 @@ get_block_hash(Height, #blockchain{db=DB, heights=HeightsCF}) ->
...
+    case rocksdb:get(DB, HeightsCF, Hash, []) of
+       {ok, <<Height:64/integer-unsigned-big>>} ->
+            {ok, Height};
+        not_found ->
+            case rocksdb:get(DB, BlocksCF, Hash, []) of
+                {ok, BinBlock} ->
```
Vagabond, line 648: Worth writing the hash lookup back to rocks here?

Vagabond:
Overall this looks good to merge once the review comments are addressed.

PR #965, opened 2021-08-24
Remove unnecessary h3_in_region work

vihu:


vihu:
Nope, this is circular, does not work

PR #971, opened 2021-08-26
fix tests

andymck:


evanmcc:
I think that this is too stale, the issues that it seems to have fixed may have been resolved by other PRs.

PR #979, opened 2021-09-01
More receipts/requests performance work

evanmcc:
This needs a lot of cleanup, and needs some additional snapshot work done, hence the WIP status, but:
- adds a pid-local cache for deserialized hex list per height
- doesn't roundtrip the hex list through a map, so we don't need to resort the resulting map
- adds a block info index section, which allows us to fetch and deserialize many fewer blocks per validation (this is what needs to be added to the snapshot so we can grab fewer blocks)
- adds mode to the denormalized gateway fields and moves most code to using that
- gets rid of the gateway cache finally


src/blockchain.erl
```diff
@@ -659,6 +661,49 @@ get_block_height(Hash, #blockchain{db=DB, heights=HeightsCF, blocks=BlocksCF}) -
...
+                    ok = rocksdb:put(DB, InfoCF, <<Height:64/integer-unsigned-big>>, term_to_binary(Info), []),
+                    {ok, Info};
+                Error ->
+                    Error
+            end;
+        Error ->
+            Error
+    end.
+
+mk_info(Hash, Block) ->
```
Vagabond, line 691: Might be better called mk_block_info, or relocated to a blockchain_block module

Vagabond:
Overall this looks good.

PR #981, opened 2021-09-02
Emit add_block events for the aux ledger instead

Vagabond:
It would be useful to be able to run common chain consuming applications (router, ETL, node, etc) against the aux ledger, to gauge the impact of changes to the ledger or chain vars. This patch set enables that feature as well as fixes a bug where the ledger returned from new_snapshot did not have the snapshot set.


src/transactions/v1/blockchain_poc_path_element_v1.erl
```diff
@@ -100,8 +100,9 @@ to_json(Elem, Opts) ->
                 [{W, []} || W <- witnesses(Elem)];
             {tagged_witnesses, TaggedWitnesses} ->
                 lists:map(fun
-                              ({true, _, W}) ->
-                                  {W, [{is_valid, true}]};
+                              ({true, ValidReason, W}) ->
```
vihu, line 103: Note: Revert this back

vihu:
Reminder to revert d3a5f23

vihu:
Already merged. Closing.

PR #983, opened 2021-09-07
Defer txn rejections from the future until it is our present

xandkar:


include/blockchain.hrl
```diff
@@ -6,14 +6,16 @@
 -define(SYNC_PROTOCOL_V2, "blockchain_sync/1.2.0").
 -define(FASTFORWARD_PROTOCOL_V2, "blockchain_fastforward/1.2.0").
 
+-define(TX_PROTOCOL_V1, "blockchain_txn/1.0.0").
+-define(TX_PROTOCOL_V2, "blockchain_txn/2.0.0").
```
Vagabond, line 10: This isn't really semver, so I'm not sure we need to go to 2.0.0?

src/handlers/blockchain_txn_handler.erl
```diff
@@ -27,47 +29,79 @@
...
-    {ok, #state{callback = Callback}}.
+init(client, _Conn, [Path, Parent, TxnHash]) ->
+    {ok, #state{parent=Parent, txn_hash=TxnHash, path=Path}};
+init(server, _Conn, [Path0, _Parent, Callback] = _Args) ->
+    lager:warning("init as server. Path0: ~p", [Path0]),
+    Path =
+        case Path0 of
+            ?TX_PROTOCOL_V1 -> ?TX_PROTOCOL_V1;
+            ?TX_PROTOCOL_V2 -> ?TX_PROTOCOL_V2;
+            "" -> % TODO Why do we get an empty protocol path/version?
```
Vagabond, line 56: I don't think this should happen, and if it does we should not wall paper over it.

src/transactions/blockchain_txn_dialer.erl
```diff
@@ -111,3 +109,57 @@ terminate(_Reason, _State) ->
...
     {ok, State}.
 
+%% Internal ===================================================================
+
+-spec dial_(#state{}) -> ok | {error, _}.
+dial_(#state{member=Member, txn_key = TxnKey, txn=Txn, parent=Parent, timeout=Ref}) ->
+    Swarm = blockchain_swarm:swarm(),
+    P2PAddress = libp2p_crypto:pubkey_bin_to_p2p(Member),
+    TxnHash = blockchain_txn:hash(Txn),
+    (fun
```
Vagabond, line 119: I thought for sure you could supply a list of protocols to dial and libp2p would figure it out for you, but apparently not, ugh.

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -207,11 +218,54 @@ handle_info({accepted, {Dialer, TxnKey, Txn, Member}}, State) ->
...
+    lager:debug(
+        "txn: ~s, rejected_by: ~p, Dialer: ~p,"
+        "my height: ~p, rejector height: ~p",
+        [
+            blockchain_txn:print(Txn), Member, Dialer,
+            CurBlockHeight, RejectorHeight
+        ]
+    ),
+    S1 =
+        case {} of
```
Vagabond, line 247: What? Why not do case `RejectorHeight > CurBlockHeight`, I don't think we need to further distinguish between `RejectorHeight < CurBlockHeight` and `RejectorHeight == CurBlockHeight`. Also if you must do this then perhaps Erlang's `if` is a better fit here?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -274,12 +328,38 @@ handle_add_block_event({add_block, BlockHash, Sync, _Ledger}, State=#state{chain
...
+        "Processing deferred rejections. "
+        "Count now current: ~b, count still deferred: ~b",
+        [length(Current), length(Deferred1)]
+    ),
+    Reject =
+        fun ({Dialer, TxnKey, Txn, Member, _}) ->
+            ok = rejected(TxnKey, Txn, Member, Dialer, CurBlockHeight, RejectF)
+        end,
+    lists:foreach(Reject, Current),
+    S#state{rejections_deferred=[]}.
```
Vagabond, line 361:
```suggestion
    S#state{rejections_deferred=Deferred1}.
```

include/blockchain.hrl
xandkar: Ah, so are there any rules or just arbitrary increment?

src/handlers/blockchain_txn_handler.erl
```diff
@@ -27,47 +29,79 @@
...
 %% libp2p_framed_stream Function Definitions
 %% ------------------------------------------------------------------
-init(client, _Conn, [Parent, TxnHash]) ->
-    {ok, #state{parent=Parent, txn_hash=TxnHash}};
-init(server, _Conn, [_Path, _Parent, Callback] = _Args) ->
-    {ok, #state{callback = Callback}}.
+init(client, _Conn, [Path, Parent, TxnHash]) ->
+    {ok, #state{parent=Parent, txn_hash=TxnHash, path=Path}};
+init(server, _Conn, [Path0, _Parent, Callback] = _Args) ->
+    lager:warning("init as server. Path0: ~p", [Path0]),
```
xandkar, line 51: TODO Remove this warning.

src/handlers/blockchain_txn_handler.erl
```diff
@@ -27,47 +29,79 @@
...
 client(Connection, Args) ->
     libp2p_framed_stream:client(?MODULE, Connection, Args).
 
-server(Connection, _Path, _TID, Args) ->
+server(Connection, Path, _TID, Args) ->
     %% NOTE: server/4 in the handler is never called.
     %% When spawning a server its handled only in libp2p_framed_stream
-    libp2p_framed_stream:server(?MODULE, Connection, [_Path | Args]).
+    %% TODO If never called, as per above, why do we need server/4?
+    libp2p_framed_stream:server(?MODULE, Connection, [Path | Args]).
```
Vagabond, line 43: I don't think you want to do this, confusingly. and do what I suggested over in the miner PR instead.

src/transactions/blockchain_txn_mgr.erl
xandkar: I did use `if` in earlier commit, but remembered that Evan is against any uses of it, so changed to this oddity.

src/transactions/blockchain_txn_mgr.erl
xandkar: :man_facepalming:

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -207,11 +246,75 @@ handle_info({accepted, {Dialer, TxnKey, Txn, Member}}, State) ->
...
+            Age when Age < 0 ->
+                %% TODO Maybe limit how far in the future?
+                lager:warning(
+                    "Received txn rejection from the future. Deferring: ~p",
+                    [Rejection]
+                ),
+                ordsets:add_element(Rejection, Deferred0);
+            %% present or recent past:
+            Age when (Age >= 0) andalso (Age < MaxRejectionAge) ->
+                if
```
andymck, line 288: nitpicking but you could get rid of the if by pushing the logging out to a function and decide which output to perform via guards

andymck (commented):
lgtm

One main comment i added is regarding the new txn handler protocol and its error msg convention.  My suggestion may be outside of the scope of this PR and if so ignore it...but i feel as we are adding a new protocol it would be good to time to address this and get better error msgs going to the txn mgr and in such a way accommodates generic error which could be added in the future....

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -21,6 +21,12 @@
          make_ets_table/0
         ]).
 
+%% Testing backdoors
+-export([
```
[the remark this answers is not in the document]
xandkar: Good call!

src/handlers/blockchain_txn_handler.erl
```diff
@@ -27,47 +29,72 @@
...
     Parent ! {blockchain_txn_response, {ok, TxnHash}},
     {stop, normal, State};
 handle_data(client, <<"no_group">>, State=#state{parent=Parent, txn_hash=TxnHash}) ->
     Parent ! {blockchain_txn_response, {no_group, TxnHash}},
     {stop, normal, State};
-handle_data(client, <<"error">>, State=#state{parent=Parent, txn_hash=TxnHash}) ->
-    Parent ! {blockchain_txn_response, {error, TxnHash}},
+handle_data(
+    client,
+    <<"error", _/binary>> = ErrorMsg,
```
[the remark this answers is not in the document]
xandkar: It seems like a good idea in general, but probably too big of a can of worms for the scope of this PR. @Vagabond are there any concerns about message size for this protocol?

src/transactions/blockchain_txn_mgr.erl
xandkar: So I can see doing something like this, but I'm worried that it leaks context. This log message only makes sense where it was and now has the potential to become stale as the parent/caller code evolves. Namely the "Counting" part - it talks about an action that is _assumed_ to be been taken by the caller of this function. If that ever changes, this message has a high chance of being misleading.

```erlang
-spec maybe_log_old_rejection(rejection(), integer(), integer()) -> ok.
maybe_log_old_rejection(Rejection, Age, Threshold) when Age > 0, Age =< Threshold ->
    lager:debug(
        "Received txn rejection from the past. From ~b blocks ago. Counting: ~p",
        [Age, Rejection]
    );
maybe_log_old_rejection(Rejection, Age, Threshold) when Age > Threshold ->
    lager:warning(
        "Received txn rejection from older, but still acceptable past. "
        "From ~b blocks ago. Counting: ~p",
        [Age, Rejection]
    );
maybe_log_old_rejection(_, _, _) ->
    ok.
```

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -154,6 +177,22 @@ handle_cast(_Msg, State) ->
...
     {noreply, State}.
 
+handle_call(
+    force_process_cached_txns,
+    _,
+    #state{
+        chain = Chain,
+        cur_block_height = CurBlockHeight,
+        submit_f = SubmitF
+    }=S
```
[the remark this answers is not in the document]
xandkar: Done.

PR #919, opened 2021-07-14
add proper witness limits

evanmcc:
this allows us to bring witnesses back at some point, but we'll need to wait until everyone has accepted the var before we activate it.

PR #985, opened 2021-09-09
optimize block loading by not validating witnesses

BenoitDuffez:
The function blockchain_ledger_v1:insert_witnesses has been emptied
with:
https://github.com/helium/blockchain-core/commit/b69a5d945468348e8ec4d73c437f7b42be08dc4b

When loading the blocks/txn/pocs/paths, the miner inspects the valid
paths only to send them to blockchain_ledger_v1:insert_witnesses.
Since this is no longer doing anything, the path validation is useless
too.

This proposes to remove the whole path validation/insert_witnesses, and
also removes dead code.

### DIFF
--- a/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
+++ b/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
@@ -703,31 +703,6 @@ tagged_path_elements_fold(Fun, Acc0, Txn, Ledger, Chain) ->
     end.
 
 
-%% Iterate over all poc_path elements and for each path element calls a given
-%% callback function with the valid witnesses and valid receipt.
-valid_path_elements_fold(Fun, Acc0, Txn, Ledger, Chain) ->
-    Path = ?MODULE:path(Txn),
-    try get_channels(Txn, Chain) of
-        {ok, Channels} ->
-            lists:foldl(fun({ElementPos, Element}, Acc) ->
-                                {PreviousElement, ReceiptChannel, WitnessChannel} =
-                                case ElementPos of
-                                    1 ->
-                                        {undefined, 0, hd(Channels)};
-                                    _ ->
-                                        {lists:nth(ElementPos - 1, Path), lists:nth(ElementPos - 1, Channels), lists:nth(ElementPos, Channels)}
-                                end,
-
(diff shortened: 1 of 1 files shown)


Vagabond:
I don't know we want to do this, as I think we want to bring witnesses back in the future.

BenoitDuffez:
This was just a continuation of the linked commit (that commented out insert_witnesses).

My edit is a bit too harsh, removing poc version < 9, but in the current state the witnesses are parsed for nothing. Perhaps comment out the witness sort/insert when ingesting a block/txn, and restore when you want to put this feature back in?

By the way, bringing the witnesses back in the future: is this for all miners or just validators? I'm a bit confused as I thought the idea was to remove features from the gateway miners in order to move them to the validator miners.

evanmcc:
We can't do this; witness validity matters for rewards, and this ignores it entirely.  Vote to close.

BenoitDuffez:
My question then is about the purpose of the block of code that I removed. In my opinion it does:

* absorb: if poc >= v9: valid_path_elements_fold
* valid_path_elements_fold: calculate valid_receipt & valid_witnesses
* valid_receipt and valid_witnesses are passive checks: they filter a list based on the validity (nothing persisted except in RAM)
* the filtered list is then sent to blockchain_ledger_v1:insert_witnesses
* blockchain_ledger_v1:insert_witnesses does nothing, it's commented out

So I don't understand why we validate receipts/witnesses in RAM and dump that list into an empty function?

BenoitDuffez:
I have changed how it's done so that the witness validation is skipped only in follow mode. I think it makes more sense, is less agressive, more compatible with existing miners on the chain and compatible with the chain history.

This is more in line with what I had initially proposed, could you review and give me your opinion?  
Thanks

evanmcc:
I guess I'm confused.  This validation should already be skipped in follow mode.  Is it still getting called somehow?

BenoitDuffez:
Enable debug traces in follow mode, you'll see that the miner will validate all witnesses and then dump the computation result.

evanmcc (commented):
I need to think about this some more.  it's not really the right thing to do, but I'm not sure how to do it "right".

evanmcc:
Closing this in favor of #998 + #919 (the latter eventually).  Thank you for bringing this to my attention!

PR #987, opened 2021-09-10
election v6 using lists

evanmcc:
ideally this would get a commit fixing the tests before we do a wide release.


Vagabond:
I approved this, for the record.

PR #988, opened 2021-09-10
Pevm/fix ct again

evanmcc:
reward suite still sad, some consistent failures in the state channel stuff


evanmcc:
maybe just get these in, I think maybe reward suite will be better now.

vihu (approved):

Issue #978, opened 2021-09-01
Simplify transfer_hotspot transaction


cokes518:
Current implementation of `transfer_hotspot` has deep dependencies on wallet-api and makes it hard for us to compartmentalize these new features for future work. 

We'd like to remove wallet-api from the equation and further simplify the user experience by removing liveness check and the amount as required fields for the transaction to be processed.

Proposed:
replace `transfer_hotspot_v1` with `transfer_hotspot_v2` that takes the following arguments:
* `owner_signature`
* `owner_address`
* `gateway_address`
* `new_owner_address`

Once submitted to the chain, the owner changes. Hotspot location, gain, elevation do not change.

madninja:
I guess one thing to take into consideration is the effect of scammers selling dead hotspots. Perhaps that's fine but it's putting the responsibility solidly on the buyer here to somehow confirm that this gateway is not a brick

cokes518:
right, but the chain shouldn't be the one to govern/be the decision maker.

madninja:
> right, but the chain shouldn't be the one to govern/be the decision maker.

I agree.. very much.. just wanted it to be very clear

abhay:
Message received and I think we can make sure the UX makes that clear as well.

PR #992, opened 2021-09-13
Add support for transfer_hotspot_v2 txn

vihu:
Support for https://github.com/helium/blockchain-core/issues/978

TODO:

- [x] Testing
- [x] Cleanup and final edits
- [x] Merge https://github.com/helium/proto/pull/91 and switch proto
- [x] Merge https://github.com/helium/proto/pull/95 and switch proto

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -516,3 +516,7 @@
 %% they do, we will re-evaluate accordingly.
 %%
 %% Refer to blockchain_region_SUITE for an example
+
+%% transfer hotspot transaction version
+%% Allowed values: 2
+-define(transfer_hotspot_txn_version, transfer_hotspot_txn_version).
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -517,6 +517,10 @@
 %%
 %% Refer to blockchain_region_SUITE for an example
 
-%% transfer hotspot transaction version
-%% Allowed values: 2
--define(transfer_hotspot_txn_version, transfer_hotspot_txn_version).
+%% ------------------------------------------------------------------
+%% This is a general purpose variable. We can use it for any transaction
+%% to specify which version of it we will support on chain.
+%%
+%% As an example, blockchain_txn_transfer_hotspot_v2, will get enabled
+%% when this value is set to >= 2.
+-define(transaction_validity_version, transfer_hotspot_txn_version).    % pos_integer
--- a/src/transactions/v1/blockchain_txn_vars_v1.erl
+++ b/src/transactions/v1/blockchain_txn_vars_v1.erl
@@ -1223,11 +1223,10 @@ validate_var(?density_tgt_res, Value) ->
 validate_var(?hip17_interactivity_blocks, Value) ->
     validate_int(Value, "hip17_interactivity_blocks", 1, 5000, false);
 
-validate_var(?transfer_hotspot_txn_version, Value) ->
+validate_var(?transaction_validity_version, Value) ->
     case Value of
-        %% Edit this condition to enable future transfer hotspot txn version updates
-        N when N == 2 -> ok;
-        _ -> throw({error, {invalid_transfer_hotspot_txn_version, Value}})
+        2 -> ok;
+        _ -> throw({error, {invalid_transaction_validity_version, Value}})
     end;
 validate_var(?assert_loc_txn_version, Value) ->
     case Value of
--- a/src/transactions/v1/blockchain_txn_vars_v1.erl
+++ b/src/transactions/v1/blockchain_txn_vars_v1.erl
@@ -1223,6 +1223,12 @@ validate_var(?density_tgt_res, Value) ->
 validate_var(?hip17_interactivity_blocks, Value) ->
     validate_int(Value, "hip17_interactivity_blocks", 1, 5000, false);
 
+validate_var(?transfer_hotspot_txn_version, Value) ->
+    case Value of
+        %% Edit this condition to enable future transfer hotspot txn version updates
+        N when N == 2 -> ok;
+        _ -> throw({error, {invalid_transfer_hotspot_txn_version, Value}})
+    end;
 validate_var(?assert_loc_txn_version, Value) ->
     case Value of
         N when is_integer(N), N >= 1, N =< 2 -> ok;
--- a/src/transactions/v2/blockchain_txn_transfer_hotspot_v2.erl
+++ b/src/transactions/v2/blockchain_txn_transfer_hotspot_v2.erl
@@ -1,0 +1,263 @@
+-module(blockchain_txn_transfer_hotspot_v2).
+-behavior(blockchain_txn).
+-behavior(blockchain_json).
+
+-include("blockchain_json.hrl").
+-include("blockchain_utils.hrl").
+-include("blockchain_txn_fees.hrl").
+-include("blockchain_vars.hrl").
+-include_lib("helium_proto/include/blockchain_txn_transfer_hotspot_v2_pb.hrl").
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-export([
+    new/3,
+    gateway/1,
+    owner/1,
+    owner_signature/1,
+    new_owner/1,
+    fee/2, fee/1,
+    fee_payer/2,
+    calculate_fee/2, calculate_fee/5,
+    hash/1,
+    sign/2,
+    is_valid/2,
+    is_valid_owner/1,
+    absorb/2,
+    print/1,
+    json_type/0,
+    to_json/2
+]).
+
+-type txn_transfer_hotspot_v2() :: #blockchain_txn_transfer_hotspot_v2_pb{}.
+-export_type([txn_transfer_hotspot_v2/0]).
+
+-spec new(
+    Gateway :: libp2p_crypto:pubkey_bin(),
+    Owner :: libp2p_crypto:pubkey_bin(),
+    NewOwner :: libp2p_crypto:pubkey_bin()
+) -> txn_transfer_hotspot_v2().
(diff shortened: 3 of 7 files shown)


src/transactions/v2/blockchain_txn_transfer_hotspot_v2.erl
```diff
@@ -0,0 +1,263 @@
...
+is_valid(
+    #blockchain_txn_transfer_hotspot_v2_pb{
+        owner_address = Owner,
+        new_owner_address = NewOwner
+    } = Txn,
+    Chain
+) ->
+    Ledger = blockchain:ledger(Chain),
+    AreFeesEnabled = blockchain_ledger_v1:txn_fees_active(Ledger),
+    Conditions = [
```
vihu, line 143: Needs close look to check if we want more validations here

src/transactions/v2/blockchain_txn_transfer_hotspot_v2.erl
```diff
@@ -0,0 +1,263 @@
...
+    Fee = ?MODULE:fee(Txn),
+    Hash = ?MODULE:hash(Txn),
+
+    {ok, GWInfo} = blockchain_ledger_v1:find_gateway_info(Gateway, Ledger),
+    %% fees here are in DC
+    case blockchain_ledger_v1:debit_fee(NewOwner, Fee, Ledger, AreFeesEnabled, Hash, Chain) of
+        {error, _Reason} = Error ->
+            Error;
+        ok ->
+            NewGWInfo = blockchain_ledger_gateway_v2:owner_address(NewOwner, GWInfo),
```
vihu, line 166: Check if we need to update any gateway specific nonce

cokes518:
will this be gated behind a chain var? app and CLI will need to update

vihu:
> will this be gated behind a chain var? app and CLI will need to update

I don't _think_ it needs to be? We can presumably just have a release at some point with support for v2. Thoughts @Vagabond?

cokes518:
> > will this be gated behind a chain var? app and CLI will need to update
> 
> I don't _think_ it needs to be? We can presumably just have a release at some point with support for v2. Thoughts @Vagabond?

It will break the app no?

vihu:
> > > will this be gated behind a chain var? app and CLI will need to update
> > 
> > 
> > I don't _think_ it needs to be? We can presumably just have a release at some point with support for v2. Thoughts @Vagabond?
> 
> It will break the app no?

True yes. We'll have to do the ETL side work in parallel. I guess we can introduce a new var? `transfer_hotspot_txn_version` and set it accordingly?

anthonyra:
I like the idea of making this simple but thinking about this in real world scenario does it make it too simple?

1) This removes the payment option of a transfer which could be fine since v1 will remain active on the chain. With a Ebay auction the buyer could be more protected than a person to person interaction. However, this should be very clear in the UX/UI...
2) About a month ago there was an individual who had his hardware transferred from his wallet due to it being compromised. In doing so the attacker essentially bricked that miner. The only thing that saved his second miner was the fact it wasn't fully synced during the attack...

I understand it's the owners reasonability to protect those keys however, I think when it comes to hardware there should be more checks and balances. As a hotspot owner, I think it would be cool to have the option to toggle additional security over my hardware. Not everyone will want to but at least they had the option.

I'm not sure if this feature has a certain amount of priority to it's release but if there was a v2 of `add_gateway` that allowed the owner to add it with more security that could be checked in this transaction type it would require the gateway to sign the transaction. Meaning who ever performs this atomic transfer would need to have physical access to the gateway and it can only sign if it good health (1200 blocks within sync for example).

cokes518:
> I like the idea of making this simple but thinking about this in real world scenario does it make it too simple?
> 
> 1. This removes the payment option of a transfer which could be fine since v1 will remain active on the chain. With a Ebay auction the buyer could be more protected than a person to person interaction. However, this should be very clear in the UX/UI...

v1 will be superseded by v2. Hotspot transfer payments can occur off chain in fiat or with other crypto currencies. 

> 2. About a month ago there was an individual who had his hardware transferred from his wallet due to it being compromised. In doing so the attacker essentially bricked that miner. The only thing that saved his second miner was the fact it wasn't fully synced during the attack...

Bad actors will be bad actors. It's always up to the buyer to do their due diligence (check explorer for activity, put payment in an HTLC transaction, etc) and make sure they receive a working product.
 
> I understand it's the owners reasonability to protect those keys however, I think when it comes to hardware there should be more checks and balances. As a hotspot owner, I think it would be cool to have the option to toggle additional security over my hardware. Not everyone will want to but at least they had the option.

This is not up to the blockchain to decide and check. The old transfer hotspot logic will never work for data-only Hotspots since they will never PoC and fail the checks of transfer_hotspot_v1. The new light hotspots will also never have to sync the chain and would fail the checks of transfer_hotspot_v1.

> I'm not sure if this feature has a certain amount of priority to it's release but if there was a v2 of `add_gateway` that allowed the owner to add it with more security that could be checked in this transaction type it would require the gateway to sign the transaction. Meaning who ever performs this atomic transfer would need to have physical access to the gateway and it can only sign if its good health (1200 blocks within sync for example).

This is out of scope for this PR. If you have a suggestion for a new transaction type, please propose in a HIP.

src/transactions/v2/blockchain_txn_transfer_hotspot_v2.erl
```diff
@@ -0,0 +1,263 @@
...
+
+-type txn_transfer_hotspot_v2() :: #blockchain_txn_transfer_hotspot_v2_pb{}.
+-export_type([txn_transfer_hotspot_v2/0]).
+
+-spec new(
+    Gateway :: libp2p_crypto:pubkey_bin(),
+    Owner :: libp2p_crypto:pubkey_bin(),
+    NewOwner :: libp2p_crypto:pubkey_bin()
+) -> txn_transfer_hotspot_v2().
+new(Gateway, Owner, NewOwner) ->
```
vihu, line 42: I don't _think_ this transaction needs a nonce of any kind since we've removed the buyer <=> seller payment system here. Replay attacks are probably not a concern as once a transfer_v2 appears on chain the previous owner would not own the hotspot and the next transaction would be invalidated.

cc: @Vagabond

anthonyra:
> Bad actors will be bad actors. It's always up to the buyer to do their due diligence (check explorer for activity, put payment in an HTLC transaction, etc) and make sure they receive a working product.

You have a lot more faith in others than I do.. maybe for good

Just some thoughts and it's interesting how the blockchain went from pretty decent protection (gateway signature asserts) to ... you better do your best to keep your things safe

src/transactions/v2/blockchain_txn_transfer_hotspot_v2.erl
Vagabond: No, because if the hotspot was sent back, the original transaction would be valid again.

Vagabond: You should have a test like this:

A transfers gateway to B
B transfers gateway back to A

At this point the original transaction should *not* be valid, or replayable.

src/transactions/v2/blockchain_txn_transfer_hotspot_v2.erl
Vagabond: I think we do, see previous comments.

Vagabond:
We do need a chain var for this to prevent a transaction of this type being submitted when we aren't fully upgraded.

src/transactions/v2/blockchain_txn_transfer_hotspot_v2.erl
vihu: Updated ed520d9eca378bc358d3579c1007b62f671cb7a7

vihu:
> will this be gated behind a chain var? app and CLI will need to update

Introduced new chain variable in ed520d9eca378bc358d3579c1007b62f671cb7a7.
`transfer_hotspot_txn_version` only allowed value is `2` which we'll need to set.

src/transactions/v2/blockchain_txn_transfer_hotspot_v2.erl
vihu: Added in ed520d9eca378bc358d3579c1007b62f671cb7a7

include/blockchain_vars.hrl
```diff
@@ -516,3 +516,7 @@
 %% they do, we will re-evaluate accordingly.
 %%
 %% Refer to blockchain_region_SUITE for an example
+
+%% transfer hotspot transaction version
+%% Allowed values: 2
+-define(transfer_hotspot_txn_version, transfer_hotspot_txn_version).
```
Vagabond, line 522: Perhaps it's better to have a general value to describe which txns are valid?

Vagabond: like `transaction_validity_set_version` or something

vihu: Updated aabe16e

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1223,6 +1223,12 @@ validate_var(?density_tgt_res, Value) ->
 validate_var(?hip17_interactivity_blocks, Value) ->
     validate_int(Value, "hip17_interactivity_blocks", 1, 5000, false);
 
+validate_var(?transaction_validity_version, Value) ->
+    case Value of
+        %% For future use allow/increment more values of N
+        N when N == 2 -> ok;
```
jadeallenx, line 1229: why not just match on `2` by itself - you don't use the binding at all afaict

jadeallenx (approved):
in general, looks good!

src/transactions/v1/blockchain_txn_vars_v1.erl
vihu: true yes, updated abb024c

Vagabond (approved):

PR #993, opened 2021-09-13
state channel grpc return types

michaeldjeffrey:


src/state_channel/blockchain_state_channel_common.erl
```diff
@@ -310,6 +311,11 @@ handle_next_offer(HandlerState=#handler_state{offer_queue=[{NextOffer, OfferTime
             {ok, HandlerState}
     end.
 
+-spec handle_offer(
+        Offer :: blockchain_state_channel_packet_offer_v1:offer(),
+        Time :: pos_integer(),
+        HandlerState :: #handler_state{}
+) -> {ok, State :: #handler_state{}, Msg :: any()}.
```
michaeldjeffrey, line 318: Also, @andymck do we want to put a timeout on the receive on `L325`?

andymck: yeah prob good to so do

andymck:
lgtm

andymck (approved):

PR #998, opened 2021-09-15
speed up absorb

evanmcc:
we're doing a lot of silly stuff that's slowing absorb down.
- we don't use score anymore but we continue to update it
- we validate witnesses that we're not going to score

this PR:
- clears all scores (for snapshot ledger convergence) and stops updating them
- removes witness validation during absorb until we can land #919

### DIFF
--- a/src/transactions/blockchain_txn.erl
+++ b/src/transactions/blockchain_txn.erl
@@ -442,7 +442,12 @@ absorb_and_commit(Block, Chain0, BeforeCommit, Rescue) ->
                             ok = blockchain_ledger_v1:commit_context(Ledger2),
                             End2 = erlang:monotonic_time(millisecond),
                             ok = absorb_delayed(Block, Chain0),
-                            ok = absorb_aux(Block, Chain0),
+                            case absorb_aux(Block, Chain0) of
+                                ok -> ok;
+                                Err ->
+                                    lager:info("aux absorb failed with: ~p", [Err]),
+                                    ok
+                            end,
                             End3 = erlang:monotonic_time(millisecond),
                             lager:info("validation took ~p absorb took ~p post took ~p ms height ~p",
                                        [End - Start, End2 - End, End3 - End2, Height]),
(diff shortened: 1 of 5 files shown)


src/transactions/blockchain_txn.erl
```diff
@@ -433,16 +433,20 @@ absorb_and_commit(Block, Chain0, BeforeCommit, Rescue) ->
...
                 {ok, Chain2} ->
                     Ledger2 = blockchain:ledger(Chain2),
                     Hash = blockchain_block:hash_block(Block),
                     case BeforeCommit(Chain2, Hash) of
                         ok ->
                             ok = blockchain_ledger_v1:commit_context(Ledger2),
-                            absorb_delayed(Block, Chain0),
-                            absorb_aux(Block, Chain0);
+                            End2 = erlang:monotonic_time(millisecond),
+                            ok = absorb_delayed(Block, Chain0),
```
Vagabond, line 444: Is the goal to crash if these secondary absorbs fail?

Vagabond (commented):
Overall this looks good, but I'm not sure about asserting the secondary absorbs succeed, especially for aux absorb.

src/transactions/blockchain_txn.erl
```diff
@@ -433,16 +433,20 @@ absorb_and_commit(Block, Chain0, BeforeCommit, Rescue) ->
...
+                            ok = absorb_aux(Block, Chain0),
+                            End3 = erlang:monotonic_time(millisecond),
+                            lager:info("validation took ~p absorb took ~p post took ~p ms height ~p",
```
abhay, line 447:
```suggestion
                            lager:info("validation took ~p absorb took ~p post took ~p ms for block height ~p",
```

abhay: optional

Vagabond (approved):

src/transactions/blockchain_txn.erl
BenoitDuffez: `lager:debug` perhaps? the traces are very verbose

evanmcc: you can tune `blockchain` `slow_txn_log_threshold` with the .1 release for the low txn logging.  this is one line per block, which I think is reasonable.

PR #1002, opened 2021-09-16
POC support over grpc

andymck:
Core part of https://github.com/helium/miner/pull/1068

### DIFF
--- a/include/blockchain.hrl
+++ b/include/blockchain.hrl
@@ -55,7 +55,7 @@
          time :: non_neg_integer(),
          height :: non_neg_integer(),
          hash :: blockchain_block:hash(),
-         pocs :: map(),
+         pocs :: [{integer(), binary()}] | map(),
          hbbft_round :: non_neg_integer(),
          election_info :: {non_neg_integer(), non_neg_integer()},
          penalties :: {binary(), [{pos_integer(), binary()}]}
--- a/include/blockchain.hrl
+++ b/include/blockchain.hrl
@@ -55,7 +55,7 @@
          time :: non_neg_integer(),
          height :: non_neg_integer(),
          hash :: blockchain_block:hash(),
-         pocs :: [{integer(), binary()}] | map(),
+         pocs :: blockchain_block_v1:poc_keys() | map(),
          hbbft_round :: non_neg_integer(),
          election_info :: {non_neg_integer(), non_neg_integer()},
          penalties :: {binary(), [{pos_integer(), binary()}]}
--- a/include/blockchain.hrl
+++ b/include/blockchain.hrl
@@ -26,6 +26,7 @@
 
 % Misc
 -define(EVT_MGR, blockchain_event_mgr).
+-define(POC_EVT_MGR, blockchain_poc_event_mgr).
 
 -define(BC_UPGRADE_NAMES, [
     <<"gateway_v2">>,
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -135,8 +135,23 @@
 %% Number of blocks to wait before a hotspot can submit a poc challenge request
 -define(poc_challenge_interval, poc_challenge_interval).
 
+%% Number of challenges per block: integer
+-define(poc_challenge_rate, poc_challenge_rate).
+
+%% Actor type of the challenger: not set or 'validator'
+-define(poc_challenger_type, poc_challenger_type).
+
 %% Allow to switch POC version
 -define(poc_version, poc_version).
+
+%% Number of blocks after a POC is started at which point it will timeout/expire: integer
+-define(poc_timeout, poc_timeout).
+
+%% Number of blocks after poc_timeout at which point the poc public data will be deleted from the ledger: integer
+%% NOTE: the public poc data is required as part of the receipt_v2 txn validations
+%%       and so this value must be sufficient as to give time for absorb to occur
+-define(poc_receipts_absorb_timeout, poc_receipts_absorb_timeout).
+
 
 %% Number of blocks to wait before a hotspot can be eligible to participate in a poc
 %% challenge. This would avoid new hotspots getting challenged before they sync to an
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -141,6 +141,9 @@
 %% Actor type of the challenger: not set or 'validator'
 -define(poc_challenger_type, poc_challenger_type).
 
+%% scale the number of validators in the heartbeat proposals to ensure there are enough keys waiting
+-define(poc_validator_ct_scale, poc_validator_ct_scale).
+
 %% Allow to switch POC version
 -define(poc_version, poc_version).
 
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -529,6 +529,7 @@
 -define(validator_minimum_stake, validator_minimum_stake).  % bones
 -define(validator_liveness_interval, validator_liveness_interval).  % blocks
 -define(validator_liveness_grace_period, validator_liveness_grace_period).  % blocks
+-define(validator_hb_reactivation_limit, validator_hb_reactivation_limit). % addresses
 -define(validator_penalty_filter, validator_penalty_filter). % float
 -define(validator_key_check, validator_key_check). % boolean
 -define(stake_withdrawal_cooldown, stake_withdrawal_cooldown). % blocks
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -209,10 +209,6 @@
 %% the number of random hexes to utilize when targeting: integer
 -define(poc_target_pool_size, poc_target_pool_size).
 
-%% The number of subset of proposed POCs to mark as active when absorbing a block
--define(poc_proposal_processing_count, poc_proposal_processing_count).
-
-
 %%%
 %%% score vars
 %%%
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -152,6 +152,10 @@
 %%       and so this value must be sufficient as to give time for absorb to occur
 -define(poc_receipts_absorb_timeout, poc_receipts_absorb_timeout).
 
+%% Number of blocks for which a validator ephemeral key will remain valid : pos_integer
+%% after this period it will be GCed
+%% NOTE the minimum value here should be greater than that of validator_liveness_interval
+-define(poc_validator_ephemeral_key_timeout, poc_validator_ephemeral_key_timeout).
 
 %% Number of blocks to wait before a hotspot can be eligible to participate in a poc
 %% challenge. This would avoid new hotspots getting challenged before they sync to an
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -157,6 +157,10 @@
 %% NOTE the minimum value here should be greater than that of validator_liveness_interval
 -define(poc_validator_ephemeral_key_timeout, poc_validator_ephemeral_key_timeout).
 
+%% Determines whether or not to filter out inactive gateways
+%% from POC targets :: boolean()
+-define(poc_activity_filter_enabled, poc_activity_filter_enabled).
+
 %% Number of blocks to wait before a hotspot can be eligible to participate in a poc
 %% challenge. This would avoid new hotspots getting challenged before they sync to an
 %% acceptable height.
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -208,6 +208,11 @@
 
 %% the number of random hexes to utilize when targeting: integer
 -define(poc_target_pool_size, poc_target_pool_size).
+
+%% The number of subset of proposed POCs to mark as active when absorbing a block
+-define(poc_proposal_processing_count, poc_proposal_processing_count).
+
+
 %%%
 %%% score vars
 %%%
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -3064,9 +3064,6 @@ get_plausible_blocks(Itr, {ok, _Key, BinBlock}, Acc) ->
 
 run_gc_hooks(Blockchain, _Block) ->
     Ledger = blockchain:ledger(Blockchain),
-%%    Hash = blockchain_block:hash_block(Block),
-%%    Height = blockchain_block:height(Block),
-
     try
         ok = blockchain_ledger_v1:maybe_gc_pocs(Blockchain, Ledger),
 
@@ -3075,9 +3072,6 @@ run_gc_hooks(Blockchain, _Block) ->
         ok = blockchain_ledger_v1:maybe_gc_h3dex(Ledger),
 
         ok = blockchain_ledger_v1:maybe_recalc_price(Blockchain, Ledger)
-
-%%        BlockPOCs = blockchain_ledger_v1:process_poc_proposals(Height, Hash, Ledger),
-%%        ok = blockchain_worker:notify({poc_keys, {Height, Hash, false, BlockPOCs}})
 
         %% ok = blockchain_ledger_v1:refresh_gateway_witnesses(Hash, Ledger)
     catch What:Why:Stack ->
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -3090,10 +3090,7 @@ run_gc_hooks(Blockchain, _Hash) ->
 
         ok = blockchain_ledger_v1:maybe_gc_scs(Blockchain, Ledger),
 
-        case config(?poc_targeting_version, Ledger) of
-            {ok, 6} -> ok = blockchain_ledger_v1:maybe_gc_h3dex(Ledger);
-            _ -> ok
-        end,
+        ok = blockchain_ledger_v1:maybe_gc_h3dex(Ledger),
 
         ok = blockchain_ledger_v1:maybe_recalc_price(Blockchain, Ledger) %,
 
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -797,43 +797,23 @@ get_block_info(Height, Chain = #blockchain{db=DB, info=InfoCF}) ->
             Error
     end.
 
+-spec mk_block_info(blockchain_block:hash(), blockchain_block:block()) -> #block_info_v2{}.
 mk_block_info(Hash, Block) ->
-    %% POCs in the block will either be the poc request txns
-    %% or the poc keys in the block metadata
-    %% which is dependant upon chain var poc_challenger_type
-    %% given mk_block_info is called when loading blocks from a snapshot
-    %% and the ledger at this point seems to be empty, its not
-    %% possible to consult the ledger to get the chain var value
-    %% in this snapshot load scenario
-    %% as such check for the presence of either
-    %% and if the poc request txns are present, go with those
-    %% if not default to poc keys
-    %% TODO: REVIEW THIS APPROACH
-    PoCKeys = blockchain_block_v1:poc_keys(Block),
-    PoCRequests = maps:from_list(
-                    lists:flatmap(
-                        fun(Txn) ->
-                                 case blockchain_txn:type(Txn) of
-                                     blockchain_txn_poc_request_v1 ->
-                                         [{blockchain_txn_poc_request_v1:onion_key_hash(Txn),
-                                           blockchain_txn_poc_request_v1:block_hash(Txn)}];
-                                     _ -> []
-                                 end
-                         end,
-                     blockchain_block:transactions(Block))
-                ),
-    PoCs =
-        case PoCRequests of
-            M when map_size(M) == 0 ->
-                PoCKeys;
-            _ ->
-                PoCRequests
-        end,
+    PoCs = lists:flatmap(
+             fun(Txn) ->
+                     case blockchain_txn:type(Txn) of
+                         blockchain_txn_poc_request_v1 ->
+                             [{blockchain_txn_poc_request_v1:onion_key_hash(Txn),
+                               blockchain_txn_poc_request_v1:block_hash(Txn)}];
+                         _ -> []
+                     end
+             end,
+             blockchain_block:transactions(Block)),
 
     #block_info_v2{time = blockchain_block:time(Block),
                    hash = Hash,
                    height = blockchain_block:height(Block),
-                   pocs = PoCs,
+                   pocs = maps:from_list(PoCs),
                    hbbft_round = blockchain_block:hbbft_round(Block),
                    election_info = blockchain_block_v1:election_info(Block),
                    penalties = {blockchain_block_v1:bba_completion(Block), blockchain_block_v1:seen_votes(Block)}}.
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -2571,7 +2571,7 @@ save_block(Block, Chain = #blockchain{db=DB}) ->
 
 -spec save_block(blockchain_block:block(), rocksdb:batch_handle(), blockchain()) -> ok.
 save_block(Block, Batch, #blockchain{default=DefaultCF, blocks=BlocksCF, heights=HeightsCF,
-                                     info=InfoCF} = Chain) ->
+                                     info=InfoCF}) ->
     Height = blockchain_block:height(Block),
     Hash = blockchain_block:hash_block(Block),
     ok = rocksdb:batch_put(Batch, BlocksCF, Hash, blockchain_block:serialize(Block)),
@@ -2581,11 +2581,6 @@ save_block(Block, Batch, #blockchain{default=DefaultCF, blocks=BlocksCF, heights
     ok = rocksdb:batch_put(Batch, HeightsCF, Hash, <<Height:64/integer-unsigned-big>>),
     ok = rocksdb:batch_put(Batch, HeightsCF, <<Height:64/integer-unsigned-big>>, Hash),
     Info = mk_block_info(Hash, Block),
-    Ledger = ?MODULE:ledger(Chain),
-    %% TODO: confirm if this is the best place for generating the key proposals
-    %% and sending the associated event
-    BlockPOCs = blockchain_ledger_v1:process_poc_proposals(Height, Hash, Ledger),
-    ok = blockchain_worker:notify({poc_keys, {Height, Hash, false, BlockPOCs}}),
     ok = rocksdb:batch_put(Batch, InfoCF, <<Height:64/integer-unsigned-big>>, term_to_binary(Info)).
 
 save_temp_block(Block, #blockchain{db=DB, temp_blocks=TempBlocks, default=DefaultCF}=Chain) ->
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -3062,10 +3062,10 @@ get_plausible_blocks(Itr, {ok, _Key, BinBlock}, Acc) ->
              end,
     get_plausible_blocks(Itr, rocksdb:iterator_move(Itr, next), NewAcc).
 
-run_gc_hooks(Blockchain, Block) ->
+run_gc_hooks(Blockchain, _Block) ->
     Ledger = blockchain:ledger(Blockchain),
-    Hash = blockchain_block:hash_block(Block),
-    Height = blockchain_block:height(Block),
+%%    Hash = blockchain_block:hash_block(Block),
+%%    Height = blockchain_block:height(Block),
 
     try
         ok = blockchain_ledger_v1:maybe_gc_pocs(Blockchain, Ledger),
@@ -3074,10 +3074,10 @@ run_gc_hooks(Blockchain, Block) ->
 
         ok = blockchain_ledger_v1:maybe_gc_h3dex(Ledger),
 
-        ok = blockchain_ledger_v1:maybe_recalc_price(Blockchain, Ledger),
+        ok = blockchain_ledger_v1:maybe_recalc_price(Blockchain, Ledger)
 
-        BlockPOCs = blockchain_ledger_v1:process_poc_proposals(Height, Hash, Ledger),
-        ok = blockchain_worker:notify({poc_keys, {Height, Hash, false, BlockPOCs}})
+%%        BlockPOCs = blockchain_ledger_v1:process_poc_proposals(Height, Hash, Ledger),
+%%        ok = blockchain_worker:notify({poc_keys, {Height, Hash, false, BlockPOCs}})
 
         %% ok = blockchain_ledger_v1:refresh_gateway_witnesses(Hash, Ledger)
     catch What:Why:Stack ->
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -1970,12 +1970,13 @@ add_bin_snapshot(_BinSnap, _Height, none, _Chain) -> {error, no_snapshot_hash};
 add_bin_snapshot(BinSnap, Height, Hash, #blockchain{db=DB, dir=Dir, snapshots=SnapshotsCF}) when is_binary(Hash) ->
     try
         SnapDir = filename:join(Dir, "saved-snaps"),
-        SnapFile = list_to_binary(io_lib:format("snap-~s", [blockchain_utils:bin_to_hex(Hash)])),
+        SnapFile = io_lib:format("snap-~s", [blockchain_utils:bin_to_hex(Hash)]),
         OhSnap = filename:join(SnapDir, SnapFile),
         ok = save_bin_snapshots(OhSnap, BinSnap),
         {ok, Batch} = rocksdb:batch(),
-        %% store the snap as a filename
-        ok = rocksdb:batch_put(Batch, SnapshotsCF, Hash, <<"file:", SnapFile/binary>>),
+        %% store the snap as a filename, which might be wrong if compressed
+        ok = rocksdb:batch_put(Batch, SnapshotsCF, Hash,
+                               <<"file:", (iolist_to_binary(SnapFile))/binary>>),
         %% lexiographic ordering works better with big endian
         ok = rocksdb:batch_put(Batch, SnapshotsCF, <<Height:64/integer-unsigned-big>>, Hash),
         ok = rocksdb:write_batch(DB, Batch, [])
@@ -2008,11 +2009,7 @@ save_bin_snapshots(DestFilename, SnapshotData) ->
               end,
 
    Results = lists:map(fun(SaveFunc) ->
-                             try
-                                 SaveFunc(DestFilename, SnapshotData)
-                             catch _Class:Error ->
-                                 {error, Error}
-                             end
+                               SaveFunc(DestFilename, SnapshotData)
                        end, SaveFuns),
 
    case lists:all(
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -797,24 +797,43 @@ get_block_info(Height, Chain = #blockchain{db=DB, info=InfoCF}) ->
             Error
     end.
 
-
--spec mk_block_info(blockchain_block:hash(), blockchain_block:block()) -> #block_info_v2{}.
 mk_block_info(Hash, Block) ->
-    PoCs = lists:flatmap(
-             fun(Txn) ->
-                     case blockchain_txn:type(Txn) of
-                         blockchain_txn_poc_request_v1 ->
-                             [{blockchain_txn_poc_request_v1:onion_key_hash(Txn),
-                               blockchain_txn_poc_request_v1:block_hash(Txn)}];
-                         _ -> []
-                     end
-             end,
-             blockchain_block:transactions(Block)),
+    %% POCs in the block will either be the poc request txns
+    %% or the poc keys in the block metadata
+    %% which is dependant upon chain var poc_challenger_type
+    %% given mk_block_info is called when loading blocks from a snapshot
+    %% and the ledger at this point seems to be empty, its not
+    %% possible to consult the ledger to get the chain var value
+    %% in this snapshot load scenario
+    %% as such check for the presence of either
+    %% and if the poc request txns are present, go with those
+    %% if not default to poc keys
+    %% TODO: REVIEW THIS APPROACH
+    PoCKeys = blockchain_block_v1:poc_keys(Block),
+    PoCRequests = maps:from_list(
+                    lists:flatmap(
+                        fun(Txn) ->
+                                 case blockchain_txn:type(Txn) of
+                                     blockchain_txn_poc_request_v1 ->
+                                         [{blockchain_txn_poc_request_v1:onion_key_hash(Txn),
+                                           blockchain_txn_poc_request_v1:block_hash(Txn)}];
+                                     _ -> []
+                                 end
+                         end,
+                     blockchain_block:transactions(Block))
+                ),
+    PoCs =
+        case PoCRequests of
+            M when map_size(M) == 0 ->
+                PoCKeys;
+            _ ->
+                PoCRequests
+        end,
 
     #block_info_v2{time = blockchain_block:time(Block),
                    hash = Hash,
                    height = blockchain_block:height(Block),
-                   pocs = maps:from_list(PoCs),
+                   pocs = PoCs,
                    hbbft_round = blockchain_block:hbbft_round(Block),
                    election_info = blockchain_block_v1:election_info(Block),
                    penalties = {blockchain_block_v1:bba_completion(Block), blockchain_block_v1:seen_votes(Block)}}.
@@ -1156,6 +1175,7 @@ add_block_(Block, Blockchain, Syncing) ->
                                              Ledger, Height, Blockchain);
                         _ -> ok
                     end,
+
                     case blockchain_txn:Fun(Block, Blockchain, BeforeCommit, IsRescue) of
                         {error, Reason}=Error ->
                             lager:error("Error absorbing transaction, Ignoring Hash: ~p, Reason: ~p", [blockchain_block:hash_block(Block), Reason]),
@@ -3218,7 +3238,8 @@ blocks_test_() ->
                                                election_epoch => 1,
                                                epoch_start => 0,
                                                seen_votes => [],
-                                               bba_completion => <<>>
+                                               bba_completion => <<>>,
+                                               poc_keys => []
                                               }),
              Hash = blockchain_block:hash_block(Block),
              ok = add_block(Block, Chain),
@@ -3296,7 +3317,8 @@ get_block_test_() ->
                                                election_epoch => 1,
                                                epoch_start => 0,
                                                seen_votes => [],
-                                               bba_completion => <<>>
+                                               bba_completion => <<>>,
+                                               poc_keys => []
                                               }),
              Hash = blockchain_block:hash_block(Block),
              ok = add_block(Block, Chain),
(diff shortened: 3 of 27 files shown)


include/blockchain_vars.hrl
```diff
@@ -131,6 +131,12 @@
 %% Number of blocks to wait before a hotspot can submit a poc challenge request
 -define(poc_challenge_interval, poc_challenge_interval).
 
+%% Number of challenges per block
+-define(poc_challenge_rate, poc_challenge_rate).
+
+%% Actor type of the challenger
```
Vagabond, line 137: Can you list the possible values here?

src/blockchain.erl
```diff
@@ -725,20 +725,10 @@ get_block_info(Height, Chain = #blockchain{db=DB, info=InfoCF}) ->
     end.
 
 mk_block_info(Hash, Block) ->
-    PoCs = lists:flatmap(
```
Vagabond, line 728: Would suggest we gate where we fetch pocs from behind the `poc_challenger_type` var

src/blockchain_block_v1.erl
```diff
@@ -417,6 +426,13 @@ wrap_vote({Idx, Vector}) ->
 unwrap_vote(#blockchain_seen_vote_v1_pb{index = Idx, vector = Vector}) ->
     {Idx, Vector}.
 
+-spec wrap_poc_key({binary(), binary()}) -> #blockchain_poc_key_pb{}.
+wrap_poc_key({MinerAddr, Key}) ->
```
Vagabond, line 430: I think it would be good to store these as indexed into the consensus group array rather than duplicating the key?

src/poc/blockchain_poc_target_v4.erl
```diff
@@ -0,0 +1,268 @@
...
+    Ledger,
+    Vars,
+    HexList,
+    [{Hex, HexRandState0} | Tail] = _Attempted
+) ->
+    {ok, Height} = blockchain_ledger_v1:current_height(Ledger),
+    %% Get a list of gateway pubkeys within this hex
+    {ok, AddrList0} = blockchain_ledger_v1:get_hex(Hex, Ledger),
+    lager:info("gateways for hex ~p: ~p", [Hex, AddrList0]),
+    %% Remove challenger if present and also remove gateways who haven't challenged
```
Vagabond, line 55: I assume since the challenger is a validator now, this doesn't need to be done?

include/blockchain_vars.hrl
Vagabond, line 137:
```suggestion
%% Actor type of the challenger: not set or `validator'
```

include/blockchain_vars.hrl
```diff
@@ -131,6 +131,12 @@
 %% Number of blocks to wait before a hotspot can submit a poc challenge request
 -define(poc_challenge_interval, poc_challenge_interval).
 
+%% Number of challenges per block
```
Vagabond, line 134:
```suggestion
%% Number of PoC challenges per block: integer
```

src/poc/blockchain_poc_report_handler.erl
```diff
@@ -0,0 +1,105 @@
+%%%-------------------------------------------------------------------
```
Vagabond, line 1: I kind of feel like this belongs in the miner because other core-using repos like ETL or router don't ever need to do any of this.

src/poc/blockchain_poc_mgr.erl
```diff
@@ -0,0 +1,915 @@
+%%%-------------------------------------------------------------------
+%%% @doc
+%%% listens for block events, inspects the POCs in the block metadata
```
Vagabond, line 3: Consider moving to miner?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2104,97 +2109,170 @@ delete_poc(OnionKeyHash, Challenger, Ledger) ->
...
-                                            NewPoCs), [compressed]),
-                      cache_put(Ledger, PoCsCF, KeyHash, BinPoCs)
-              end,
-              Alters),
-            ok;
-        _ ->
-            ok
-    end.
+maybe_gc_pocs(_Chain, _Ledger) ->
+    ok.
```
Vagabond, line 2190: One idea for GC we've used elsewhere is to write a key listing what should be GC'd at a height in the future, so we can avoid linear scan GCing.

src/poc/blockchain_poc_report_handler.erl
andymck: yeah prob right, when i did this I had it in my head that miner was going to go away completely, ie replaced with the rust client but of course thats not true...it will actually just move to providing validator functionality....

src/poc/blockchain_poc_mgr.erl
andymck: yeah as above

src/blockchain_block_v1.erl
andymck: will do, have that on a cleanup list somewhere from when you previously mentioned

src/blockchain.erl
andymck: yip, still adding gating of that chainvar

src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -13,6 +13,7 @@
 -include("blockchain_vars.hrl").
 -include("blockchain_utils.hrl").
 -include_lib("helium_proto/include/blockchain_txn_poc_receipts_v1_pb.hrl").
```
Vagabond, line 15: Revert changes to this file if you're doing a v2?

src/ledger/v1/blockchain_ledger_poc_v3.erl
```diff
@@ -0,0 +1,231 @@
...
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+-record(poc_v3, {
+    onion_key_hash :: binary(),
+    challenger :: libp2p_crypto:pubkey_bin(),
+    block_hash :: binary(),
+    start_height :: pos_integer(),
+    orig_version :: pos_integer(),
+    secret_hash :: binary() | undefined %% TODO, carried over from v2 for backwards compatibility...maybe remove
```
Vagabond, line 32: It does seem unused, propose you remove it.

test/blockchain_vars_SUITE.erl
```diff
@@ -13,17 +13,17 @@
...
 
 %% Setup ----------------------------------------------------------------------
 
 all() ->
     [
         version_change_test,
-        master_key_test,
-        cache_test
+        master_key_test
+%%        cache_test
```
evanmcc, line 26: just rip this out entirely as we do caching differently now?

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
```diff
@@ -0,0 +1,1571 @@
...
+    BaseTxn = Txn#blockchain_txn_poc_receipts_v2_pb{signature = <<>>},
+    EncodedTxn = blockchain_txn_poc_receipts_v2_pb:encode_msg(BaseTxn),
+    Txn#blockchain_txn_poc_receipts_v2_pb{signature=SigFun(EncodedTxn)}.
+
+%%--------------------------------------------------------------------
+%% @doc
+%% @end
+%%--------------------------------------------------------------------
+-spec is_valid(txn_poc_receipts(), blockchain:blockchain()) -> ok | {error, atom()} | {error, {atom(), any()}}.
+is_valid(Txn, Chain) ->
```
evanmcc, line 153: there should probably be a gate here to keep these and v1s from being valid at the same time.

also maybe we should reconsider duplicating all this stuff if it's possible not to do so?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -895,6 +897,20 @@ validate_var(?min_assert_h3_res, Value) ->
     validate_int(Value, "min_assert_h3_res", 0, 15, false);
 validate_var(?poc_challenge_interval, Value) ->
     validate_int(Value, "poc_challenge_interval", 10, 1440, false);
+validate_var(?poc_challenge_rate, Value) ->
+    validate_int(Value, "poc_challenge_rate", 1, 1000, false);
```
evanmcc, line 901: cap here should probably be higher

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -895,6 +897,20 @@ validate_var(?min_assert_h3_res, Value) ->
...
+validate_var(?poc_timeout, Value) ->
+    validate_int(Value, "poc_timeout", 1, 1000, false);
+validate_var(?poc_receipts_absorb_timeout, Value) ->
+    validate_int(Value, "poc_receipts_absorb_timeout", 10, 10000, false);
```
evanmcc, line 905: cap here should be much, much lower.

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -895,6 +897,20 @@ validate_var(?min_assert_h3_res, Value) ->
     validate_int(Value, "min_assert_h3_res", 0, 15, false);
 validate_var(?poc_challenge_interval, Value) ->
     validate_int(Value, "poc_challenge_interval", 10, 1440, false);
+validate_var(?poc_challenge_rate, Value) ->
+    validate_int(Value, "poc_challenge_rate", 1, 1000, false);
+validate_var(?poc_timeout, Value) ->
+    validate_int(Value, "poc_timeout", 1, 1000, false);
```
evanmcc, line 903: cap probably way higher than this should ever be

include/blockchain.hrl
```diff
@@ -55,8 +55,8 @@
          time :: non_neg_integer(),
          height :: non_neg_integer(),
          hash :: blockchain_block:hash(),
-         pocs :: map(),
+         pocs :: [{integer(), binary()}] | [] | map(),
```
evanmcc, line 58: don't need the `[]` here

src/ledger/v1/blockchain_ledger_poc_v3.erl
```diff
@@ -0,0 +1,218 @@
...
+-export_type([poc/0, pocs/0, poc_result_type/0, poc_result_types/0]).
+
+-spec new(binary(), libp2p_crypto:pubkey_bin(), binary(), pos_integer()) -> poc().
+new(OnionKeyHash, Challenger, BlockHash, StartHeight) ->
+    #poc_v3{
+        onion_key_hash=OnionKeyHash,
+        challenger=Challenger,
+        block_hash=BlockHash,
+        start_height=StartHeight,
+        orig_version = 3
```
evanmcc, line 51: ?

src/ledger/v1/blockchain_ledger_poc_v3.erl
```diff
@@ -0,0 +1,218 @@
...
+
+-spec serialize(poc()) -> binary().
+serialize(PoC) ->
+    %% intentionally don't compress here, we compress these in batches
+    %% in the ledger code, which should get better compression anyway
+    BinPoC = erlang:term_to_binary(PoC),
+    <<3, BinPoC/binary>>.
+
+-spec deserialize(binary()) -> poc().
+deserialize(<<1, Bin/binary>>) ->
```
evanmcc, line 102: probably should remove the copied conversion stuff?

wondering if we should call the new poc-ish thing something else rather than poc v3?

src/poc/blockchain_poc_packet_v2.erl
```diff
@@ -0,0 +1,375 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Blockchain PoC Packet V2 ==
+%% @end
+%%%-------------------------------------------------------------------
```
evanmcc, line 5: can you explain briefly why we need a whole new version of the packet stuff?

evanmcc (commented):
I still need to more closely review the changes in the ledger module itself, but I worry that I won't get through that before tomorrow, so I wanted this to be waiting for you in the morning.

Things I think we can separate out and merge before the rest of this lands:
- poc keys in blocks and info support
- var hooks
- rip out cache test 
- really suspect that we can not do a receipt v2, or at least more aggressively factor out common code (prefer the former).
- do we need packet v2?

will have more tomorrow

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
andymck: there could be v1s and v2s in flight at the same time at the point we switch from hotspot generated POCs to validator generation POCs.  V1s could still be in the buffer or submitted by a hotspot after the switchover.  

Thinking we should allow those v1s to continue to be processed rather than forcing them to fail ?

src/transactions/v1/blockchain_txn_vars_v1.erl
andymck: reduced to 100

src/ledger/v1/blockchain_ledger_poc_v3.erl
andymck: i added this to keep track of which POCs had been upgraded.  Orig version then is the version they were upgraded from.

I think I may have had a use for it at one point but can no longer recall what it was.  Can likely be removed.  Will look at further....

src/ledger/v1/blockchain_ledger_poc_v3.erl
andymck: We actually have no need to support older version in this v3.  Older POCs are purged when v3 is activated.  In addition should any v2s hit the ledger after the v3 activation they will subsequently be GC in the next GC run.

So I can likely remove v1 and v2 upgraing

src/poc/blockchain_poc_packet_v2.erl
andymck: This was mostly to avoid making a mess of the v1 module.  In the v2 module we have no need for and no use of the ledger ( as this has to run on spots which have no chain ).

v1 has the Ledger var threaded everywhere and handling both cases was proving very messy. 

Ultimately once the switchover has bedded in we will have no use for the v1 and could consider dropping altogether or renaming v2 as v1.....

src/poc/blockchain_poc_packet.erl
```diff
@@ -420,4 +420,4 @@ encrypt_decrypt(Ledger) ->
     ?assertEqual(IVs, compute_ivs(IV, KeysAndData)),
     ok.
 
--endif.
```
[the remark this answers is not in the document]
andymck: i suspect this was done manually, i had originally reworked some of this module and then added to the v2 module instead

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -717,7 +717,8 @@ load_blocks(Ledger0, Chain, Snapshot) ->
                       Chain1 = blockchain:ledger(Ledger2, Chain),
                       Rescue = blockchain_block:is_rescue_block(Block),
                       {ok, _Chain} = blockchain_txn:absorb_block(Block, Rescue, Chain1),
-                      %% Hash = blockchain_block:hash_block(Block),
+                      Hash = blockchain_block:hash_block(Block),
```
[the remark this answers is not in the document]
andymck: done

src/region/blockchain_region_v1.erl
```diff
@@ -59,6 +59,7 @@ get_all_region_bins(Ledger) ->
 h3_to_region(H3, Ledger) ->
     h3_to_region(H3, Ledger, no_prefetch).
 
+
```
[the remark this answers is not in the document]
andymck: done

src/transactions/v1/blockchain_poc_witness_v1.erl
```diff
@@ -33,7 +33,7 @@
 -endif.
 
 -type poc_witness() :: #blockchain_poc_witness_v1_pb{}.
--type poc_witnesses() :: [poc_witness()].
+-type poc_witnesses() :: [poc_witness()] | [].
```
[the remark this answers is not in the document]
andymck: done

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1454,6 +1470,42 @@ validate_region_params(Var, Value) when is_binary(Value) ->
...
+
+%% separate out hook functions and call them in separate functions
+%% below the hook section.
+
+%% poc challenger type has been modified
+%% we want to clear out the pocs CF
+%% we dont care about its value, if its been
+%% updated then we wipe all POCs
+var_hook(?poc_challenger_type, _, Ledger) ->
+    lager:info("poc_challenger_type changed, puring pocs", []),
```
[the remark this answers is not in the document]
andymck: fixed

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1454,6 +1470,42 @@ validate_region_params(Var, Value) when is_binary(Value) ->
...
+    purge_pocs(Ledger),
+    ok;
+var_hook(_Var, _Value, _Ledger) ->
+    ok.
+
+unset_hook(?poc_challenger_type, Ledger) ->
+    lager:info("poc_challenger_type unset, puring pocs", []),
```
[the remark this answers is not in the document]
andymck: fixed

test/blockchain_vars_SUITE.erl
andymck: done

src/transactions/v1/blockchain_txn_vars_v1.erl
andymck: reduced to 50,  from talking with professor the feeling was we will likely want to set this as low as possible.  But not clear how low as yet

include/blockchain.hrl
andymck: done

src/ledger/v1/blockchain_ledger_poc_v3.erl
andymck: Re calling V3 something else, not sure...i mean it is a new version of POC and shares much of the same APIs as v2

andymck:
Why not do a receipt v2?  It makes for a much cleaner implementation as we no longer have to handle legacy flows relating to older POC versions

src/ledger/v1/blockchain_ledger_poc_v3.erl
andymck: Removed upgrade from v1 and v2 support

src/ledger/v1/blockchain_ledger_poc_v3.erl
andymck: removed orig_version support

andymck:
Re breaking out some parts of this PR such as the poc_keys in blocks into separate PRs, I am not sure how much value we get from that.  All of the new functionality is gated behind chain vars and in the most part that new functionality is additive...it is implemented in the most part without touching or modifying existing functionality or flows.  The degree of changes to existing in use code is light

include/blockchain_vars.hrl
```diff
@@ -135,9 +135,24 @@
...
+%% Number of challenges per block: integer
+-define(poc_challenge_rate, poc_challenge_rate).
+
+%% Actor type of the challenger: not set or 'validator'
+-define(poc_challenger_type, poc_challenger_type).
+
 %% Allow to switch POC version
 -define(poc_version, poc_version).
 
+%% Number of blocks after a POC is started at which point it will timeout/expire: integer
```
Vagabond, line 147: How do we distinguish these expiry times, and why is it useful to do so?

Is the first one about how long the challenger will respond to requests about the challenge (or witnesses will bother reporting to the challenger)? How is this one enforced?

andymck: the first timeout determines the lifespan of a POC, when that expires the receipt txn will be submitted.  Any subsequent witness reports will be rejected by the receiving validator and the client returned an expired result.

The second timeout then determines the point at which we will GC the public data from the ledger.  The public data is required to remain on the ledger until the txn is absorbed as it is required by absorb.  
Under normal conditions, the public data will be GC upon a successful absorb but should that fail for some unknown reason or not be triggered ( it the txn fails to be absorbed ) then we will eventually give up on the txn  being absorbed and GC the data here.  This is a fail safe...

Vagabond: Does this mean a late receipts_v2 txn after this number of blocks is also invalid?

src/blockchain.erl
```diff
@@ -796,24 +796,43 @@ get_block_info(Height, Chain = #blockchain{db=DB, info=InfoCF}) ->
...
-                     end
-             end,
-             blockchain_block:transactions(Block)),
+    %% POCs in the block will either be the poc request txns
+    %% or the poc keys in the block metadata
+    %% which is dependant upon chain var poc_challenger_type
+    %% given mk_block_info is called when loading blocks from a snapshot
+    %% and the ledger at this point seems to be empty, its not
+    %% possible to consult the ledger to get the chain var value
+    %% in this snapshot load scenario
```
Vagabond, line 806: The ledger actually exists with the right vars, but it's not "installed" yet in a place it can have a handle to it gotten, so this would be fixable with some juggling.

include/blockchain.hrl
```diff
@@ -55,7 +55,7 @@
          time :: non_neg_integer(),
          height :: non_neg_integer(),
          hash :: blockchain_block:hash(),
-         pocs :: map(),
+         pocs :: [{integer(), binary()}] | map(),
```
Vagabond, line 58: I think this is no longer correct?

src/blockchain_block_v1.erl
```diff
@@ -58,7 +59,8 @@
                        rescue_signature => binary(),
                        seen_votes => [{pos_integer(), binary()}],
                        bba_completion => binary(),
-                       snapshot_hash => binary()
+                       snapshot_hash => binary(),
+                       poc_keys => [any()]
```
Vagabond, line 63: Better type here?

src/blockchain_block_v1.erl
```diff
@@ -437,6 +446,13 @@ wrap_vote({Idx, Vector}) ->
 unwrap_vote(#blockchain_seen_vote_v1_pb{index = Idx, vector = Vector}) ->
     {Idx, Vector}.
 
+-spec wrap_poc_key({libp2p_crypto:pub_key_bin(), binary()}) -> #blockchain_poc_key_pb{}.
```
vihu, line 449:
```suggestion
-spec wrap_poc_key({libp2p_crypto:pubkey_bin(), binary()}) -> #blockchain_poc_key_pb{}.
```

include/blockchain_vars.hrl
andymck: yes, the receipts_v2 will fail validations after (poc_timeout + poc_receipts_absorb_timeout) has passed as the public poc data will be GC and thus is_valid will fail

src/blockchain.erl
andymck: i am now actually wondering if we need to save the POC keys to the snapshot like this at all. The keys will be in the bloc which is obviously part of the snapshot and the keys will be processed and populated to the ledger when those blocks are processed during snapshot load

src/blockchain_block_v1.erl
andymck: fixed here: https://github.com/helium/blockchain-core/pull/1002/commits/294207c73e08fe3d189ecafaf95148a6b8d3e8d9

include/blockchain.hrl
andymck: fixed here: https://github.com/helium/blockchain-core/commit/294207c73e08fe3d189ecafaf95148a6b8d3e8d9

src/poc/blockchain_poc_target_v6.erl
```diff
@@ -0,0 +1,232 @@
+%%%-----------------------------------------------------------------------------
+%%% @doc blockchain_poc_target_v6 implementation.
+%%%
+%%% The targeting mechanism is based on the following conditions:
+%%% - Deterministically identify a target region based on public key
+%%% - Deterministically select a challengee from target region based on private key
```
Vagabond, line 6: Comment why v6 is different from v5

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -5020,6 +5239,27 @@ load_threshold_txns(Txns, Ledger) ->
...
+    end.
+
+-spec snapshot_pocs(atom(), ledger()) -> [{binary(), binary()}].
+snapshot_pocs(validator, Ledger) ->
+    PoCsCF = pocs_cf(Ledger),
+    lists:sort(
+      maps:to_list(
+        cache_fold(
+          Ledger, PoCsCF,
+          fun({OnionKeyHash, BValue}, Acc) ->
```
Vagabond, line 5256: isn't there a version byte here?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -5020,6 +5239,27 @@ load_threshold_txns(Txns, Ledger) ->
...
 snapshot_pocs(Ledger) ->
+    case blockchain:config(?poc_challenger_type, Ledger) of
+        {ok, Type} ->
+            snapshot_pocs(Type, Ledger);
+        _ ->
+            snapshot_pocs(undefined, Ledger)
+    end.
+
+-spec snapshot_pocs(atom(), ledger()) -> [{binary(), binary()}].
+snapshot_pocs(validator, Ledger) ->
```
Vagabond, line 5250: it might be better to use snapshot_raw here, if possible to avoid the expense of de/re serializing

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -4501,29 +4715,34 @@ maybe_gc_h3dex(Ledger) ->
...
-            {ok, #block_info_v2{hash = BlockHash}} = get_block_info(Height, Ledger),
-            RandState = blockchain_utils:rand_from_hash(BlockHash),
-            RequestFilter = fun(T) ->
-                                    blockchain_txn:type(T) == blockchain_txn_poc_receipts_v1
-                            end,
-            case blockchain_utils:find_txn(Block, RequestFilter) of
-                [] ->
-                    %% no receipts, don't do any GC
+            case get_block(Height, Ledger)  of
+                {error, _} ->
```
Vagabond, line 4719: How can we be missing the block here and why do we tolerate that?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2068,7 +2080,190 @@ delete_poc(OnionKeyHash, Challenger, Ledger) ->
...
+          []
+    ).
+
 maybe_gc_pocs(Chain, Ledger) ->
+    case blockchain:config(?poc_challenger_type, Ledger) of
+        {ok, validator} ->
+            maybe_gc_pocs(Chain, Ledger, validator);
+        _ ->
+            maybe_gc_pocs(Chain, Ledger, undefined)
+    end.
```
Vagabond, line 2208: add a newline here between function arities

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
Vagabond: Since you remove the old pocs from the ledger the v1s won't validate after the var hook fires. It would be nice to not duplicate this large, complex module and simply have both kinds of challenges output a receipts_v1?

src/ledger/v1/blockchain_ledger_v1.erl
andymck: no, the version bytes wraps the poc payload bin ie <<3, POCBIN>>

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
andymck: when switching between hotspot generated and validator generated POCs, currently active POCs will fail to validate.  Those POCs will be lost.

We discussed previously about handling it all in receipts v1 and it was agreed better to use a new v2 as it makes for  easier maintenance and extending going forward....the module is cleaner as we can drop all the legacy poc version support

src/poc/blockchain_poc_target_v6.erl
andymck: fixed here: https://github.com/helium/blockchain-core/pull/1002/commits/4220211c858f8840e7967487c98d249c7ee47988

src/ledger/v1/blockchain_ledger_v1.erl
andymck: fixed here: https://github.com/helium/blockchain-core/pull/1002/commits/4220211c858f8840e7967487c98d249c7ee47988

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -607,9 +607,14 @@ load_into_ledger(Snapshot, L0, Mode) ->
     case blockchain_ledger_v1:check_key(<<"poc_upgrade">>, L) of
         true -> ok;
         _ ->
-            %% have to do this here, otherwise it'll break block loads
-            blockchain_ledger_v1:upgrade_pocs(L),
-            blockchain_ledger_v1:mark_key(<<"poc_upgrade">>, L)
+            case blockchain_ledger_v1:config(?poc_challenger_type, L) of
+                {ok, _validator} ->
```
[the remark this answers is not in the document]
andymck: fixed here: https://github.com/helium/blockchain-core/pull/1002/commits/4220211c858f8840e7967487c98d249c7ee47988

src/ledger/v1/blockchain_ledger_v1.erl
Vagabond: depending on the expiry times you might not have all the blocks included that you need to re-create the poc keys?

src/ledger/v1/blockchain_ledger_v1.erl
Vagabond: Yeah, I think we shold just crash here

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
```diff
@@ -146,11 +164,43 @@ absorb(Txn, Chain) ->
     Validator = address(Txn),
     Version = version(Txn),
     TxnHeight = height(Txn),
-
+    POCKeyProposals = poc_key_proposals(Txn),
```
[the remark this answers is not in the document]
Vagabond: so is_valid needs to be amended for this txn, Addtionally hbs shold not contain these fields until the var is active?

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
```diff
@@ -0,0 +1,1599 @@
...
+
+-spec check_path_continuation(Elements :: [blockchain_poc_path_element_v1:poc_element()]) -> boolean().
+check_path_continuation(Elements) ->
+    lists:any(fun(E) ->
+                      blockchain_poc_path_element_v1:receipt(E) /= undefined orelse
+                      blockchain_poc_path_element_v1:witnesses(E) /= []
+              end,
+              Elements).
+
+%%-spec assign_alpha_beta(HasContinued :: boolean(),
```
vihu, line 430: nit: remove if not needed

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
```diff
@@ -0,0 +1,1599 @@
...
+                                       Res -> Res
+                               catch _:_ ->
+                                             Witnesses = lists:reverse(blockchain_poc_path_element_v1:witnesses(Element)),
+                                             {undefined, [{false, <<"tagged_witnesses_crashed">>, Witness} || Witness <- Witnesses]}
+                               end,
+
+                                Fun(Element, {TaggedWitnesses, FilteredReceipt}, Acc)
+                        end, Acc0, lists:zip(lists:seq(1, length(Path)), Path));
+        {error, request_block_hash_not_found} -> []
+    catch
```
vihu, line 587: Do we think we can handle the errors originating from `get_channels` slightly better? Presumably only `unknown_region` can still happen, the remaining errors shouldn't trigger anymore afaict. Maybe we can move the error handling up to `get_channels` itself?

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
```diff
@@ -0,0 +1,1599 @@
...
+            lists:foldl(fun({_ElementPos, Element}, Acc) ->
+                                Witnesses = lists:reverse(blockchain_poc_path_element_v1:witnesses(Element)),
+                                Fun(Element, {[{false, <<"regulatory_regions_not_set">>, Witness} || Witness <- Witnesses], undefined}, Acc)
+                        end, Acc0, lists:zip(lists:seq(1, length(Path)), Path))
+    end.
+
+
+%% again this is broken because of the current witness situation
+
+%% Iterate over all poc_path elements and for each path element calls a given
```
vihu, line 617: nit: remove if unused

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
```diff
@@ -0,0 +1,1599 @@
...
+        lager:info("POC POC POC ~p", [PoC]),
+        case blockchain_ledger_poc_v3:verify(PoC, Challenger, BlockHash) of
+            false ->
+                {error, invalid_poc};
+            true ->
+                %% get rid of this for the time being, we will need to restore it later when
+                %% the clean witness restore thing lands
+
+                %% %% Add filtered witnesses with poc-v9
+                %% ok = valid_path_elements_fold(fun(Element, {FilteredWitnesses, FilteredReceipt}, _) ->
```
vihu, line 678: remove

src/poc/blockchain_poc_target_v5.erl
```diff
@@ -0,0 +1,223 @@
+%%%-----------------------------------------------------------------------------
+%%% @doc blockchain_poc_target_v5 implementation.
```
vihu, line 2: This module is never actually used? Why not remove this and switch target_v6 to be target_v5?

andymck: v5 is actually in use by default.  It will be in use until we switch to h3dex targeting aka the v6 module

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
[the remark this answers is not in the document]
andymck: so put the bounds checking in is_valid, right ? make sense

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2068,7 +2080,191 @@ delete_poc(OnionKeyHash, Challenger, Ledger) ->
...
+) -> ok.
+process_poc_keys(_Block, 1, _BlockHash, _Ledger) ->
+    ok;
+process_poc_keys(Block, BlockHeight, BlockHash, Ledger) ->
+    %% we need to update the ledger with public poc data
+    %% based on the blocks poc ephemeral keys
+    %% these will be a prop with tuples: {ChallengerAddr :: binary(), PocKeyHash :: binary()}
+    case blockchain:config(?poc_challenger_type, Ledger) of
+        {ok, validator} ->
+            BlockPocEphemeralKeys = blockchain_block_v1:poc_keys(Block),
```
Vagabond, line 2097: Instead of this, use the block hash as entropy to select the ephemeral keys from the porential ephemeral keys in the ledger.

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
```diff
@@ -146,11 +164,43 @@ absorb(Txn, Chain) ->
...
+            {ok, ConsensusAddrs} = blockchain_ledger_v1:consensus_members(Ledger),
+            case lists:member(Validator, ConsensusAddrs) of
+                true ->
+                    ok;
+                false ->
+                    case blockchain:config(poc_challenger_type, Ledger) of
+                        {ok, validator} ->
+                            case application:get_env(blockchain, poc_mgr_mod) of
+                                {ok, POCMgrMod} ->
+                                    POCMgrMod:save_poc_key_proposals(Validator, POCKeyProposals, TxnHeight);
```
Vagabond, line 182: I think the proposals need to be saved for everyone following the chain, if this is not currently the case.

src/ledger/v1/blockchain_ledger_v1.erl
andymck: redundant, poc keys being removed from block

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
andymck: was decided to keep as it

src/ledger/v1/blockchain_ledger_v1.erl
andymck: redundant

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
andymck: redundant

src/transactions/blockchain_txn.erl
```diff
@@ -910,7 +919,7 @@ absorb_aux(Block0, Chain0) ->
 plain_absorb_(Block, Chain0) ->
     case ?MODULE:absorb_block(Block, Chain0) of
         {ok, _} ->
-            %% Hash = blockchain_block:hash_block(Block),
+            _Hash = blockchain_block:hash_block(Block),
```
Vagabond, line 922: Why?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -687,12 +687,16 @@ process_hooks(Vars, Unsets, Ledger) ->
...
-var_hook(?poc_targeting_version, 4, Ledger) ->
-    %% v4 targeting enabled, build the h3dex lookup
+var_hook(?poc_targeting_version, 6, Ledger) ->
+    %% purge any active POCs
+    purge_pocs(Ledger),
+    %% build the h3dex lookup
     {ok, Res} = blockchain_ledger_v1:config(?poc_target_hex_parent_res, Ledger),
     blockchain_ledger_v1:build_random_hex_targeting_lookup(Res, Ledger),
     ok;
-var_hook(?poc_targeting_version, 3, Ledger) ->
```
Vagabond, line 695: I'm not sure it's safe to remove old var hooks as that will probably break resync, I think you must add new ones or broaden them

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2068,7 +2082,243 @@ delete_poc(OnionKeyHash, Challenger, Ledger) ->
...
+) -> [blockchain_ledger_poc_v3:pocs()].
+process_poc_proposals(1, _BlockHash, _Ledger) ->
+    [];
+process_poc_proposals(BlockHeight, BlockHash, Ledger) ->
+    %% we need to update the ledger with public poc data
+    %% based on the blocks poc ephemeral keys
+    case blockchain:config(?poc_challenger_type, Ledger) of
+        {ok, validator} ->
+            %% Get only the proposed POCs from ledger
+            %% TODO: This will likely be slow AF
```
Vagabond, line 2098: We can't ship with this, I thought we discussed doing a random sub select from rocksdb and to avoid pulling the whole list.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2068,7 +2082,243 @@ delete_poc(OnionKeyHash, Challenger, Ledger) ->
...
+            {ok, blockchain_ledger_poc_v3:deserialize(BinPoC)};
+        not_found ->
+            {error, not_found};
+        Error ->
+            Error
+    end.
+
+-spec delete_public_poc(binary(), ledger()) -> ok | {error, any()}.
+delete_public_poc(OnionKeyHash, Ledger) ->
+    case ?MODULE:find_public_poc(OnionKeyHash, Ledger) of
```
Vagabond, line 2195: Just delete it?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2068,7 +2082,243 @@ delete_poc(OnionKeyHash, Challenger, Ledger) ->
...
+                     Ledger :: ledger()) -> ok | {error, _}.
+update_public_poc(POC, Ledger) ->
+    POCAddr = blockchain_ledger_poc_v3:onion_key_hash(POC),
+    Bin = blockchain_ledger_poc_v3:serialize(POC),
+    POCsCF = pocs_cf(Ledger),
+    cache_put(Ledger, POCsCF, POCAddr, Bin).
+
+
+-spec active_public_pocs(ledger()) -> [blockchain_ledger_poc_v3:poc()].
+active_public_pocs(Ledger) ->
```
Vagabond, line 2215: This doesn't distinguish proposed from active?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -4724,6 +5001,23 @@ fold_validators(Fun, InitAcc, Ledger) ->
       InitAcc
      ).
 
+-spec validator_count(ledger()) -> {ok, non_neg_integer()} | {error, any()}.
+validator_count(Ledger) ->
```
Vagabond, line 5005: Do we need a ledger upgrade hook or something to bootstrap this value?

src/transactions/blockchain_txn.erl
andymck: that is from recent testing of the new poc_key_proposals...ignore

src/ledger/v1/blockchain_ledger_v1.erl
andymck: @evanmcc added a bootstrap upon the poc_challenge_type being set to `validator`

https://github.com/helium/blockchain-core/blob/andymck/poc-grpc-v2/src/transactions/v1/blockchain_txn_vars_v1.erl#L720

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -5281,7 +5613,7 @@ load_hexes(Hexes0, Ledger) ->
 -spec snapshot_h3dex(ledger()) -> [{binary(), binary()}].
 snapshot_h3dex(Ledger) ->
     case config(?poc_targeting_version, Ledger) of
-        {ok, N} when N >= 4 ->
+        {ok, N} when N >= 6 ->
```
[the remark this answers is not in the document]
andymck: v6 is the h3dex version of targeting on this branch.  V5 is the non h3dex version
So the change here is consistent with that already there, ie we only snapshot h3dex if using that targeting version or later ?

src/ledger/v1/blockchain_ledger_v1.erl
andymck: Lets see where we end up with the POC CFs, and whether we have a sep for proposed and active

src/ledger/v1/blockchain_ledger_v1.erl
andymck: done

src/transactions/v1/blockchain_txn_vars_v1.erl
andymck: doh brainfart, fixed

src/blockchain.erl
andymck: removed as part of the new poc_key_proposals via ledger support

jadeallenx (approved):

RKB00100:
Fix chin.

madninja:
> Fix chin.

excuse me?

RKB00100:
Yes.

madninja:
> Yes.

Yes what, please ask a question, make a comment or stop

PR #1007, opened 2021-09-20
[DNM] Allow exporting gps viz compat poc-11 aux data

vihu:
This adds aux ledger specific functions to export poc-v11 gps-visualizer compatibility data.


vihu:
This is not required in core as it is sort of temporary code.

PR #1023, opened 2021-09-28
pull block gossip

evanmcc:
currently, we send a lot more blocks out than we need to.  this costs a lot in terms of bandwidth, especially as blocks grow larger and larger.  this PR attempts to move gossip from push to pull, where network nodes only pull blocks that they're unfamiliar with.

depends on: https://github.com/helium/proto/pull/94

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -33,6 +33,8 @@
     compact/1,
 
     is_block_plausible/2,
+    get_plausible_block/2,
+    have_plausible_block/2,
 
     last_block_add_time/1,
 
@@ -994,6 +996,8 @@ get_key_or_keys(Ledger) ->
     end.
 
 add_block_(Block, Blockchain, Syncing) ->
+    %% TODO: we know that swarm calls can block, it would be nice if we had all the pubkey bin and
+    %% tid calls threaded through from the top
     Ledger = blockchain:ledger(Blockchain),
     {ok, LedgerHeight} = blockchain_ledger_v1:current_height(Ledger),
     {ok, BlockchainHeight} = blockchain:height(Blockchain),
@@ -1006,6 +1010,7 @@ add_block_(Block, Blockchain, Syncing) ->
                     Hash = blockchain_block:hash_block(Block),
                     Sigs = blockchain_block:signatures(Block),
                     MyAddress = try blockchain_swarm:pubkey_bin() catch _:_ -> nomatch end,
+                    SwarmTID = blockchain_swarm:tid(),
                     BeforeCommit = fun(FChain, FHash) ->
                                            lager:debug("adding block ~p", [Height]),
                                            ok = ?save_block(Block, Blockchain),
@@ -1041,10 +1046,16 @@ add_block_(Block, Blockchain, Syncing) ->
                             end,
                             Error;
                         ok ->
+                            blockchain_gossip_handler:regossip_block(Block, Height, Hash, SwarmTID),
                             run_absorb_block_hooks(Syncing, Hash, Blockchain)
                     end;
                 plausible ->
-                    case save_plausible_block(Block, Blockchain) of
+                    %% regossip plausible blocks
+                    Height = blockchain_block:height(Block),
+                    Hash = blockchain_block:hash_block(Block),
+                    SwarmTID = blockchain_swarm:tid(),
+                    blockchain_gossip_handler:regossip_block(Block, Height, Hash, SwarmTID),
+                    case save_plausible_block(Block, Hash, Blockchain) of
                         exists -> ok; %% already have it
                         ok -> plausible %% tell the gossip handler
                     end;
@@ -2533,9 +2544,8 @@ is_block_plausible(Block, Chain) ->
     end.
 
 
-save_plausible_block(Block, #blockchain{db=DB, plausible_blocks=PlausibleBlocks}) ->
+save_plausible_block(Block, Hash, #blockchain{db=DB, plausible_blocks=PlausibleBlocks}) ->
     true = blockchain_lock:check(), %% we need the lock for this
-    Hash = blockchain_block:hash_block(Block),
     case rocksdb:get(DB, PlausibleBlocks, Hash, []) of
         {ok, _} ->
             %% already got it, thanks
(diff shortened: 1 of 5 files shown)


src/handlers/blockchain_gossip_handler.erl
```diff
@@ -66,6 +72,34 @@ handle_gossip_data(_StreamPid, Data, [SwarmTID, Blockchain]) ->
...
+            OtherHash when is_binary(OtherHash) ->
+                lager:warning("got non-matching hash ~p for height ~p from ~p",
+                              [Hash, Height, blockchain_utils:addr2name(From)]),
+                ok;
+            {error, not_found} ->
+                %% don't appear to have the block, do we have a plausible block?
+                case blockchain:have_plausible_block(Hash, Blockchain) of
+                    true -> ok;
+                    false ->
+                        %% don't have it in plausible either, try to sync it from the sender.
```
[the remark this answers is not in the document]
Vagabond: or do it over in start_block_sync and make it a general thing. Often we have one missing block (that we missed gossip on) but we get back 10, 9 of which we already have.

evanmcc:
deployment plan is: ship this with config set to v1 or unset (release 1).  once it's out there and broadly released, we can do another version with config set to v2 (release 2) and everyone should smoothly upgrade.  people who're left behind (on release 0 or before) will still be able to sync, but won't understand incoming gossip.

src/blockchain.erl
```diff
@@ -2515,6 +2516,12 @@ save_plausible_block(Block, #blockchain{db=DB, plausible_blocks=PlausibleBlocks}
             rocksdb:put(DB, PlausibleBlocks, Hash, blockchain_block:serialize(Block), [{sync, true}])
     end.
 
+have_plausible_block(Hash, #blockchain{db=DB, plausible_blocks=CF}) ->
```
vihu, line 2519:
```suggestion
-spec have_plausible_block(Hash :: blockchain_block:hash(), Chain :: blockchain()) -> boolean().
have_plausible_block(Hash, #blockchain{db=DB, plausible_blocks=CF}) ->
```

src/blockchain.erl
```diff
@@ -1002,6 +1004,8 @@ add_block_(Block, Blockchain, Syncing) ->
                     Hash = blockchain_block:hash_block(Block),
                     Sigs = blockchain_block:signatures(Block),
                     MyAddress = try blockchain_swarm:pubkey_bin() catch _:_ -> nomatch end,
+                    Swarm = blockchain_swarm:swarm(),
```
Vagabond, line 1007: I think you can do blockchain_swarm:tid() which won't block.

src/handlers/blockchain_sync_handler.erl
```diff
@@ -134,23 +149,46 @@ handle_data(server, Data, #state{blockchain=Blockchain, batch_size=BatchSize,
...
+                            {ok, StartingBlock} ->
+                                blockchain:build(StartingBlock, Blockchain, BatchSize);
+                            {error, _Reason} ->
+                                []
+                        end;
+                    R when is_list(R) ->
+                        %% just send these.  if there are more of them than the batch size, then just
+                        %% send the batch and remove them from the list
+                        R2 = case length(Requested) > BatchSize of
+                                 true -> lists:sublist(Requested, BatchSize);
```
Vagabond, line 168: You can always use sublist safely, no need for the case check here.

src/handlers/blockchain_sync_handler.erl
```diff
@@ -134,23 +149,46 @@ handle_data(server, Data, #state{blockchain=Blockchain, batch_size=BatchSize,
...
+                        lists:flatmap(
+                          fun(Hsh) ->
+                                  case blockchain:get_block(Hsh, Blockchain) of
+                                      {ok, B} -> [B];
+                                      _ -> []
+                                  end
+                          end,
+                          R2)
+                end,
+            Requested1 = Requested -- Blocks,
```
Vagabond, line 180: Aren't you subtracting a list of blocks from a list of heights here?

src/handlers/blockchain_sync_handler.erl
```diff
@@ -134,23 +149,46 @@ handle_data(server, Data, #state{blockchain=Blockchain, batch_size=BatchSize,
...
+                        end;
+                    R when is_list(R) ->
+                        %% just send these.  if there are more of them than the batch size, then just
+                        %% send the batch and remove them from the list
+                        R2 = case length(Requested) > BatchSize of
+                                 true -> lists:sublist(Requested, BatchSize);
+                                 false -> Requested
+                             end,
+                        lists:flatmap(
+                          fun(Hsh) ->
```
Vagabond, line 172: Call this Height for clarity? Since Requested is matched from a 'heights' field in the proto?

src/handlers/blockchain_sync_handler.erl
```diff
@@ -134,23 +149,46 @@ handle_data(server, Data, #state{blockchain=Blockchain, batch_size=BatchSize,
...
+                [_|_] ->
+                    Msg1 = #blockchain_sync_blocks_pb{blocks=[blockchain_block:serialize(B) || B <- Blocks]},
+                    Msg0 = blockchain_sync_handler_pb:encode_msg(Msg1),
+                    Msg = case Path of
+                              ?SYNC_PROTOCOL_V1 -> Msg0;
+                              ?SYNC_PROTOCOL_V2 -> zlib:compress(Msg0)
+                          end,
+                    {noreply, State#state{batches_sent=Sent+1, block=lists:last(Blocks), requested = Requested1}, Msg}
             end;
         #blockchain_sync_req_pb{msg={response, true}} when Sent < Limit, State#state.block /= undefined ->
```
Vagabond, line 193: In theory we'd want to check if Requested was not the empty list here? Otherwise your first batch of blocks are from your Requested list and all further batches are straight linear sync.

src/handlers/blockchain_sync_handler.erl
```diff
@@ -134,23 +149,46 @@ handle_data(server, Data, #state{blockchain=Blockchain, batch_size=BatchSize,
...
+                          fun(Hsh) ->
+                                  case blockchain:get_block(Hsh, Blockchain) of
+                                      {ok, B} -> [B];
+                                      _ -> []
+                                  end
+                          end,
+                          R2)
+                end,
+            Requested1 = Requested -- Blocks,
+            case Blocks of
```
Vagabond, line 181: You could also stop if Requested was /= [] but Requested1 == [], as in all the explicitly requested blocks were provided.

src/handlers/blockchain_sync_handler.erl
```diff
@@ -132,47 +147,89 @@ handle_data(client, Data0, #state{blockchain=Chain, path=Path}=State) ->
...
-                                      ?SYNC_PROTOCOL_V2 -> zlib:compress(Msg0)
-                                  end,
-                            {noreply, State#state{batches_sent=Sent+1, block=lists:last(Blocks)}, Msg}
-                    end;
-                {error, _Reason} ->
-                    {stop, normal, State}
+        #blockchain_sync_req_pb{msg={hash,
+                                     #blockchain_sync_hash_pb{hash = Hash,
+                                                              heights = Requested}}} ->
+            %% for resuming partial requests, remaining blocks saved in the state override what
```
Vagabond, line 155: resuming partial requests would fall through to the {response, true} case below?

src/handlers/blockchain_gossip_handler.erl
```diff
@@ -28,52 +30,108 @@
...
+                            true ->
+                                case find_missing_block(Height, Blockchain) of
+                                    {ok, Missing} ->
+                                        blockchain_worker:target_sync(From, Missing);
+                                    _Err ->
+                                        ok
+                                end;
+                            false ->
+                                %% don't have it in plausible either, try to sync it from the sender.
+                                blockchain_worker:target_sync(From)
```
Vagabond, line 71: potentially use find_missing_block here as well, at least when Height is > blockchain:height(Blockchain) + 1?

src/handlers/blockchain_gossip_handler.erl
```diff
@@ -28,52 +30,108 @@
...
     lager:debug("gossiping block to peers on init"),
-    {send, gossip_data(SwarmTID, Block)};
+    case application:get_env(blockchain, gossip_version, 1) of
+        1 ->
+            {ok, Block} = blockchain:head_block(Blockchain),
+            {send, gossip_data_v1(SwarmTID, Block)};
+        2 ->
+            {ok, Height} = blockchain_ledger_v1:current_height(blockchain:ledger(Blockchain)),
+            {ok, #block_info{hash = Hash}} = blockchain:get_block_info(Height, Blockchain),
+            lager:debug("gossiping block to peers on init"),
```
vihu, line 41:
```suggestion
            lager:debug("gossiping block @ ht: ~p to peers on init", [Height]),
```

Issue #1029, opened 2021-10-01
vars checking needs to improve before big binary blobs go in


evanmcc:
https://github.com/helium/blockchain-core/blob/1c2b2c916470a546fe4bfb93d127b3ff23c7a9a7/src/transactions/v1/blockchain_txn_poc_receipts_v1.erl#L1165

The above function gets called once per receipt, and once we have bigger binary blob vars, will quickly become a performance problem.  It needs to be rewritten to take a specific list of vars to fetch individually so we don't spend a lot of time fetching megs of data we never use.  Ideally this will happen before poc v11 gets activated.

Vagabond:
I propose we always pull from the ledger (remove the vars extraction done here and in the rewards txn and possibly elsewhere) and introduce a e2qc cache of configurable size over the vars (keyed by {chain_var_nonce, var_name}). This way if the chain var nonce changes the cache seamlessly transitions to the new vars.

evanmcc:
fixed by #1030

PR #1030, opened 2021-10-02
WIP add var cache

vihu:
Attempt to address https://github.com/helium/blockchain-core/issues/1029


src/blockchain_utils.erl
```diff
@@ -545,6 +551,29 @@ do_condition_check([{Condition, Error}|Tail], _PrevErr, true) ->
 majority(N) ->
     (N div 2) + 1.
 
+-spec get_vars(VarList :: [atom()], Ledger :: blockchain_ledger_v1:ledger()) -> #{atom() => any()}.
+get_vars(VarList, Ledger) ->
+    lists:foldl(
```
evanmcc, line 556: an optimization here would be to pull the mode and nonce once and then do the below query for each var, so we're not refetching everything dozens of times.

vihu: Update 9710a47

src/blockchain_utils.erl
```diff
@@ -545,6 +551,45 @@ do_condition_check([{Condition, Error}|Tail], _PrevErr, true) ->
...
+    e2qc:cache(
+        ?VAR_CACHE,
+        {Mode, VarsNonce, VarName},
+        fun() ->
+            get_var_(VarName, Ledger)
+        end
+    ).
+
+-spec get_var(VarName :: atom(), Ledger :: blockchain_ledger_v1:ledger()) -> any().
+get_var(VarName, Ledger) ->
```
Vagabond, line 577: Can we point blockchain:config at this?

src/blockchain_utils.erl
```diff
@@ -545,6 +551,45 @@ do_condition_check([{Condition, Error}|Tail], _PrevErr, true) ->
...
+    {ok, VarsNonce} = blockchain_ledger_v1:vars_nonce(Ledger),
```
Vagabond, line 578: Making this a persistent value via ETS or a counter would be good (just increment it on absorb and set it when doing ledger_ats)

src/blockchain_utils.erl
```diff
@@ -545,6 +551,45 @@ do_condition_check([{Condition, Error}|Tail], _PrevErr, true) ->
...
+    Mode = blockchain_ledger_v1:mode(Ledger),
```
Vagabond, line 579: You do not need to distinguish between active and delayed, just between aux and anything else.

src/blockchain_utils.erl
```diff
@@ -545,6 +551,45 @@ do_condition_check([{Condition, Error}|Tail], _PrevErr, true) ->
...
+        ?VAR_CACHE,
+        {Mode, VarsNonce, VarName},
+        fun() ->
+            get_var_(VarName, Ledger)
+        end
+    ).
+
+-spec get_var_(VarName :: atom(), Ledger :: blockchain_ledger_v1:ledger()) -> any().
+get_var_(VarName, Ledger) ->
+    {ok, Value} = blockchain:config(atom_to_binary(VarName), Ledger),
```
Vagabond, line 590: Use `blockchain_ledger_v1:config(VarName, Ledger)` here?

src/blockchain_utils.erl
vihu: Ok, I will switch the key for the cache to be: `{has_aux, vars_nonce, var_name}`

src/blockchain_utils.erl
vihu: Added 1734847

src/blockchain_utils.erl
vihu: Added 1734847

src/blockchain_utils.erl
evanmcc: I suspect that we can live without it for now, as the semantics for that are pretty tricky.

evanmcc (approved):

PR #1031, opened 2021-10-05
don't let validation continue if it cannot succeed

evanmcc:
sometimes we get receipts that we cannot possibly validate, since they would require fetching ledgers past the trailing ledger.  this causes noisy crashes in the logs, and we should handle it better.


Vagabond (approved):

xandkar:
> cc @xandkar for the work he's doing on reworking txn validation.

ACK

PR #1034, opened 2021-10-07
Add witness reward decay rate as a legal var and add an exclusion count

Vagabond:
So for example, setting the exclusion to 4 would mean your first 4
witness events per epoch would be at full value and the exponential
decay woulds only kick in for the 5th reward and above.

Example of reward curve:

```erlang
18> WitnessDecay = fun(Count, Exclusion, DecayRate) -> case Count < Exclusion of true -> 1; false -> math:exp((Count - Exclusion) * -1 * DecayRate) end end.
#Fun<erl_eval.19.126501267>

19> [WitnessDecay(N, 3, 0.08) || N <- lists:seq(0, 14)].
[1,1,1,1.0,0.9231163463866358,0.8521437889662113,
 0.7866278610665535,0.7261490370736909,0.6703200460356393,
 0.6187833918061408,0.5712090638488149,0.5272924240430485,
 0.4867522559599717,0.44932896411722156,0.4147829116815814]

20> lists:sum([WitnessDecay(N, 3, 0.08) || N <- lists:seq(0, 14)]).
11.02650609098551
```

Note that the decay is calculated over the *current* # of witness events in the epoch, so for the initial witness that will be 0. Also note that for an exponential decay curve the first value is 1.0, so an exclusion of 3 protects the first four witness rewards in the epoch from any decay.

For completeness here's some examples with no exclusion:

```erlang
33> [WitnessDecay(N, 0, 0.08) || N <- lists:seq(0, 14)].
[1.0,0.9231163463866358,0.8521437889662113,
 0.7866278610665535,0.7261490370736909,0.6703200460356393,
 0.6187833918061408,0.5712090638488149,0.5272924240430485,
 0.4867522559599717,0.44932896411722156,0.4147829116815814,
 0.38289288597511206,0.35345468195878016,0.32627979462303947]

34> lists:sum([WitnessDecay(N, 0, 0.08) || N <- lists:seq(0, 14)]).
9.08913345354244
```

### DIFF
--- a/src/transactions/v1/blockchain_txn_vars_v1.erl
+++ b/src/transactions/v1/blockchain_txn_vars_v1.erl
@@ -1012,6 +1012,10 @@ validate_var(?hip15_tx_reward_unit_cap, Value) ->
     %% According to HIP-15, the cap should be set to 2.0
     %% 5.0 is just for future proofing if need be
     validate_float(Value, "hip15_tx_reward_unit_cap", 0.0, 5.0);
+validate_var(?witness_reward_decay_rate, Value) ->
+    validate_float(Value, "witness_reward_decay_rate", 0.0, 5.0);
+validate_var(?witness_reward_decay_exclusion, Value) ->
+    validate_int(Value, "witness_reward_decay_exclusion", 0, 10, false);
 
 %% bundle vars
 validate_var(?max_bundle_size, Value) ->
--- a/src/transactions/v2/blockchain_txn_rewards_v2.erl
+++ b/src/transactions/v2/blockchain_txn_rewards_v2.erl
@@ -694,6 +694,13 @@ get_reward_vars(Start, End, Ledger) ->
                                  _ -> undefined
                              end,
 
+    WitnessRewardDecayExclusion =
+        case blockchain:config(?witness_reward_decay_exclusion, Ledger) of
+            {ok, Exc} -> Exc;
+            _ -> undefined
+        end,
+
+
     EpochReward = calculate_epoch_reward(Start, End, Ledger),
     #{
         monthly_reward => MonthlyReward,
@@ -716,7 +723,8 @@ get_reward_vars(Start, End, Ledger) ->
         election_interval => ElectionInterval,
         election_restart_interval => ElectionRestartInterval,
         block_time => BlockTime,
-        witness_reward_decay_rate => WitnessRewardDecayRate
+        witness_reward_decay_rate => WitnessRewardDecayRate,
+        witness_reward_decay_exclusion => WitnessRewardDecayExclusion
     }.
 
 -spec calculate_epoch_reward(pos_integer(), pos_integer(), blockchain_ledger_v1:ledger()) -> float().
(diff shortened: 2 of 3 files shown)


src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1012,6 +1012,10 @@ validate_var(?hip15_tx_reward_unit_cap, Value) ->
     %% According to HIP-15, the cap should be set to 2.0
     %% 5.0 is just for future proofing if need be
     validate_float(Value, "hip15_tx_reward_unit_cap", 0.0, 5.0);
+validate_var(?witness_reward_decay_rate, Value) ->
+    validate_float(Value, "hip15_tx_reward_unit_cap", 0.0, 5.0);
```
jeffgrunewald, line 1016: these are using the same binary name because they're two different versions/generations of the same setting, yes?

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -1484,9 +1491,18 @@ witness_decay(Count, Vars) ->
         {ok, undefined} ->
             1;
         {ok, DecayRate} ->
-            Scale = math:exp(Count * -1 * DecayRate),
-            lager:info("scaling witness reward by ~p", [Scale]),
-            Scale;
+            Exclusion = case maps:find(witness_reward_decay_exclusion, Vars) of
+                            {ok, undefined} -> 0;
```
jeffgrunewald, line 1495: any reason not to push this `0` up into the Vars config map? is it accessed anywhere else where we'd want to reference that it wasn't explicitly defined and thus defaulting to 0? then just set `Exclusion = maps:get(witness_reward_decay_exclusion, Vars)`?

jeffgrunewald (commented):
if we currently have a uniform witness decay rate and exclusion count would we want to (in the future potentially) consider an additional parameter that incorporates the local density of hotspots doing witness events to further tweak this for sparse rural vs. dense urban environments?

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -1484,9 +1491,18 @@ witness_decay(Count, Vars) ->
...
+                            {ok, ExclusionValue} -> ExclusionValue
+                        end,
+            case Count < Exclusion of
+                true ->
+                    1;
+                false ->
+                    Scale = math:exp((Count - Exclusion) * -1 * DecayRate),
+                    lager:info("scaling witness reward by ~p", [Scale]),
```
evanmcc, line 1503: debug, I think

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -1484,9 +1492,18 @@ witness_decay(Count, Vars) ->
...
```
Vagabond, line 1504:
```suggestion
                    lager:debug("scaling witness reward by ~p", [Scale]),
```

PR #1051, opened 2021-10-15
Make it so the worker check for it's own expiration

macpie:
This avoid left over state channel that missed their expiration (restart or other). This was seen on dev

### DIFF
--- a/src/state_channel/blockchain_state_channels_worker.erl
+++ b/src/state_channel/blockchain_state_channels_worker.erl
@@ -39,6 +39,7 @@
 -define(OVERSPENT, overspent).
 
 -record(state, {
+    parent :: pid(),
     id :: blockchain_state_channel_v1:id(),
     state_channel :: blockchain_state_channel_v1:state_channel(),
     skewed :: skewed:skewed(),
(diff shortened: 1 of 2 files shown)


src/state_channel/blockchain_state_channels_worker.erl
```diff
@@ -198,6 +204,28 @@ terminate(Reason, #state{id=ID, state_channel=SC, skewed=Skewed, db=DB, owner={_
 %% Internal Function Definitions
 %% ------------------------------------------------------------------
 
+-spec check_expiration(WorkerPid :: pid(), BlockHash :: blockchain_block:hash(), State :: state()) -> ok.
+check_expiration(WorkerPid, BlockHash, #state{id=ID, state_channel=SC, chain=Chain}) ->
+    erlang:spawn(fun() ->
```
jeffgrunewald, line 209: How come this is run in a separate process instead of just in the current worker process and then call `blockchain_state_channels_worker:expire(self())`? Are you need to keep allowing messages to the worker inbox in between this `handle_info/2` and receiving the `expire`?

macpie: The async nature was there before I have just moved it to the worker. This is done in a spawn because fetching the block (sadly) is expensive and I don't want to lock myself for that long. An option I saw would be to send the height in the blockchain event message but we would need to update everything using it :/

PR #1055, opened 2021-10-20
Witness reward decay testing

jeffgrunewald:
Introduces a framework for evaluating the effect of modifying the `witness_reward_decay_rate` and `witness_reward_decay_exclusion` chain variables on PoC witness node rewards.

Also explicitly sets the initialization of the e2qc var cache in the primary blockchain supervisor _not_ be initialized by default during testing as this causes concurrency collisions in the cache that crashes the ct suite run.

### DIFF
--- a/test/witness_reward_decay_SUITE.erl
+++ b/test/witness_reward_decay_SUITE.erl
@@ -208,6 +208,15 @@ init_per_testcase(TestCase, Config0) ->
     ),
 
     {ok, Height} = blockchain:height(Chain),
+    ?assertEqual({ok, 11}, blockchain:height(Chain)),
+
+    DecayRate =
+        case blockchain_utils:get_var(?witness_reward_decay_rate, Ledger) of
+            {ok, DecayRateVar} -> DecayRateVar;
+            {error, not_found} -> not_found
+        end,
+    ct:print("Decay Rate Var : ~p", [DecayRate]),
+
     {ok, RewardsMd} = blockchain_txn_rewards_v2:calculate_rewards_metadata(1, Height, Chain),
     WitnessRewards = maps:get(poc_witness, RewardsMd),
     ct:print("WitnessRewards : ~p", [WitnessRewards]),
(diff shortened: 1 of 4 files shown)


test/witness_reward_decay_SUITE.erl
```diff
@@ -0,0 +1,386 @@
...
+%% group setup
+%%--------------------------------------------------------------------
+
+init_per_group(Group, Config) ->
+    ExtraVars =
+        case Group of
+            no_vars ->
+                #{};
+            with_decay ->
+                #{?witness_reward_decay_exclusion => 3}
```
vihu, line 75: This is proposed 3 or 4?

test/witness_reward_decay_SUITE.erl
```diff
@@ -0,0 +1,386 @@
...
+init_per_suite(Config) ->
+    {ok, StorePid} = blockchain_test_reward_store:start(),
+    blockchain_test_reward_store:insert(witness_shares, #{}),
+    [{store_pid, StorePid} | Config].
+
+%%--------------------------------------------------------------------
+%% suite teardown
+%%--------------------------------------------------------------------
+
+end_per_suite(_Config) ->
```
vihu, line 100: I think we should have atleast _some_ assertion in either end_per_suite or end_per_group. Something like comparing the results between the various decay_* test runs. We know that the higher the decay the lesser the witness shares would be. Right?

test/witness_reward_decay_SUITE.erl
jeffgrunewald: I recall you mentioned 4 exclusions (meaning the first 5 witness shares are awarded at full value. I’ll construct one additional beacon event and feed it through.

test/witness_reward_decay_SUITE.erl
jeffgrunewald: If I’m reading it correctly it’s only the witnesses that experience 6 or more witness events that will see a reduction in shares and any other witnesses will actually see a _gain_ in shares since the pool of shares is fixed and distributed amongst all witnesses. 

That being said, these should be deterministic for each test case so I can make an assertion on the actual shares vs what we expect and that wouldn’t fluctuate without the rates or exclusions changing. What would be the preferred place to see the assertions called out? Any particular reason to make them at the end of each case vs. each group vs. the entire suite?

vihu: I believe it should be at the end of the suite. At that point all the groups (and their corresponding tests cases) would have completed. So the order goes like:

`suite -> group1 (t1, t2) -> end_group1 -> group2 (t1, t2) -> end_group2 -> end suite`

evanmcc (approved):

PR #1056, opened 2021-10-21
revert var cache

evanmcc:
this PR turns the var cache into a noop, retaining the poc var selection and list->map stuff that came in with #1030 rather than reverting it.

I think that a cache is a good idea, and have some ideas for making it safer/faster than this one.

### DIFF
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -620,8 +620,8 @@ get_vars(VarList, Ledger) ->
       end, #{}, VarList).
 
 -spec get_var_(VarName :: atom(),
-               HasAux :: boolean(),
-               VarsNonce :: non_neg_integer(),
+               IsAux :: atom(),         % XXX: Temporary
+               VarsNonce :: atom(),     % XXX: Temporary
                Ledger :: blockchain_ledger_v1:ledger()) -> {ok, any()} | {error, any()}.
 get_var_(VarName, _HasAux, _VarsNonce, Ledger) ->
     get_var_(VarName, Ledger).


src/blockchain_utils.erl
```diff
@@ -626,49 +623,31 @@ get_vars(VarList, Ledger) ->
...
+    get_var_(VarName, Ledger).
 
 -spec get_var_(VarName :: atom(), Ledger :: blockchain_ledger_v1:ledger()) -> {ok, any()} | {error, any()}.
 get_var_(VarName, Ledger) ->
     blockchain_ledger_v1:config(VarName, Ledger).
 
 -spec var_cache_stats() -> list().
 var_cache_stats() ->
-    e2qc:stats(?VAR_CACHE).
+    %%e2qc:stats(?VAR_CACHE).
```
vihu, line 639: Debugging left around?

src/blockchain_utils.erl
```diff
@@ -606,69 +606,48 @@ do_condition_check([{Condition, Error}|Tail], _PrevErr, true) ->
...
-    {ok, VarsNonce} = blockchain_ledger_v1:vars_nonce(Ledger),
-    IsAux = blockchain_ledger_v1:is_aux(Ledger),
     lists:foldl(
       fun(VarName, Acc) ->
-              %% NOTE: This isn't ideal but in order for get_var/2 to
-              %% correspond with blockchain:config/2, it returns {ok, ..} | {error, ..}
-              %% So we just put undefined for any error lookups here.
-              %% The callee must handle those situations.
-              case get_var_(VarName, IsAux, VarsNonce, Ledger) of
+              case get_var_(VarName, isaux, varsnonce, Ledger) of
```
andymck, line 616: why not just remove isaux and varsnonce both here and as a param of get_var_ ?  Guess maybe hoping to bring it back into play

andymck (approved):
dont have full context but end result lgtm

PR #1058, opened 2021-10-22
allow the s3 code to get the latest snap from s3

evanmcc:
use the latest data written in https://github.com/helium/blockchain-etl/pull/249 to dynamically fetch the latest snap on s3.  probably need to do a bit more work around checking the height.

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -1003,27 +1003,35 @@ start_snapshot_sync(Hash, Height, Peer,
                     #state{blockchain=Chain, swarm_tid=SwarmTID, sync_paused=SyncPaused}) ->
     spawn_monitor(fun() ->
                           try
-                              BaseUrl = application:get_env(blockchain, s3_base_url, undefined),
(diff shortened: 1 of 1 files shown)


jadeallenx (approved):

src/blockchain_worker.erl
```diff
@@ -1003,27 +1003,35 @@ start_snapshot_sync(Hash, Height, Peer,
                     #state{blockchain=Chain, swarm_tid=SwarmTID, sync_paused=SyncPaused}) ->
     spawn_monitor(fun() ->
                           try
-                              BaseUrl = application:get_env(blockchain, s3_base_url, undefined),
+                              BaseUrl = application:get_env(blockchain, snap_source_base_url, undefined),
```
abhay, line 1006: Does this need to change?

madninja: Oh you changed it because you don’t like the use of “s3” in a url completely unrelated to S3. Makes sense

PR #1068, opened 2021-11-02
Make JSON decoding configurable

jadeallenx:
Problem to solve: We are currently hardcoding jsx as the JSON library in core, and that's fine in the context of miner where jsx is already a defined dep - but we use core in other contexts where jsx is not a defined dependency. Additionally, we are copypasta http client code which we ought to centralize for maintenance purposes (everyone knows httpc only gets more complex over time)

Solution: Pull httpc into a common module. Make json decoding configurable in the environment but provide a default method based on jsx.


jadeallenx:
In the context of #1254 this needs to be reworked quite a bit

PR #1071, opened 2021-11-02
Don't dump entire state channel to log

michaeldjeffrey:
Purchasing is less of a problem because there's less of a chance of
running a full system with debug active, but the warning coming with a
fully formatted 2000 actor state channel can get pretty heavy when you
have the maximu number of state channels all filled up.

### DIFF
--- a/src/state_channel/blockchain_state_channels_worker.erl
+++ b/src/state_channel/blockchain_state_channels_worker.erl
@@ -257,12 +257,12 @@ offer(
                 {error, _Reason} ->
                     lager:warning(
                         "dropping this packet because: ~p ~p",
-                        [_Reason, lager:pr(SC, blockchain_state_channel_v1)]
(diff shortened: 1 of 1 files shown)


src/state_channel/blockchain_state_channels_worker.erl
```diff
@@ -257,12 +257,12 @@ offer(
                 {error, _Reason} ->
                     lager:warning(
                         "dropping this packet because: ~p ~p",
-                        [_Reason, lager:pr(SC, blockchain_state_channel_v1)]
+                        [_Reason, blockchain_state_channel_v1:id(SC)]
```
macpie, line 260: sorry picky but can we make it the SC name instead of ID?

Vagabond (approved):

vihu (approved):

PR #1081, opened 2021-11-09
Teardown H3 region cache on var txn absorb

vihu:
Problem
----
We need to teardown the H3 region cache whenever a new chain var is absorbed on chain.

Solution
----
I've added the teardown for region cache in `save_threshold_txn` fun in ledger, that's the thing which gets invoked on var txn absorb afaict.

PS: There's some brokenness with the calls to an undefined function, that's fixed here too to make dialyzer happy.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -1121,6 +1121,8 @@ save_threshold_txn(Txn, Ledger) ->
     DefaultCF = default_cf(Ledger),
     Bin = term_to_binary(Txn),
     Name = threshold_name(Txn),
+    %% This is done to invalidate the region cache on chain var txn absorption
+    _ = blockchain_region_v1:teardown_h3_to_region_cache(),
     cache_put(Ledger, DefaultCF, Name, Bin).
 
 threshold_name(Txn) ->
--- a/src/region/blockchain_region_v1.erl
+++ b/src/region/blockchain_region_v1.erl
@@ -7,7 +7,7 @@
 
 -include("blockchain_vars.hrl").
 
--export([get_all_regions/1, h3_to_region/2, h3_in_region/3]).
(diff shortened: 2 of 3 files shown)


src/state_channel/blockchain_state_channels_worker.erl
```diff
@@ -239,7 +239,7 @@ offer(
                 "dropping this packet because it will overspend DC ~p, (cost: ~p, total_dcs: ~p)",
                 [DCAmount, NumDCs, TotalDCs]
             ),
-            %% This allow for packets (accepted offer) to come threw 
+            %% This allow for packets (accepted offer) to come threw
```
abhay, line 242: Nit: other preexisting typo.

jadeallenx (approved):

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1121,6 +1121,8 @@ save_threshold_txn(Txn, Ledger) ->
     DefaultCF = default_cf(Ledger),
     Bin = term_to_binary(Txn),
     Name = threshold_name(Txn),
+    %% This is done to invalidate the region cache on chain var txn absorption
+    _ = blockchain_region_v1:teardown_h3_to_region_cache(),
```
jeffgrunewald, line 1125: Is this match to silence dialyzer?

src/region/blockchain_region_v1.erl
```diff
@@ -107,3 +107,6 @@ h3_in_region_(H3, RegionVar, Ledger) ->
         _ ->
             {error, {region_var_not_set, RegionVar}}
     end.
+
+teardown_h3_to_region_cache() ->
```
jeffgrunewald, line 111:
```suggestion
-spec teardown_h3_to_region_cache() -> ok
teardown_h3_to_region_cache() ->
```

vihu:
We don't need this because the region cache is tied to the `vars_nonce` key. This needs to be fixed on the miner side in `miner_lora`

PR #1082, opened 2021-11-09
Demote error log spam to debug level

vihu:
Too much error log spam when the region is unknown in POCV11 land, this demotes it to a debug msg if we need to debug it later.

### DIFF
--- a/src/transactions/v2/blockchain_txn_rewards_v2.erl
+++ b/src/transactions/v2/blockchain_txn_rewards_v2.erl
@@ -1202,7 +1202,8 @@ poc_witness_reward(Txn, AccIn,
                 Path)
     catch
         What:Why:ST ->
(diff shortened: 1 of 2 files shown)


jadeallenx (approved):

jeffgrunewald (approved):

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -1202,7 +1202,8 @@ poc_witness_reward(Txn, AccIn,
                 Path)
     catch
         What:Why:ST ->
-            lager:error("failed to calculate poc_witnesses_rewards, error ~p:~p:~p", [What, Why, ST]),
+            lager:error("error: ~p", [What]),
```
evanmcc, line 1205: I think that you want `Why` here, `What` is going to be `error` almost every time.

evanmcc (approved):

PR #1089, opened 2021-11-11
Lookup for_region with e2qc

vihu:
Put `for_region` lookup behind cache.


vihu:
Already handled differently, closing.

PR #1091, opened 2021-11-15
Do not send a stale state channel

macpie:


macpie:
I have also removed `handlers` they are unused here anyway

michaeldjeffrey (approved):

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -74,21 +74,26 @@ get_all() ->
...
+                fun({SC, SCState, Pid}) ->
+                    ID = blockchain_state_channel_v1:id(SC),
+                    try blockchain_state_channels_worker:get(Pid) of
+                        UpToDateSC ->
+                            {ID, {UpToDateSC, SCState, Pid}}
+                        
+                    catch _C:_E ->
+                        Name = blockchain_utils:addr2name(ID),
+                        lager:error("failed to get sc ~p(~p) ~p/~p", [Name, Pid, _C, _E]),
+                        lager:error("~p", [recon:info(Pid)]),
```
vihu, line 91: Is this okay? I guess depends on the occurrence?

vihu (approved):

src/state_channel/blockchain_state_channels_server.erl
jeffgrunewald: calling recon:info/1 or returning undefined where we used to return that nested tuple?

vihu: Just error spam, the undefined later would get filtered out.

PR #1093, opened 2021-11-16
add gossip size limit

evanmcc:
I pushed a few things somewhat unwisely to the main branch.

### DIFF
--- a/src/handlers/blockchain_gossip_handler.erl
+++ b/src/handlers/blockchain_gossip_handler.erl
@@ -71,27 +71,34 @@ handle_gossip_data(_StreamPid, Data, [SwarmTID, Blockchain]) ->
                         end
                 end;
             #blockchain_gossip_block_pb{from=From, block=BinBlock} ->
-                Block = blockchain_block:deserialize(BinBlock),
-                case blockchain_block:type(Block) of
-                    undefined ->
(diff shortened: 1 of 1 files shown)


Vagabond (approved):

src/handlers/blockchain_gossip_handler.erl
```diff
@@ -71,27 +71,34 @@ handle_gossip_data(_StreamPid, Data, [SwarmTID, Blockchain]) ->
...
-                Block = blockchain_block:deserialize(BinBlock),
-                case blockchain_block:type(Block) of
-                    undefined ->
-                        lager:notice("gossip_handler unknown block: ~p", [Block]);
-                    _ ->
-                        case blockchain:has_block(Block, Blockchain) of
-                            true ->
-                                %% already got this block, just return
-                                ok;
+                SizeLimit = application:get_env(blockchain, gossip_block_limit_mb, 25) * 1024 * 1024,
```
jeffgrunewald, line 74: should this rebase off pull 1094 since it's already merged in and consolidate these size checks or is it supposed to be a different size (25 instead of 50)?

evanmcc: this is basically disabling v1 gossip, since all it is is the doom block, so the size doesn't really matter.

PR #1097, opened 2021-11-16
Remove two more unnecessary uses of head_block

Vagabond:


madninja (approved):

andymck (approved):

src/handlers/blockchain_sync_handler.erl
```diff
@@ -163,7 +163,7 @@ handle_data(server, Data, #state{blockchain=Blockchain, batch_size=BatchSize,
                             {stop, normal, State, Msg};
                         _ ->
                             {noreply, State#state{batches_sent=Sent+1,
-                                                  block=lists:last(Blocks),
+                                                  block=element(1, lists:last(Blocks)),
```
madninja, line 166: Can you break this out to avoid using element? Makes it easier to read the intent in the future (and is better for dialyzer?)

src/handlers/blockchain_sync_handler.erl
```diff
@@ -182,7 +182,7 @@ handle_data(server, Data, #state{blockchain=Blockchain, batch_size=BatchSize,
...
```
madninja, line 185: Ditto

madninja (approved):

src/handlers/blockchain_fastforward_handler.erl
```diff
@@ -110,10 +110,10 @@ init(server, _Conn, [_, _HandlerModule, [Path, Blockchain]] = _Args) ->
...
     #blockchain_sync_hash_pb{hash=Hash} =
     blockchain_sync_handler_pb:decode_msg(Data, blockchain_sync_hash_pb),
-    case blockchain:get_block(Hash, Blockchain) of
-        {ok, Block} ->
-            Blocks = blockchain:build(Block, Blockchain, 200),
-            Msg0 = #blockchain_sync_blocks_pb{blocks=[blockchain_block:serialize(B) || B <- Blocks]},
+    case blockchain:get_block_height(Hash, Blockchain) of
+        {ok, Height} ->
+            Blocks = blockchain:build(Height, Blockchain, 200),
+            Msg0 = #blockchain_sync_blocks_pb{blocks=[B || {_H, B} <- Blocks]},
```
jeffgrunewald, line 116: These no longer need to be serialized? It looks like `build/2` is producing raw blocks in that tuple accumulator

jeffgrunewald (commented):
Seems like the purpose of these is to _avoid_ serializing these blocks as much as possible but just wanna make sure that’s intended for blocks you’re putting into a pb record (with the intention to send over the wire I imagine)

src/handlers/blockchain_fastforward_handler.erl
Vagabond: yeah, those are already serialized

madninja (approved):

jeffgrunewald (approved):

src/blockchain.erl
```diff
@@ -1337,19 +1332,21 @@ fees_since(_Height, _CurrentHeight, _Chain) ->
...
-build(StartingBlock, Blockchain, N, Acc) ->
-    Height = blockchain_block:height(StartingBlock) + 1,
-    case ?MODULE:get_block(Height, Blockchain) of
+build(Height, Blockchain, N, Acc) ->
+    case ?MODULE:get_raw_block(Height, Blockchain) of
         {ok, NextBlock} ->
-            build(NextBlock, Blockchain, N-1, [NextBlock|Acc]);
-        {error, _Reason} ->
+            build(Height + 1, Blockchain, N-1, [{Height, NextBlock}|Acc]);
+        _ ->
```
xandkar, line 1349: Why the switch to catch-all pattern?

src/handlers/blockchain_sync_handler.erl
```diff
@@ -162,8 +162,9 @@ handle_data(server, Data, #state{blockchain=Blockchain, batch_size=BatchSize,
                         true ->
                             {stop, normal, State, Msg};
                         _ ->
+                            {LastHeight, _LastBlock} = lists:last(Blocks),
                             {noreply, State#state{batches_sent=Sent+1,
-                                                  block=lists:last(Blocks),
+                                                  block=LastHeight,
```
vihu, line 167: nit: change `block` to `last_block_height` or something?

src/blockchain.erl
Vagabond: because it can return not_found or {error, _}

vihu (approved):

evanmcc (approved):

PR #1109, opened 2021-11-17
Adt/v2 plausible eager gossip

Vagabond:


src/blockchain.erl
```diff
@@ -2610,17 +2613,76 @@ is_block_plausible(Block, Chain) ->
             false
     end.
 
+save_plausible_blocks(Blocks, #blockchain{db=DB}=Chain) ->
+    %% XXX ASSUMPTION this is only called from the sync pid and thus we won't check
```
evanmcc, line 2617: perhaps call `blockchain_lock:check()` here?
https://github.com/helium/blockchain-core/blob/master/src/blockchain_lock.erl#L60

src/blockchain.erl
```diff
@@ -818,10 +821,10 @@ get_raw_block(Height, #blockchain{db=DB, heights=HeightsCF}=Blockchain) ->
 %% checks if we have this block in any of the places we store blocks
 %% note that if we only use this for gossip we will never see assume valid blocks
 %% so we don't need to check for them
-has_block(Block, #blockchain{db=DB, blocks=BlocksCF,
+has_block(Block, #blockchain{db=DB, info=InfoCF,
```
vihu, line 824:
```suggestion
-spec has_block(Block :: blockchain_block:block(), Chain :: blockchain()) -> boolean() | {error, any()}.
has_block(Block, #blockchain{db=DB, info=InfoCF,
```

src/blockchain.erl
```diff
@@ -2610,17 +2614,76 @@ is_block_plausible(Block, Chain) ->
             false
     end.
 
+save_plausible_blocks(Blocks, #blockchain{db=DB}=Chain) ->
+    %% XXX ASSUMPTION this is only called from the sync pid and thus we won't check
+    %% for a blockchain lock
+    {ok, Batch} = rocksdb:batch(),
+    Result = lists:foldl(fun(Block, Acc) ->
+                          case has_block(Block, Chain) == false andalso is_block_plausible(Block, Chain) of
```
vihu, line 2622: `has_block/2` may return an error...perhaps rarely but it can?

src/blockchain.erl
```diff
@@ -2610,17 +2614,76 @@ is_block_plausible(Block, Chain) ->
...
+
+save_plausible_block(Batch, Block, Hash, #blockchain{db=DB, plausible_blocks=PlausibleBlocks}) ->
     case rocksdb:get(DB, PlausibleBlocks, Hash, []) of
         {ok, _} ->
             %% already got it, thanks
             exists;
         _ ->
-            rocksdb:put(DB, PlausibleBlocks, Hash, blockchain_block:serialize(Block), [{sync, true}])
+            Height = blockchain_block:height(Block),
+            PlausiblesAtThisHeight = case rocksdb:get(DB, PlausibleBlocks, <<Height:64/integer-unsigned-big>>, []) of
```
vihu, line 2661: This is not the same as calling `get_raw_plausibles/2`?

src/blockchain.erl
```diff
@@ -2610,17 +2614,76 @@ is_block_plausible(Block, Chain) ->
...
+                {ok, BinList} ->
+                    binary_to_term(BinList);
+                _ ->
+                    []
+            end,
+            rocksdb:batch_put(Batch, PlausibleBlocks, Hash, blockchain_block:serialize(Block)),
+            rocksdb:batch_put(Batch, PlausibleBlocks, <<Height:64/integer-unsigned-big>>, term_to_binary([Hash|PlausiblesAtThisHeight]))
     end.
 
+remove_plausible_block(#blockchain{db=DB, plausible_blocks=CF}, Batch, Hash, Height) ->
```
vihu, line 2671:
```suggestion
-spec remove_plausible_block(blockchain(), Batch :: rockdb:batch_handle(), Hash :: binary(), Height :: non_neg_integer()) -> ok.
remove_plausible_block(#blockchain{db=DB, plausible_blocks=CF}, Batch, Hash, Height) ->
```

src/blockchain.erl
Vagabond: We cannot take the lock as we don't want to be saving plausible blocks under the lock. Instead we want to do an unlocked write from the singleton sync client process.

src/blockchain.erl
Vagabond: It shouldn't though

src/blockchain.erl
Vagabond: it might be, I probably had not written that function yet.

src/blockchain.erl
Vagabond: alternatively we'd add a lock for writing plausible blocks and take that other lock here.

src/blockchain.erl
Vagabond: No, this is just to get the plausible hashes at this height and append to them.

madninja (approved):

PR #1111, opened 2021-11-18
Try to make snap loading more memory efficent

Vagabond:
* Add option to disable 'rocksdb bulk load' mode
* Try to keep less on the stack at once
* Try to tail recurse more
* Try an interator for packed raw key/value fields
* Handle the fact that blocks are a list of t2b'd list of binaries by writing a custom decoder

### DIFF
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -451,36 +451,41 @@ deserialize(DigestOpt, <<Bin0/binary>>) ->
 -spec import(blockchain:blockchain(), binary(), snapshot()) ->
     blockchain_ledger_v1:ledger().
 import(Chain, SHA, #{version := v6}=Snapshot) ->
+    print_memory(),
     CLedger = blockchain:ledger(Chain),
     Dir = blockchain:dir(Chain),
+
+    %% potentially open ledger with compaction disabled so
+    %% we can bulk load
+
     %% clean the ledger in case we had a partial snapshot load
     blockchain_ledger_v1:clean(CLedger),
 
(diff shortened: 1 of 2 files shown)


src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -451,36 +451,41 @@ deserialize(DigestOpt, <<Bin0/binary>>) ->
...
-                ]
-               ),
+      Dir,
+      false,
+      blockchain:db_handle(Chain),
+      blockchain:blocks_cf(Chain),
+      blockchain:heights_cf(Chain),
+      blockchain:info_cf(Chain),
+      %% these options taken from rocksdb's PrepareForBulkLoad()
+      %% and are only used if allow_bulk_snapshot_loads is true
```
JayKickliter, line 472: Am I reading this correctly: if `application:get_env(blockchain, allow_bulk_snapshot_loads, true)` returns `false` then the entirety of the options will be `[]`?

Vagabond: yes

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -1323,6 +1353,16 @@ bin_pair_to_iolist({<<K/binary>>, V}) ->
         V
     ].
 
+mk_bin_iterator(<<>>) ->
```
jeffgrunewald, line 1356: i'm confused why this is function isn't defined in the same area that calls/decodes the binary iterator to walk the list of functions; is are those binary encoding byte sizes and formats specific to a v1 snapshot?

PR #1112, opened 2021-11-18
Properly release/decrement the blockchain lock in check_plausible_blocks

Vagabond:


src/blockchain.erl
```diff
@@ -2756,7 +2756,8 @@ check_plausible_blocks(#blockchain{db=DB}=Chain, GossipedHash) ->
                                   remove_plausible_block(Chain, Batch, Hash, blockchain_block:height(Block))
                           end
                   end, SortedBlocks),
-    rocksdb:write_batch(DB, Batch, [{sync, true}]).
+    rocksdb:write_batch(DB, Batch, [{sync, true}]),
+    blockchain_lock:release().
```
jeffgrunewald, line 2760: was rocks supposed to have been doing this implicitly prior to this?

Vagabond: no, blockchain lock is our own mutex. We don't use rocks transactionally in core, unfortunately, because we didn't have that mode available via the nif at that point.

madninja (approved):

PR #1116, opened 2021-11-19
[DNM] Test batch signature verification

vihu:


src/blockchain.erl
```diff
@@ -1047,44 +1047,48 @@ can_add_block(Block, Blockchain) ->
...
-                                                        false ->
-                                                            {error, wrong_txn_order};
-                                                        true ->
-                                                            {true, IsRescue}
-                                                    end;
-                                                true ->
-                                                    {true, IsRescue}
-                                            end
-                                    end
+                                    verify_signatures(Block, ConsensusAddrs, Sigs, N - F, KeyOrKeys, Txns)
```
vihu, line 1050: I can move this back if its annoying to review, I separated this out from here to basically do comparison tests between the two signature verification schemes.

PR #1118, opened 2021-11-19
Parallelize getting block hash for state channels on sc_server start

michaeldjeffrey:
For large blocks, this can take a long enough time that gen_server calls
timeout and applications restart themselves before any process is done.


macpie:
I will include this in https://github.com/helium/blockchain-core/pull/1119

PR #1119, opened 2021-11-20
Fixup cache to be started by SUP only

macpie:


Vagabond:
This looks reasonable, but will all the processes using the cache notice if it gets wiped by a supervisor crash?

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -385,26 +375,49 @@ handle_offer_(Offer, Ledger, HandlerPid, Retry) ->
...
+                found -> {true, {0, Pid}}
+            end
+        end,
+        maps:values(Actives)
+    ),
+    Sorted = lists:sort(fun({A, _}, {B, _}) -> A < B end, Filtered),
+    case Sorted of
+        [] ->
+            {error, not_found};
+        [{Spots, Pid}|[]] when Spots < MaxActorsAllowed/10 ->
```
michaeldjeffrey, line 415: I think this is meant to be
```suggestion
        [{Spots, Pid}|_] when Spots < MaxActorsAllowed/10 ->
```

michaeldjeffrey: This will only match the last Sorted item, but I'm thinking now that might be what we want. 

When there's only 1 state channel left and there's less than 1/10 the spots allowed.

macpie: My intent here is: if we have only 1 active left with room (and we have only 10% left) lets go an attempt to activate a new one

src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -81,14 +81,14 @@ get_actives() ->
             blockchain_utils:pmap(
                 fun({SC, SCState, Pid}) ->
                     ID = blockchain_state_channel_v1:id(SC),
-                    try blockchain_state_channels_worker:get(Pid) of
+                    try blockchain_state_channels_worker:get(Pid, timer:seconds(1000)) of
```
macpie, line 84:
```suggestion
                    try blockchain_state_channels_worker:get(Pid, timer:seconds(1)) of
```

jadeallenx (commented):
It isn't clear to me if this cache is truly a cache or if it's holding stateful data that if it goes away would cause issues with current channels state

If it is holding critical state shouldn't that be durably stored somewhere in case the cache is terminated?

src/state_channel/blockchain_state_channel_common.erl
```diff
@@ -323,9 +324,11 @@ handle_offer(Offer, Time, HandlerState) ->
         ok ->
             %% offer is pending, just block the stream waiting for the purchase or rejection
             receive
-                {send_purchase, SignedPurchaseSC, Hotspot, PacketHash, Region} ->
+                {send_purchase, PurchaseSC, Hotspot, PacketHash, Region, OwnerSigFun} ->
+                    % TODO: This is async and does not always return the right PacketHash for this offer
```
Vagabond, line 328: Is this a problem, or do we not check it client side?

macpie: Not sure if this is a real problem or not, I just saw it happens few time and made a note

macpie:
> It isn't clear to me if this cache is truly a cache or if it's holding stateful data that if it goes away would cause issues with current channels state
> 
> If it is holding critical state shouldn't that be durably stored somewhere in case the cache is terminated?

The cache does not store valuable information it just helps for lookups, like hotspot -> state channel or list of active state channels

Vagabond (approved):

PR #1124, opened 2021-11-23
Try to use more memory efficent parsing for hexes in snapshots

Vagabond:
term to binary does not preserve sub binary sharing, in testing loading a snap without this the erlang binary heap went from 204758320 to 414104184 bytes when loading the 'hexes' key.

With this patch the binary heap went from 205835720 to 206280848.

This is similar to the patch used to deserialize the list of blocks more efficently.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -620,6 +620,7 @@ load_blocks(Ledger0, Chain, Snapshot) ->
             {ok, Bs} ->
                 lager:info("blocks binary is ~p", [byte_size(Bs)]),
                 print_memory(),
+                %% use a custom decoder here to preserve sub binary references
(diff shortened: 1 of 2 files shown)


Vagabond:
Additonally this patch also avoids some pointless map/list conversions when loading hexes into the ledger.

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -1327,6 +1386,8 @@ deserialize_pairs(<<Bin/binary>>) ->
     ).
 
 -spec deserialize_field(key(), binary()) -> term().
+deserialize_field(hexes, Bin) ->
+    binary_to_proplist(Bin);
```
Vagabond, line 1390: if we're worried about fragility here, we could try/catchj this and fall back to binary to term if it crashes.

madninja (approved):

PR #1125, opened 2021-11-23
Add CLI handler for fast-forwarding a remote peer.

ke6jjj:


ke6jjj:
Despite the CI issue, which appears to be breaking on the vars-nonce-emergency upgrade. I compiled a version of `blockchain-node` that uses this branch and I can confirm that it:

1. Doesn't crash.
2. Appears to work as promised against a working `/p2p/` transport.
3. Appears to work as promised against a working `/ip4/` transport.
4. Returns an error against a botched `/p2p/` transport string.
5. Returns an error against a botched `/ip4/` transport string.
6. Returns a timeout error against a valid `/ip4/` transport string that points to a closed TCP port.

Vagabond (approved):

PR #1126, opened 2021-11-23
Use seed-node libp2p

michaeldjeffrey:


Vagabond:
We cannot and should not merge this until the libp2p seed node PR lands

michaeldjeffrey:
Converting to draft so it cannot be merged

PR #1127, opened 2021-11-23
Test fix for no genesis boot crash

vihu:
Problem
----
When testing locally and supplying `{honor_quick_sync, false}`, miner would not boot at all and crash catastrophically.

Solution
----
If there is no genesis block, just return the bare chain without trying to quick sync.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -165,6 +165,9 @@ new(Dir, undefined, QuickSyncMode, QuickSyncData) ->
             lager:error("DB corrupted cleaning up ~p", [_Corrupted]),
             ok = clean(Blockchain),
             new(Dir, undefined, QuickSyncMode, QuickSyncData);
+        {Blockchain, {error, not_found}} ->
+            lager:info("no genesis block found, returning bare chain"),
+            {no_genesis, Blockchain};
         {Blockchain, {error, _Reason}} ->
             lager:info("no genesis block found: ~p", [_Reason]),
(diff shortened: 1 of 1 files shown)


src/blockchain.erl
```diff
@@ -165,6 +165,9 @@ new(Dir, undefined, QuickSyncMode, QuickSyncData) ->
             lager:error("DB corrupted cleaning up ~p", [_Corrupted]),
             ok = clean(Blockchain),
             new(Dir, undefined, QuickSyncMode, QuickSyncData);
+        {Blockchain, {error, not_found}} ->
+            lager:info("no genesis block found, returning bare chain"),
```
vihu, line 169: I'm not entirely sure whether we need to `mark_upgrade` on the chain ledger here, I'm assuming no, because there is no genesis_block to begin with. Presumably the subsequent `nonce_upgrade` hook is not required to run unless the user has loaded from a really old snapshot?

Vagabond: I think that's fine.

Vagabond:
We do try to init quick sync in the next clause, even though we don't have a genesis block? Maybe we should just have one clause when there's no genesis block, that does what this one does?

vihu:
> We do try to init quick sync in the next clause, even though we don't have a genesis block? Maybe we should just have one clause when there's no genesis block, that does what this one does?

Yeah I can just remove the next clause, no genesis is rarely a condition anyone hits I feel, and yea no point in "quick syncing" if there is no genesis block.

Vagabond:
Can you just merge the clauses and we can merge this.

PR #1128, opened 2021-11-24
Batch signature verification for poc receipts and witnesses

vihu:
Problem
----
Both poc witness and poc receipt signature verification process times can be increase significantly if we do it in batches.

Solution
----
- Add a boolean flag to `is_valid` function for poc_receipt and poc_witness
- Add `verify_signatures` function for poc_receipt and poc_witness
- Invoke batch verification for poc_receipt and poc_witness in poc_receipts_txn

TODO:

- [ ] Resyncing
- [ ] Does this need a chain variable, I suppose we'll find out with resync
- [ ] Check that the new error `invalid_receipt_signatures`  does not break past behavior


src/transactions/v1/blockchain_poc_receipt_v1.erl
```diff
@@ -174,18 +183,30 @@ is_valid(Receipt=#blockchain_poc_receipt_v1_pb{gateway=Gateway, signature=Signat
         false ->
             false;
         true ->
-            PubKey = libp2p_crypto:bin_to_pubkey(Gateway),
-            BaseReceipt = Receipt#blockchain_poc_receipt_v1_pb{signature = <<>>, addr_hash=undefined},
-            EncodedReceipt = blockchain_txn_poc_receipts_v1_pb:encode_msg(BaseReceipt),
-            case libp2p_crypto:verify(EncodedReceipt, Signature, PubKey) of
```
Vagabond, line 180: You could do CheckSignature orelse libp2p_crypto:verify... and avoid code duplication

src/transactions/v1/blockchain_poc_receipt_v1.erl
```diff
@@ -26,10 +26,11 @@
     addr_hash/1, addr_hash/2,
     tx_power/1, tx_power/2,
     sign/2,
-    is_valid/2,
+    is_valid/2, is_valid/3,
     print/1,
     json_type/0,
-    to_json/2
+    to_json/2,
+    verify_signatures/1
```
Vagabond, line 33: Would suggest this returns the batch, so it could be combined with the batch from witnesses?

src/transactions/v1/blockchain_poc_receipt_v1.erl
vihu: Updated in 7b4be74

src/transactions/v1/blockchain_poc_receipt_v1.erl
vihu: Updated in 73bcd21

PR #1135, opened 2021-11-30
Remove unused sim_delay check

Vagabond:


vihu (approved):

vihu:
This was _supposed_ to get used in one of the EQCs to simulate RPi disk lookup, but even that has never been used.

PR #1137, opened 2021-12-01
try to skip another place where the var can be unset

evanmcc:
since ETL goes through this path a different way (it validates all witnesses at the height they're absorbed) we can hit this code at a slightly different angle.  However I could be completely wrong about this, and it might have nothing to do with the clause above.


src/transactions/v1/blockchain_txn_poc_receipts_v1.erl
```diff
@@ -738,12 +738,17 @@ tagged_path_elements_fold(Fun, Acc0, Txn, Ledger, Chain) ->
         {error, request_block_hash_not_found} -> []
     catch
         throw:{error,{region_var_not_set,Region}} ->
+            Path = ?MODULE:path(Txn),
+            lists:foldl(fun({_ElementPos, Element}, Acc) ->
+                                Witnesses = lists:reverse(blockchain_poc_path_element_v1:witnesses(Element)),
+                                Fun(Element, {[{false, list_to_binary(io_lib:format("missing_region_parameters_for_~p", [Region])), Witness} || Witness <- Witnesses], undefined}, Acc)
+                        end, Acc0, lists:zip(lists:seq(1, length(Path)), Path));
+        error:{badmatch, {error, {not_set, Region}}} ->
```
vihu, line 746:
```suggestion
        %% blockchain-etl specific ignore error path to continue absorbing poc receipts
        error:{badmatch, {error, {not_set, Region}}} ->
```

vihu (approved):

Issue #1060, opened 2021-10-25
Support balance clearing / closing payments


abhay:
Spec: A payment_v2 transaction can contain a single payment instruction with a specified `payee`, an empty `amount`, and a `max` boolean set to true (default: false) in order to perform a balance clearing transaction. This transaction would implicit burn the appropriate amount of HNT for the fee and send the remaining balance to the payee.

Some Edge cases:

- **What if multiple payees exist?** Transfer the specified `amount`s to their respective `payee`s and send the remainder to the `payee` with the unspecified `amount`.
- **What if multiple payees exist with unspecified amounts?** Reject the transaction.
- **What about HST?** Defer. Only support HNT clearing transactions.

PR #1138, opened 2021-12-02
Implement balance-clearing txn payments

jeffgrunewald:
Addresses https://github.com/helium/blockchain-core/issues/1060 to allow for adding a single "balance-clearing" txn payment within a txn_payment_v2 that will empty the balance of the payer's account and direct it to the selected payee for whom `max` has been specified as the amount for the payment instead of a defined integer.

Validates that only a single balance-clearing (max) txn is accepted at a given time and implicitly queries the user's balance from the ledger based on the presence of such a payment in a txn payment list to retrieve the available balance and deduct any fees or other specified payments before directing the remainder to the payee in question.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -40,7 +40,7 @@
     {erlang_stats, ".*", {git, "https://github.com/helium/erlang-stats.git", {branch, "master"}}},
     {e2qc, ".*", {git, "https://github.com/helium/e2qc", {branch, "master"}}},
     {vincenty, ".*", {git, "https://github.com/helium/vincenty", {branch, "master"}}},
-    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "master"}}},
+    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "jg/clean_close_payment"}}},
     {merkerl, ".*", {git, "https://github.com/helium/merkerl.git", {branch, "master"}}},
     {xxhash, {git, "https://github.com/pierreis/erlang-xxhash", {branch, "master"}}},
     {exor_filter, ".*", {git, "https://github.com/mpope9/exor_filter", {branch, "master"}}},
--- a/rebar.config
+++ b/rebar.config
@@ -40,7 +40,7 @@
     {erlang_stats, ".*", {git, "https://github.com/helium/erlang-stats.git", {branch, "master"}}},
     {e2qc, ".*", {git, "https://github.com/helium/e2qc", {branch, "master"}}},
     {vincenty, ".*", {git, "https://github.com/helium/vincenty", {branch, "master"}}},
-    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "jg/clean_close_payment"}}},
+    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "master"}}},
     {merkerl, ".*", {git, "https://github.com/helium/merkerl.git", {branch, "master"}}},
     {xxhash, {git, "https://github.com/pierreis/erlang-xxhash", {branch, "master"}}},
     {exor_filter, ".*", {git, "https://github.com/mpope9/exor_filter", {branch, "master"}}},
--- a/src/transactions/v1/blockchain_txn_vars_v1.erl
+++ b/src/transactions/v1/blockchain_txn_vars_v1.erl
@@ -1087,8 +1087,11 @@ validate_var(?poc_v4_randomness_wt, Value) ->
     validate_float(Value, "poc_v4_randomness_wt", 0.0, 1.0);
 validate_var(?poc_v5_target_prob_randomness_wt, Value) ->
     validate_float(Value, "poc_v5_target_prob_randomness_wt", 0.0, 1.0);
-validate_var(?poc_typo_fixes, Value) when is_boolean(Value) -> ok;
-validate_var(?poc_typo_fixes, Value) -> throw({error, {invalidate_poc_typo_fixes, Value}});
+validate_var(?poc_typo_fixes, Value) ->
+    case Value of
+        Val when is_boolean(Val) -> ok;
+        _ -> throw({error, {invalid_poc_typo_fixes, Value}})
+    end;
 validate_var(?poc_target_hex_parent_res, Value) ->
     validate_int(Value, "poc_target_hex_parent_res", 3, 7, false);
 validate_var(?poc_target_hex_collection_res, Value) ->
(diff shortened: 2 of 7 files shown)


rebar.config
```diff
@@ -40,7 +40,7 @@
     {erlang_stats, ".*", {git, "https://github.com/helium/erlang-stats.git", {branch, "master"}}},
     {e2qc, ".*", {git, "https://github.com/helium/e2qc", {branch, "master"}}},
     {vincenty, ".*", {git, "https://github.com/helium/vincenty", {branch, "master"}}},
-    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "master"}}},
+    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "jg/clean_close_payment"}}},
```
vihu, line 43: Plz open the corresponding proto PR, don't see it over there

evanmcc (commented):
this looks good so far, but it does need a chain variable to gate the feature.

evanmcc:
You need to consider two conditions:
1) unknown protobuf fields will be ignored.  I assume that 0 `amount` payments are invalid, but it would be good to check.  Instances running older software won't see the `max` field.
2) on later software this will be known, and they'll be valid.   but this is bad, because they can be partially accepted and hald the group.   So you would add the gate probably right around or before max validation, so we can make sure that enough of the validators can be upgraded before these become valid.

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -370,15 +370,22 @@ memo_check(Txn, Ledger) ->
 
 -spec amount_check(Txn :: txn_payment_v2(), Ledger :: blockchain_ledger_v1:ledger()) -> boolean().
 amount_check(Txn, Ledger) ->
-    TotAmount = ?MODULE:total_amount(Txn, Ledger),
     Payments = ?MODULE:payments(Txn),
-    case blockchain:config(?allow_zero_amount, Ledger) of
-        {ok, false} ->
-            %% check that no more than one payment has a zero amount and only if max is true
-            has_non_zero_amounts(Payments) orelse has_single_max_payment(Payments);
```
evanmcc: I think that you need both of these checks in the enabled clause

test/blockchain_payment_v2_SUITE.erl
```diff
@@ -622,6 +622,8 @@ extra_vars(valid_memo_test) ->
     #{?max_payments => ?MAX_PAYMENTS, ?allow_zero_amount => false, ?allow_payment_v2_memos => true};
 extra_vars(negative_memo_test) ->
     #{?max_payments => ?MAX_PAYMENTS, ?allow_zero_amount => false, ?allow_payment_v2_memos => true};
+extra_vars(balance_clearing_test) ->
+    #{?max_payments => ?MAX_PAYMENTS, ?allow_zero_amount => false, ?enable_balance_clearing => true};
```
evanmcc, line 626: can we add a disabled test to make sure that we're properly rejecting stuff when balance clearing is disabled?

src/transactions/v2/blockchain_payment_v2.erl
```diff
@@ -75,6 +98,11 @@ is_valid_memo(#payment_pb{memo = Memo}) ->
...
     end.
 
+-spec is_valid_max(Payment :: payment()) -> boolean().
+is_valid_max(#payment_pb{amount=Amount, max=false}) when Amount > 0 -> true;
+is_valid_max(#payment_pb{amount=0, max=true}) -> true;
+is_valid_max(_) -> false.
+
 print(undefined) ->
     <<"type=payment undefined">>;
 print(#payment_pb{payee=Payee, amount=Amount, memo=Memo}) ->
```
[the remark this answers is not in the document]
vihu: Also add `max` to `to_json`

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1075,6 +1075,13 @@ validate_var(?allow_zero_amount, Value) ->
         _ -> throw({error, {invalid_allow_zero_amount, Value}})
     end;
 
+validate_var(?enable_balance_clearing, Value) ->
+    case Value of
+        true -> ok;
+        false -> ok;
+        _ -> throw({error, {invalid_enable_balance_clearing, Value}})
+    end;
```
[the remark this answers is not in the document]
jeffgrunewald: validate_var(?enable_balance_clearing, Value) when is_boolean(Value) -> ok;
validate_var(?enable_balance_clearing, Value) -> throw({error, {invalid_enable_balance_clearing, Value}});

src/transactions/v2/blockchain_payment_v2.erl
```diff
@@ -31,24 +33,41 @@
...
+       memo=Memo,
+       max=false
+      };
+new(Payee, max, Memo) ->
+    #payment_pb{
+       payee=Payee,
+       amount=0,
+       memo=Memo,
+       max=true
       }.
```
xandkar, line 68: Doing this with function clauses makes it harder to see at a glance where the differences between the constructions really are. Would be better to highlight where the different values are coming from in a single construction:

```suggestion
new(Payee, Amount0, Memo) ->
    {Amount, IsMax} =
        case Amount0 of
            max -> {0, true};
            N when is_integer(N) andalso N >= 0 -> {N, false}
            %% TODO What is the meaning of IsMax?
            %%      Does it only matter that the txn was constructed with the
            %%      symbolic `max` and not matter if the concrete max of `0`
            %%      was passed?
        end,
    #payment_pb{
        payee  = Payee,
        amount = Amount,
        memo   = Memo,
        max    = IsMax
    }.
```

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -356,102 +456,161 @@ has_default_memos(Payments) ->
 -ifdef(TEST).
 
 new_test() ->
-    Payments = [blockchain_payment_v2:new(<<"x">>, 10),
-                blockchain_payment_v2:new(<<"y">>, 20),
-                blockchain_payment_v2:new(<<"z">>, 30)],
+    Payments = [
+        blockchain_payment_v2:new(<<"x">>, 10),
+        blockchain_payment_v2:new(<<"y">>, 20),
+        blockchain_payment_v2:new(<<"z">>, 30)
```
xandkar, line 462: @Vagabond Not specific to this line or PR, but do we want to continue allowing txn constructions with invalid addresses?

I would argue that all `new`s of txns should be protected by a contract, just like the validation.

Vagabond: yeah we probably should not tolerate it.

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1089,9 +1089,8 @@ validate_var(?poc_v5_target_prob_randomness_wt, Value) ->
     validate_float(Value, "poc_v5_target_prob_randomness_wt", 0.0, 1.0);
 validate_var(?poc_typo_fixes, Value) ->
     case Value of
-        true -> ok;
-        false -> ok;
-        _ -> throw({error, {invalid_poc_typo_fixes, Value}})
+        Val when is_boolean(Val) -> ok;
+        _ -> throw({error, {invalidate_poc_typo_fixes, Value}})
```
evanmcc, line 1093: inadvertent spelling change?

jeffgrunewald: i'd like to claim it was an intentional ironic joke but now i'm just "lucky"

src/transactions/v2/blockchain_txn_payment_v2.erl
madninja: @xandkar presumably this will go elsewhere, so closing for now

evanmcc (approved):

src/transactions/v2/blockchain_txn_payment_v2.erl
xandkar: Yeah, it just caught my eye here at the time, but it is a broader issue to be addressed elsewhere.

PR #1140, opened 2021-12-03
Maybe get new active when overspent

macpie:
- Maybe get a new active state channel when one goes down
- Overspent earlier if we know next packet will be too much
- Test to prove all this


macpie:
Fix for https://github.com/helium/router/issues/532

michaeldjeffrey (approved):

michaeldjeffrey (approved):

Vagabond (approved):

PR #1142, opened 2021-12-03
handle empty block list for testing

evanmcc:
rewrite missed a case.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -626,7 +626,6 @@ load_blocks(Ledger0, Chain, Snapshot) ->
                 []
         end,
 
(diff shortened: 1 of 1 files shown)


Vagabond (approved):

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -711,6 +710,8 @@ load_blocks(Ledger0, Chain, Snapshot) ->
 %% 104 is small tuple
 %% 100 is atom ext
 %% https://www.erlang.org/doc/apps/erts/erl_ext_dist.html
+binary_to_list_of_binaries(<<131, 106>>) ->
```
jeffgrunewald, line 713: probably a naive question: wouldn't this binary be `<<131, 108, 106>>`? if i'm reading this as "ETF byte, start of list byte, end of list byte"

PR #1147, opened 2021-12-07
Macpie/banner signing

macpie:
When banner are sent we need the SC to be signed as the client will verify banner and close connection if banner is unsigned.

For simplicity and speed the worker will not sign on demand (`handle_call(get, ...`) but inside the public function so that every call to `blockchain_state_channels_server:get` will always return a signed SC but (the actual work is done by the caller to not slow down the SC worker).

### DIFF
--- a/src/state_channel/blockchain_state_channels_server.erl
+++ b/src/state_channel/blockchain_state_channels_server.erl
@@ -121,14 +121,17 @@ update_state_channel(SC) ->
     HandlerPid :: pid()
 ) -> ok | reject.
 handle_offer(Offer, SCPacketHandler, Ledger, HandlerPid) ->
-    lager:debug("handle_offer ~p from ~p", [Offer, HandlerPid]),
+    HotspotID = blockchain_state_channel_offer_v1:hotspot(Offer),
+    HotspotName = blockchain_utils:addr2name(HotspotID),
+    lager:debug("handle_offer ~p from ~p", [Offer, {HandlerPid, HotspotName}]),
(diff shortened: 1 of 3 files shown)


src/state_channel/blockchain_state_channels_server.erl
```diff
@@ -121,14 +121,17 @@ update_state_channel(SC) ->
...
-            lager:debug("offer failed to validate ~p ~p", [_Reason, Offer]),
+            lager:debug("offer from ~p failed to validate ~p ~p", [HotspotName, _Reason, Offer]),
             reject;
         true ->
             case SCPacketHandler:handle_offer(Offer, HandlerPid) of
                 {error, _Why} ->
+                    lager:debug("offer from ~p rejected by ~p because ~p ~p", [HotspotName, SCPacketHandler, _Why, Offer]),
                     reject;
                 ok ->
                     handle_offer(Offer, Ledger, HandlerPid)
```
michaeldjeffrey, line 137:
```suggestion
                    handle_offer(Offer, Ledger, HandlerPid)
                    case handle_offer(Offer, Ledger, HandlerPid) of
                        ok ->
                            ok;
                        Ans ->
                            lager:warning(
                              "handle_offer resp mismatch [~p: ok] [state_channel_server: ~p]",
                              [SCPacketHandler, Ans]
                             ),
                            Ans
                    end
```

macpie: `handle_offer(Offer, Ledger, HandlerPid)` has a bunch of lager calls already.

vihu (approved):

PR #1155, opened 2021-12-13
Use pmap to lookup h3 region

vihu:
Switch to using pmap for looking up h3 -> region calls.


vihu:
Handled differently in a performance pass. Closing.

PR #1156, opened 2021-12-13
speed up poc v11 validation

evanmcc:
done:
- coarsen lookups
- add durable cache of gateway region and gain

todo:
- invalidation of gateway regions when region vars change

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -1216,32 +1216,46 @@ load_validators(Gateways, Ledger) ->
                     ledger()) -> ok | {error, _}.
 load_gateways(Gws, Ledger) ->
     AGwsCF = active_gateways_cf(Ledger),
-    GwDenormCF = gw_denorm_cf(Ledger),
     maps:map(
       fun(Address, Gw) ->
               Bin = blockchain_ledger_gateway_v2:serialize(Gw),
-              Location = blockchain_ledger_gateway_v2:location(Gw),
-              Mode = blockchain_ledger_gateway_v2:mode(Gw),
-              Gain = blockchain_ledger_gateway_v2:gain(Gw),
-              Region =
-                  case blockchain_region_v1:h3_to_region(Location, Ledger) of
-                      {ok, Reg} ->
-                          Reg;
-                      _ -> unknown
-                  end,
-              LastChallenge = blockchain_ledger_gateway_v2:last_poc_challenge(Gw),
-              Owner = blockchain_ledger_gateway_v2:owner_address(Gw),
-              cache_put(Ledger, GwDenormCF, <<Address/binary, "-loc">>, term_to_binary(Location)),
-              cache_put(Ledger, GwDenormCF, <<Address/binary, "-last-challenge">>,
-                        term_to_binary(LastChallenge)),
-              cache_put(Ledger, GwDenormCF, <<Address/binary, "-owner">>, Owner),
-              cache_put(Ledger, GwDenormCF, <<Address/binary, "-mode">>, term_to_binary(Mode)),
-              cache_put(Ledger, GwDenormCF, <<Address/binary, "-gain">>, term_to_binary(Gain)),
-              cache_put(Ledger, GwDenormCF, <<Address/binary, "-region">>, term_to_binary(Region)),
+              write_gw_denorm_values(Address, Gw, Ledger, false),
               cache_put(Ledger, AGwsCF, Address, Bin)
       end,
       maps:from_list(Gws)),
     ok.
+
+write_gw_denorm_values(Address, Gw, Ledger, DoRegion) ->
+    GwDenormCF = gw_denorm_cf(Ledger),
+    Location = blockchain_ledger_gateway_v2:location(Gw),
+    Mode = blockchain_ledger_gateway_v2:mode(Gw),
+    Gain = blockchain_ledger_gateway_v2:gain(Gw),
+    case DoRegion of
+        true ->
+            Region =
+                case Location == undefined of
+                    true ->
+                        unknown;
+                    false ->
+                        case blockchain_region_v1:h3_to_region(Location, Ledger) of
+                            {ok, Reg} ->
+                                Reg;
+                            _ -> unknown
+                        end
+                end,
+            cache_put(Ledger, GwDenormCF, <<Address/binary, "-region">>, term_to_binary(Region));
+        false ->
+            ok
+    end,
+
+    LastChallenge = blockchain_ledger_gateway_v2:last_poc_challenge(Gw),
+    Owner = blockchain_ledger_gateway_v2:owner_address(Gw),
+    cache_put(Ledger, GwDenormCF, <<Address/binary, "-loc">>, term_to_binary(Location)),
+    cache_put(Ledger, GwDenormCF, <<Address/binary, "-last-challenge">>,
+              term_to_binary(LastChallenge)),
+    cache_put(Ledger, GwDenormCF, <<Address/binary, "-owner">>, Owner),
+    cache_put(Ledger, GwDenormCF, <<Address/binary, "-mode">>, term_to_binary(Mode)),
+    cache_put(Ledger, GwDenormCF, <<Address/binary, "-gain">>, term_to_binary(Gain)).
 
 -spec entries(ledger()) -> entries().
 entries(Ledger) ->
@@ -1326,6 +1340,42 @@ vars(Vars, Unset, Ledger) ->
               cache_put(Ledger, DefaultCF, var_name(K), term_to_binary(V))
       end,
       Vars),
+    %% we need to invalidate regions in gateways if these potentially change
+    ChangedRegions0 =
+        maps:filter(
+          fun(Var, _V) ->
+                  case atom_to_list(Var) of
+                      "region_" ++ _ -> true;
+                      _ -> false
+                  end
+          end,
+          Vars),
+    ChangedRegions = maps:keys(ChangedRegions0),
+
+    AGwsCF = active_gateways_cf(Ledger),
+    GwDenormCF = gw_denorm_cf(Ledger),
+    cache_fold(
+      Ledger,
+      AGwsCF,
+      fun({GwAddr, _Binary}, _) ->
+              case cache_get(Ledger, GwDenormCF, <<GwAddr/binary, "-region">>, []) of
+                  {ok, BinRegion} ->
+                      case binary_to_term(BinRegion) of
+                          unknown ->
+                              ok;
+                          Region ->
+                              case lists:member(Region, ChangedRegions) of
+                                  true ->
+                                      cache_delete(Ledger, GwDenormCF, <<GwAddr/binary, "-region">>);
+                                  false ->
+                                      ok
+                              end
+                      end;
+                  _ ->
+                      ok
+              end
+      end,
+      ignore),
     lists:foreach(
       fun(K) ->
               cache_delete(Ledger, DefaultCF, var_name(K))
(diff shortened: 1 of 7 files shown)


src/region/blockchain_region_v1.erl
```diff
@@ -35,11 +35,12 @@ get_all_regions(Ledger) ->
 h3_to_region(H3, Ledger) ->
     {ok, VarsNonce} = blockchain_ledger_v1:vars_nonce(Ledger),
     HasAux = blockchain_ledger_v1:has_aux(Ledger),
+    Parent = h3:parent(H3, 8),
```
JayKickliter, line 38: I suggest moving `8` to a macro with a big loud warning that this may break things if we ever generate region polyfills with a pre-compaction resoltion finer than 7

JayKickliter: better yet: if poc11 encodes base resolution, use that

Vagabond: we *should* encode it, but have a default value for now

JayKickliter: Out of band discussion led us to realize we can cache at res 7 due to the fact that region polyfills have a base res of 7 themselves.
```suggestion
    Parent = h3:parent(H3, 7),
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1214,24 +1216,47 @@ load_validators(Gateways, Ledger) ->
...
-              cache_put(Ledger, GwDenormCF, <<Address/binary, "-owner">>, Owner),
-              cache_put(Ledger, GwDenormCF, <<Address/binary, "-mode">>, term_to_binary(Mode)),
+              write_gw_denorm_values(Address, Gw, Ledger, false),
               cache_put(Ledger, AGwsCF, Address, Bin)
       end,
       maps:from_list(Gws)),
     ok.
 
+write_gw_denorm_values(Address, Gw, Ledger, DoRegion) ->
+    GwDenormCF = gw_denorm_cf(Ledger),
```
jeffgrunewald, line 1229: would it not be safe to make this call only once before the `maps:map/2` call and pass it in as an additional parameter to `write_gw_denorm_values/4-5`?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1214,24 +1216,47 @@ load_validators(Gateways, Ledger) ->
...
+    end,
+
+    LastChallenge = blockchain_ledger_gateway_v2:last_poc_challenge(Gw),
+    Owner = blockchain_ledger_gateway_v2:owner_address(Gw),
+    cache_put(Ledger, GwDenormCF, <<Address/binary, "-loc">>, term_to_binary(Location)),
+    cache_put(Ledger, GwDenormCF, <<Address/binary, "-last-challenge">>,
+              term_to_binary(LastChallenge)),
+    cache_put(Ledger, GwDenormCF, <<Address/binary, "-owner">>, Owner),
+    cache_put(Ledger, GwDenormCF, <<Address/binary, "-mode">>, term_to_binary(Mode)),
+    cache_put(Ledger, GwDenormCF, <<Address/binary, "-gain">>, term_to_binary(Gain)).
```
jeffgrunewald, line 1258: any boosts to be gained from putting these separate writes into a batch? like wherever we've got 4+ puts in quick succession?

Vagabond: these go into ETS until the cache gets comitted

jeffgrunewald: Where/when are we setting the ledger context cache to `direct` this this would not be the case? Just in tests?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -970,6 +970,8 @@ validate_var(?check_snr, Value) ->
         false -> ok;
         _ -> throw({error, {invalid_check_snr, Value}})
     end;
+validate_var(?polyfill_resolution, Value) ->
+    validate_int(Value, "polyfill_resolution", 0, 1000, false);
```
Vagabond, line 974: H3 hexes are only 0-15

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1214,24 +1216,87 @@ load_validators(Gateways, Ledger) ->
...
+            %%                         {ok, Reg} -> Reg;
+            %%                         _ -> unknown
+            %%                     end
+            %%             end,
+            %%         maybe_update_field(<<Address/binary, "-region">>, term_to_binary(Region), GwDenormCF, Ledger);
+            %%     false -> ok
+            %% end;
+            ok;
+        false -> ok
+    end,
```
jeffgrunewald, line 1277: this whole case seems to be unnecessary now

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1214,24 +1216,87 @@ load_validators(Gateways, Ledger) ->
...
+%% this requires the values to be pre-encoded, and returns a boolean that is true if the value was written
+maybe_update_field(Field, Value, CF, Ledger) ->
+    case cache_get(Ledger, CF, Field, []) of
+        {ok, Value} ->
+            false;
+        %% this catchall covers if the value was different or missing
+        _ ->
+            cache_put(Ledger, CF, Field, Value),
+            true
+    end.
```
jeffgrunewald, line 1298: could you add a `cache_put_new/4` for these that uses `ets:insert_new/2` and then avoid the case/get pattern here and just always attempt the write? for that matter, could you then replace `insert/2` altogether with `insert_new/2` and then combine the `write_gw_denorm_values(_, _, new | update, _, _)` into a single function?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1315,6 +1380,42 @@ vars(Vars, Unset, Ledger) ->
...
       end,
       Vars),
+    %% we need to invalidate regions in gateways if these potentially change
+    ChangedRegions0 =
+        maps:filter(
+          fun(Var, _V) ->
+                  case Var of
+                      <<"region_", _/binary>> -> true;
+                      _ -> false
+                  end
```
jeffgrunewald, line 1390:
```suggestion
          fun(<<"region_", _/binary>>, _V) -> true;
               (_, _) -> false
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1315,6 +1380,42 @@ vars(Vars, Unset, Ledger) ->
...
+          end,
+          Vars),
+    ChangedRegions = maps:keys(ChangedRegions0),
```
jeffgrunewald, line 1393:
```suggestion
    ChangedRegions =
        maps:keys(
            maps:filter(
                fun(<<"region_", _/binary>>, _V) -> true;
                     (_, _) -> false
                end,
                Vars)),
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1458,6 +1560,66 @@ find_gateway_mode(Address, Ledger) ->
...
+                end,
+            case Res of
+                {ok, Region} ->
+                    {ok, Region};
+                _ ->
+                    {error, {unknown_region, Location}}
+            end;
+        Error ->
+            Error
+    %% end
```
jeffgrunewald, line 1619:
```suggestion
```

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1214,24 +1216,86 @@ load_validators(Gateways, Ledger) ->
...
-              cache_put(Ledger, GwDenormCF, <<Address/binary, "-mode">>, term_to_binary(Mode)),
+              write_gw_denorm_values(Address, Gw, new, Ledger, false),
               cache_put(Ledger, AGwsCF, Address, Bin)
       end,
       maps:from_list(Gws)),
     ok.
 
+write_gw_denorm_values(Address, Gw, new, Ledger, DoRegion) ->
+    GwDenormCF = gw_denorm_cf(Ledger),
+    Location = blockchain_ledger_gateway_v2:location(Gw),
```
Vagabond, line 1230: Hoist this behind the DoRegion check?

src/ledger/v1/blockchain_ledger_v1.erl
Vagabond: there's a readthrough to rocksdb on a cache miss here, so it's a bit more complex

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1214,24 +1216,86 @@ load_validators(Gateways, Ledger) ->
...
+    maybe_update_field(<<Address/binary, "-gain">>, term_to_binary(Gain), GwDenormCF, Ledger),
+    LastChallenge = blockchain_ledger_gateway_v2:last_poc_challenge(Gw),
+    maybe_update_field(<<Address/binary, "-last-challenge">>, term_to_binary(LastChallenge),
+                       GwDenormCF, Ledger),
+    Owner = blockchain_ledger_gateway_v2:owner_address(Gw),
+    maybe_update_field(<<Address/binary, "-owner">>, Owner, GwDenormCF, Ledger),
+    ok.
+
+%% this requires the values to be pre-encoded, and returns a boolean that is true if the value was written
+maybe_update_field(Field, Value, CF, Ledger) ->
```
Vagabond, line 1289: Given this potentially requires a bunch of ETS cache misses and then rocksdb cache misses before we do a write, is this really saving us any work?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1335,36 +1435,16 @@ config(ConfigName, Ledger) ->
     end.
 
 vars_nonce(Ledger) ->
-    case ?MODULE:get_context(Ledger) of
```
Vagabond, line 1338: Removing this means we have to pull the vars nonce a bunch now (once per chain var lookup), no?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1315,6 +1379,42 @@ vars(Vars, Unset, Ledger) ->
               cache_put(Ledger, DefaultCF, var_name(K), term_to_binary(V))
       end,
       Vars),
+    %% we need to invalidate regions in gateways if these potentially change
```
Vagabond, line 1382: You may also need to invalidate if poc version changes, or if we change the *order* of the regions?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1458,6 +1538,65 @@ find_gateway_mode(Address, Ledger) ->
...
+            end;
+        _ ->
+            case find_gateway_location(Address, Ledger) of
+                {ok, Location} ->
+                    Res =
+                        case RegionBins of
+                            no_prefetch ->
+                                blockchain_region_v1:h3_to_region(Location, Ledger);
+                            unset ->
+                        {error, {unknown_region, Location}};
```
Vagabond, line 1585: Won't this get conflated with having no region, vs having no way to resolve the region?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1214,24 +1216,86 @@ load_validators(Gateways, Ledger) ->
...
+        true ->
+            case DoRegion of
+                true ->
+                    Region =
+                        case Location == undefined of
+                            true -> unknown;
+                            false ->
+                                case blockchain_region_v1:h3_to_region(Location, Ledger) of
+                                    {ok, Reg} -> Reg;
+                                    _ -> unknown
```
Vagabond, line 1269: I think we have to be careful when we write 'unknown' region here so that we only write it when there's truly no region that matches, vs cases when not all the vars are set, or the vars are invalid (like when we had whitesspace in the list of regions), etc. If we write unknown too aggressively we essentially poison the region cache for this hotspot

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -1315,6 +1379,42 @@ vars(Vars, Unset, Ledger) ->
...
+      Ledger,
+      AGwsCF,
+      fun({GwAddr, _Binary}, _) ->
+              case cache_get(Ledger, GwDenormCF, <<GwAddr/binary, "-region">>, []) of
+                  {ok, BinRegion} ->
+                      case binary_to_term(BinRegion) of
+                          unknown ->
+                              ok;
+                          Region ->
+                              case lists:member(Region, ChangedRegions) of
```
Vagabond, line 1406: gateways with an unknown region also need to be recalculated, as the regions may have grown?

evanmcc:
closing in favor of #1163

PR #1160, opened 2021-12-15
If the chain is missing the genesis block due to corruption repair it

Vagabond:
On startup, if we cannot fetch the genesis block, but the chain is not
blank, clean the chain and then load the genesis block again. This
resolves a bootup crash loop.


vihu (approved):
lgtm

PR #1173, opened 2021-12-22
Garbage collect inactive hotspots from the h3dex

Vagabond:
Problem to solve: we have 2 problems with targeting. First that it uses the `hexes` which are giant monolithic datastructures that are slow to update and are scaling poorly across 400k+ hotspots. Secondly both the hexes and the h3dex ledger data does not take into account inactive hotspots when giving you a raw count of hotspots in an area.

This PR starts work on garbage collecting inactive hotspots from the h3dex. Additional work is needed to rewrite targeting to *use* the h3index, maybe @andymck has some work he's done here we can reuse. Targeting using the h3dex is a bit more complicated to do performantly than I'd expected and I think we need to explore the idea of dropping the requirement that targeting evaluates every possible res 5 hex (of which we have 14,000 populated out of a possible 2 million). Something that picked a random subset of populated res 5 hexes and then targeted across those might be better?

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -170,7 +170,7 @@
 %% determines which hexing type to use for gateways: hex_h3dex | h3dex or not set
 %% hex_h3dex will result in both hexes and h3dex being updated
 %% h3dex will result in only h3dex being updated
-%% not set will result in hexes being updated
+%% not set will result in both hexes and h3dex being updated
 -define(poc_hexing_type, poc_hexing_type).
 
 %% max number of hexes to GC in the h3dex per block: integer
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -166,6 +166,22 @@
 
 %% resolution for h3 polyfills, defaulted to 7 before we set it
 -define(polyfill_resolution, polyfill_resolution).
+
+%% determines which hexing type to use for gateways: hex_h3dex | h3dex or not set
+%% hex_h3dex will result in both hexes and h3dex being updated
+%% h3dex will result in only h3dex being updated
+%% not set will result in hexes being updated
+-define(poc_hexing_type, poc_hexing_type).
+
+%% max number of hexes to GC in the h3dex per block: integer
+-define(h3dex_gc_width, h3dex_gc_width).
+
+%% the version number of poc targeting in use: integer
+%% if not set, code paths with default to 3 ( blockchain_poc_target_v3 )
+-define(poc_targeting_version, poc_targeting_version).
+
+%% the number of random hexes to utilize when targeting: integer
+-define(poc_target_pool_size, poc_target_pool_size).
 %%%
 %%% score vars
 %%%
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -4416,7 +4416,7 @@ add_gw_to_h3dex(Hex, GWAddr, Res, Ledger) ->
     BinHex = h3_to_key(Hex),
     case cache_get(Ledger, H3CF, BinHex, []) of
         not_found ->
-            case count_gateways_in_hex(h3:parent(Hex, 5), Ledger) of
+            case count_gateways_in_hex(h3:parent(Hex, Res), Ledger) of
                 0 ->
                     %% populating a hex means we need to recalculate the set of populated
                     %% hexes
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -77,6 +77,7 @@
     update_gateway_score/3, gateway_score/2,
     update_gateway_oui/4,
     gateway_count/1,
+    gateway_update_challenge/5,
 
     find_pocs/2,
     find_poc/3,
@@ -84,6 +85,7 @@
     delete_poc/3,
     maybe_gc_pocs/2,
     maybe_gc_scs/2,
+    maybe_gc_h3dex/1,
 
     upgrade_pocs/1,
 
@@ -155,16 +157,21 @@
     set_hexes/2, get_hexes/1, get_hexes_list/1,
     set_hex/3, get_hex/2, delete_hex/2,
 
-    add_to_hex/3,
-    remove_from_hex/3,
+    add_to_hex/4,
+    remove_from_hex/4,
 
     clean_all_hexes/1,
 
     bootstrap_h3dex/1,
     get_h3dex/1, delete_h3dex/1,
     lookup_gateways_from_hex/2,
-    add_gw_to_hex/3,
-    remove_gw_from_hex/3,
+    add_gw_to_h3dex/4,
+    remove_gw_from_h3dex/4,
+    count_gateways_in_hex/2,
+    count_gateways_in_hexes/2,
+    random_targeting_hex/2,
+    build_random_hex_targeting_lookup/2,
+    clean_random_hex_targeting_lookup/1,
     add_commit_hook/4, add_commit_hook/5,
     remove_commit_hook/2,
 
@@ -1573,7 +1580,7 @@ add_gateway(OwnerAddr,
                     {ok, V} when V > 6 ->
                         {ok, Res} = blockchain:config(?poc_target_hex_parent_res, Ledger),
                         Hex = h3:parent(Location, Res),
-                        add_to_hex(Hex, GatewayAddress, Ledger),
+                        add_to_hex(Hex, GatewayAddress, Res, Ledger),
                         NewGw0;
                     {ok, V} when V > 3 ->
                         Gateways = active_gateways(Ledger),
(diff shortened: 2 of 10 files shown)


Vagabond:
Stuff outstanding:

- [ ] We should review/merge the `var hooks` PR and use it to re-calculate the h3dex if the targeting resolution changes
- [ ] This needs to be tested to see if it actually works as expected
- [ ] We probably need more chain vars to enable the switch over
- [ ] We need a strategy to drop/stop updating the old `hexes` data once the switch over is complete and deemed stable because it's very expensive

Vagabond:
```2022-01-20 16:36:54.262 1 [info] <0.349.1>@blockchain_utils:maybe_log_duration:681 updated_hexes took 306605 usec
2022-01-20 16:36:54.270 1 [info] <0.349.1>@blockchain_utils:maybe_log_duration:681 updated_h3dex took 7880 usec
2022-01-20 16:36:54.270 1 [info] <0.349.1>@blockchain_txn:absorb:{562,21} took 316 ms to absorb blockchain_txn_assert_location_v2
```

As you can see, almost all of the absorb time in assert_location is spent updating the `hexes` list, and h3dex is much, much cheaper.

Vagabond:
I think we could also GC when computing hip17 information?

evanmcc (commented):
I like this approach, (and I still need to go over andy's PR again), but I have concerns about how often we rework the targeting index.  I think that GCing every block is probably OK, but we get several asserts or adds per block now and this approach seems slower than the existing approach?   the timings from chat seems to be on the order of a second per recalculation, vs ~40ms for the existing approach.  given the number of receipts, maybe it nets out better, but I wonder if we couldn't move it to a periodic recalculation at the cost of some spots not being able to be immediately challenged.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -4210,10 +4353,23 @@ add_gw_to_hex(Hex, GWAddr, Ledger) ->
     BinHex = h3_to_key(Hex),
     case cache_get(Ledger, H3CF, BinHex, []) of
         not_found ->
+            case count_gateways_in_hex(h3:parent(Hex, 5), Ledger) of
```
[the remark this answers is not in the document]
Vagabond: That should be the targeting res var value

Vagabond:
> I like this approach, (and I still need to go over andy's PR again), but I have concerns about how often we rework the targeting index. I think that GCing every block is probably OK, but we get several asserts or adds per block now and this approach seems slower than the existing approach? the timings from chat seems to be on the order of a second per recalculation, vs ~40ms for the existing approach. given the number of receipts, maybe it nets out better, but I wonder if we couldn't move it to a periodic recalculation at the cost of some spots not being able to be immediately challenged.

As noted in discord, the recalculation only happens when a hex goes from populated to unpopulated, or vice versa, so at res5 this is pretty rare.

include/blockchain_vars.hrl
```diff
@@ -166,6 +166,22 @@
 
 %% resolution for h3 polyfills, defaulted to 7 before we set it
 -define(polyfill_resolution, polyfill_resolution).
+
+%% determines which hexing type to use for gateways: hex_h3dex | h3dex or not set
+%% hex_h3dex will result in both hexes and h3dex being updated
+%% h3dex will result in only h3dex being updated
+%% not set will result in hexes being updated
```
Vagabond, line 173: not being set is the same as hex_h3dex?

Vagabond:
Notes on aux ledger testing:

* Add an aux ledger dir to the blockchain stanza of your config: `{aux_ledger_dir, "aux"}`
* Start the node (this might take a moment to bootstrap the aux ledger)
*  `miner repair sync_pause`
* Check you have a working aux ledger: `miner eval 'blockchain_ledger_v1:has_aux(blockchain:ledger()).`
* `blockchain_aux_ledger_v1:set_vars(#{poc_target_pool_size => 100, poc_targeting_version => 4}, blockchain_ledger_v1:mode(aux, blockchain:ledger())).`
* `TempLedger = blockchain_ledger_v1:new_context(blockchain_ledger_v1:mode(aux, blockchain:ledger())), blockchain_ledger_v1:build_random_hex_targeting_lookup(5, TempLedger), blockchain_ledger_v1:commit_context(TempLedger).`
*
```Vars = [poc_version,poc_v4_target_challenge_age,                                                                                                                                                                                                                                 poc_v5_target_prob_randomness_wt,
 poc_witness_consideration_limit,poc_v4_prob_rssi_wt,
 poc_v4_prob_time_wt,poc_v4_prob_count_wt,
 poc_v4_prob_no_rssi,poc_v4_prob_good_rssi,
 poc_v4_prob_bad_rssi,poc_v4_parent_res,
 poc_v4_exclusion_cells,poc_v4_randomness_wt,
 poc_centrality_wt,poc_good_bucket_low,poc_good_bucket_high,
 poc_max_hop_cells,poc_v4_target_prob_score_wt,
 poc_v4_target_prob_edge_wt,poc_v5_target_prob_randomness_wt,
 poc_v4_target_score_curve,poc_v4_target_exclusion_cells,
 poc_path_limit]
```

Now you can run a targeting run, comparing both modules with the same pubkey and the same random seed:

```
f(PubKey), f(PubKeyBin), #{public:=PubKey} = libp2p_crypto:generate_keys(ecc_compact), PubKeyBin = libp2p_crypto:pubkey_to_bin(PubKey), Ledger = blockchain:ledger(), AuxLedger= blockchain_ledger_v1:mode(aux, blockchain:ledger()), [ begin Hash = crypto:strong_rand_bytes(32), {element(1, element(2, blockchain_poc_target_v3:target(PubKeyBin, Hash, Ledger, blockchain_utils:get_vars(Vars, Ledger)))), element(1, element(2, blockchain_poc_target_v4:target(PubKeyBin, Hash, AuxLedger, blockchain_utils:get_vars(Vars, AuxLedger))))} end || _ <- lists:seq(1, 100)].```

This outputs a list of 2-tuples which are the pubkey of the targeted hotspot the first element is the current v3 targeting, the second is the v4 targeting against the aux ledger.

**Comment by Vagabond** [comment]
[target-50.gz](https://github.com/helium/blockchain-core/files/8027681/target-50.gz)
[target-100.gz](https://github.com/helium/blockchain-core/files/8027860/target-100.gz)
[target-10.gz](https://github.com/helium/blockchain-core/files/8027903/target-10.gz)

File: include/blockchain_vars.hrl (hunk 2 above)
Comment (andymck) [reply include/blockchain_vars.hrl:173]: yes, think the comment was meant to read "not set will result in BOTH hexes being updated"

Comment (andymck) [inline include/blockchain_vars.hrl:173]:
```suggestion
%% not set will result in both hexes being updated
```

Vagabond: `both hexes and h3dex`?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -687,10 +687,40 @@ process_hooks(Vars, Unsets, Ledger) ->
...
+    %% rebuild hexes since we're back to updating them
+    blockchain:bootstrap_hexes(Ledger),
+    ok;
 var_hook(_Var, _Value, _Ledger) ->
     ok.
 
 -spec unset_hook(Var :: atom(), Ledger :: blockchain_ledger_v1:ledger()) -> ok.
+unset_hook(?poc_targeting_version, Ledger) ->
+    %% going back to the default, which is v3 so remove the h3dex lookup
+    blockchain_ledger_v1:clean_random_hex_targeting_lookup(Ledger),
```
evanmcc, line 718: do we need a boostrap hex here too?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -687,10 +687,40 @@ process_hooks(Vars, Unsets, Ledger) ->
 %% Separate out hook functions and call them in separate functions below the hook section.
 
 -spec var_hook(Var :: atom(), Value :: any(), Ledger :: blockchain_ledger_v1:ledger()) -> ok.
+var_hook(?poc_targeting_version, 4, Ledger) ->
+    %% v4 targeting enabled, build the h3dex lookup
+    {ok, Res} = blockchain_ledger_v1:config(?poc_target_hex_parent_res, Ledger),
+    blockchain_ledger_v1:build_random_hex_targeting_lookup(Res, Ledger),
+    ok;
```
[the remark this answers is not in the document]
Vagabond: there's a separate var that disables updating hexes that has a hook that deletes the hexes list

src/transactions/v1/blockchain_txn_vars_v1.erl
Vagabond: In theory we'd enable targeting v4 and leave the hexes being updated for a bit before disabling hexes updating

evanmcc (approved):

evanmcc:
I think that this is ready to go once we've got approval and some indications that fairness the same or better than the old code.

Vagabond:
Sampling to a file:

```
f(PubKey), f(PubKeyBin), f(F), {ok, F} = file:open("/tmp/target-100", [raw,  write]), #{public:=PubKey} = libp2p_crypto:generate_keys(ecc_compact), PubKeyBin = libp2p_crypto:pubkey_to_bin(PubKey), Ledger = blockchain:ledger(), AuxLedger= blockchain_ledger_v1:mode(aux, blockchain:ledger()), [ begin Hash = crypto:strong_rand_bytes(32), file:write(F, io_lib:format("~s,~s~n", [libp2p_crypto:bin_to_b58(element(1, element(2, blockchain_poc_target_v3:target(PubKeyBin, Hash, Ledger, blockchain_utils:get_vars(Vars, Ledger))))), libp2p_crypto:bin_to_b58(element(1, element(2, blockchain_poc_target_v4:target(PubKeyBin, Hash, AuxLedger, blockchain_utils:get_vars(Vars, AuxLedger)))))])) end || _ <- lists:seq(1, 750000)], file:close(F).
```

PR #1175, opened 2021-12-28
Allow snapshots to be loaded off disk rather than read into ram first

Vagabond:
This change should help devices with less memory load larger snapshots.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -1479,7 +1479,7 @@ deserialize_pairs(<<Bin/binary>>) ->
     ).
 
 -spec deserialize_field(key(), binary()) -> term().
-deserialize_field(hexes, Bin) ->
+deserialize_field(hexes, Bin) when is_binary(Bin) ->
     %% hexes are encoded as a term_to_binary of a big
     %% key/value list of {h3() -> [binary()]},
     %% and a single 'list' key which points to a map of
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -14,7 +14,7 @@
 
          snapshot/3,
          snapshot/4,
-         import/3,
+         import/5,
          load_into_ledger/3,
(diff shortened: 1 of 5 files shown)


src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -28,11 +28,11 @@
          diff/2
         ]).
 
--ifdef(TEST).
+%-ifdef(TEST).
```
jadeallenx, line 31: If this should be exported, just export it?

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -610,18 +623,28 @@ load_blocks(Ledger0, Chain, Snapshot) ->
     print_memory(),
```
jadeallenx, line 623: Hopefully this print function is conditional or routed through lager log levels.

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
Vagabond: This was a side effect of my shell testing, it does not need to be exported.

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -1444,6 +1552,19 @@ mk_bin_iterator(<<SizK:32/little-unsigned-integer, K:SizK/binary,
...
 
+mk_file_iterator(_FD, End, End) ->
+    fun() -> ok end;
+mk_file_iterator(FD, Pos, End) when Pos < End ->
+    fun() ->
+            {ok, Pos} = file:position(FD, {bof, Pos}),
+            {ok, <<SizK:32/integer-unsigned-little>>} = file:read(FD, 4),
+            {ok, <<K:SizK/binary, SizV:32/integer-unsigned-little>>} = file:read(FD, SizK + 4),
+            {ok, V} = file:read(FD, SizV),
+            io:format("read key of size ~p and value of size ~p", [SizK, SizV]),
```
evanmcc, line 1563: remove

evanmcc (approved):

PR #1180, opened 2022-01-04
Add support for sc_merge_version

vihu:
Problem
----
Existing way of merging state channels is slow and we haven't been able to faithfully reproduce sync crash when using the new style of state channel merge.

Solution
----
Move to new style state channel merge guarded behind a new chain variable `sc_merge_version`; the only allowed value for this var would be `1` (and would have to be manually incremented and validated in the future if we have to update it).

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -341,6 +341,8 @@
 %% The "new" behavior (i.e., `true') is to count state channels
 %% that are only in the open state and ignore closed channels.
 -define(sc_only_count_open_active, sc_only_count_open_active).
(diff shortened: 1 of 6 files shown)


Vagabond:
I think @macpie is working on an alternative that would mean we don't need to merge SCs that will hopefully render this PR redundant.

vihu:
> I think @macpie is working on an alternative that would mean we don't need to merge SCs that will hopefully render this PR redundant.

Roger, we can close this when there's an alternate approach

Vagabond:
Closing because its redundant

PR #1191, opened 2022-01-10
Enable LOAD_SNAPSHOT os env var

jadeallenx:
Problem to solve: It would be convenient for automation to temporarily set an OS environment variable `LOAD_SNAPSHOT` to a false value to prevent a snapshot from loading at start up.

Solution: Add a function to test for an os var and have it return `true` or `false` - defaults to `true` if the variable is unset or set to a value other than "0", "no" or "false"

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -839,23 +839,6 @@ snapshot_sync(State) ->
     lager:info("snapshot_sync starting ~p ~p", [Pid, Ref]),
     State#state{sync_pid = Pid, sync_ref = Ref, mode = snapshot, snapshot_info=SnapInfo}.
 
-%% @doc If this os environment variable is set and is a "truthy" value,
-%% return `true'; otherwise, `false'. Unset variables return the default value.
-%%
-%% False values are "0", "no", "false".  True values are any value that aren't
-%% listed as explicitly false.
-get_boolean_os_var_or_default(VarName, Default) ->
-    case os:getenv(VarName, undefined) of
-        "0" -> false;
-        "no" -> false;
-        "NO" -> false;
-        "FALSE" -> false;
-        "false" -> false;
-        undefined -> Default;
-        _ -> true
-    end.
-
-
 reset_ledger_to_snap(Hash, Height, State) ->
     lager:info("clearing the ledger now"),
     State1 = pause_sync(State),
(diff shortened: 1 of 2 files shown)


src/blockchain_worker.erl
```diff
@@ -830,18 +830,42 @@ maybe_sync_blocks(#state{blockchain = Chain} = State) ->
...
+            lager:debug("Not loading snapshot because 'LOAD_SNAPSHOT' is set false"),
+            State
     end.
 
+%% @doc If this os environment variable is set and is a "truthy" value,
+%% return `true'; otherwise, `false'. Unset variables return the default value.
+%%
+%% False values are "0", "no", "false".  True values are any value that aren't
+%% listed as explicitly false.
+get_boolean_os_var_or_default(VarName, Default) ->
```
jeffgrunewald, line 857: this seems like a more generic `blockchain_util` function

jadeallenx: Yeah it probably should be

madninja:
So how would one actually _use_ this? Set the env var,  then use `snapshot load` to load a snapshot, which loads the given snapshot and then starts the world normally? Or do I need to restart?

jadeallenx:
Yeah set the OS VAR, load the snapshot using CLI, then unset the OS VAR

jadeallenx:
Also:

1. Turn off miner
2. Delete all the DB files
3. Set OS VAR
4. Turn on miner
5. Load the snapshot
6. unset OS VAR

madninja:
> Also:
> 
> 1. Turn off miner
> 2. Delete all the DB files
> 3. Set OS VAR
> 4. Turn on miner
> 5. Load the snapshot
> 6. unset OS VAR


1. Wouldn't the OS VAR already be set when you turn miner back on? turning it of would have no effect unless you restart it? Didn't think the environment was dynamic like that
2. How does this interact with systemd stop/start which we'd do in step 4 and 5?

jadeallenx:
> 1. Wouldn't the OS VAR already be set when you turn miner back on? turning it of would have no effect unless you restart it? Didn't think the environment was dynamic like that

I'm confused by this use case. The behavior is that snapshot loading at start up (and during periodic checks) is prevented only when the OS VAR exists, and is set to a "false" value.  If the var exists and set to a true value, snapshot loading is enabled. If the var doesn't exist, snapshot loading is enabled.

> 2. How does this interact with systemd stop/start which we'd do in step 4 and 5?

The environment variable state gets propagated through systemd

madninja:
> > 1. Wouldn't the OS VAR already be set when you turn miner back on? turning it of would have no effect unless you restart it? Didn't think the environment was dynamic like that
> 
> I'm confused by this use case. The behavior is that snapshot loading at start up (and during periodic checks) is prevented only when the OS VAR exists, and is set to a "false" value. If the var exists and set to a true value, snapshot loading is enabled. If the var doesn't exist, snapshot loading is enabled.
> 
> > 2. How does this interact with systemd stop/start which we'd do in step 4 and 5?
> 
> The environment variable state gets propagated through systemd

Gotcha, ok so this can be used to automate upgrades. I like it

madninja (approved):

PR #1192, opened 2022-01-11
Remove state channels that have expired, in case we miss the close

Vagabond:


src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -325,6 +325,13 @@ handle_info({blockchain_event, {add_block, BlockHash, false, Ledger}},
                                                    end,
                                                    [SCID|Acc];
                                                false ->
+                                                   case LedgerHeight > ExpireAt + SCGrace of
+                                                       true ->
+                                                           %% it's dead, remove it
```
jeffgrunewald, line 330: Log statement to highlight where/how often we're missing closes?

jeffgrunewald: To eventually be followed/replaced by some or telemetry

Vagabond: this will all go away with the gateway-rs integration soon?

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -228,6 +228,25 @@ handle_info({dial_success, OUIOrAddress, Stream}, State0) ->
         _ ->
             {noreply, maybe_send_packets(OUIOrAddress, Stream, State1)}
     end;
+handle_info({blockchain_event, {add_block, _BlockHash, true, Ledger}}, State) ->
```
vihu, line 231: Should this only trigger when the client got the add_block via sync? The `true` flag is for syncing here. I think this should trigger regardless of how the block came in (sync or otherwise) perhaps?

Vagabond: the other clause handles the other kinds of state channels, originally this was on the same clause, but I think the issue is that hotspots doing a lot of syncing never can GC their SCs

jadeallenx: This won't cause issues with rewards on validators?

Vagabond: no, this is the local copy of a state channel's client version of a state channel, it is not on-chain data

vihu (approved):

Issue #1130, opened 2021-11-29
Investigate the why the vars nonce was wrong on part of the fleet.


evanmcc:
Everything that loaded the rescue block manually seems to have had the wrong vars nonce, but gateways and other nodes that loaded a snap containing the rescue block halted because their nonce was wrong.  Probably just not running all the hooks correctly post block save.

xandkar:
@evanmcc Until full investigation of what exactly goes wrong in the presence of rescue blocks, we discussed a workaround with @Vagabond - just avoid making snaps with rescue blocks: https://github.com/helium/blockchain-core/pull/1195

PR #1195, opened 2022-01-11
Skip snap-making attempts when some blocks are rescue blocks

xandkar:
This should step around the var nonce problem (#1130), until its mystery is solved. As discussed with @Vagabond.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -846,10 +846,11 @@ upgrade_block_info(#block_info{hash = Hash, height = Height}, Block, Chain = #bl
     deserialize_block_info(InfoBin, Chain).
 
 %% @doc read blocks from the db without deserializing them
--spec get_raw_block(blockchain_block:hash() | integer(), blockchain()) -> {ok, binary()} | not_found | {error, any()}.
-get_raw_block(Hash, #blockchain{db=DB, blocks=BlocksCF}) when is_binary(Hash) ->
+-spec get_raw_block(blockchain_block:hash() | integer(), blockchain()) ->
+    {ok, binary()} | not_found | {error, any()}.
+get_raw_block(<<Hash/binary>>, #blockchain{db=DB, blocks=BlocksCF}) ->
     rocksdb:get(DB, BlocksCF, Hash, []);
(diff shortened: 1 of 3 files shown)


src/blockchain.erl
```diff
@@ -1205,29 +1210,38 @@ process_snapshot(ConsensusHash, MyAddress, Signers,
                         lager:info("skipping previously failed snapshot at height ~p", [Height]);
                     _ ->
                         Blocks = blockchain_ledger_snapshot_v1:get_blocks(Blockchain),
```
Vagabond, line 1212: Would it not be better to do this check in `blockchain_ledger_snapshot_v1:get_blocks`?

xandkar: @Vagabond That was my first thought as well, but here is just one of the uses of `blockchain_ledger_snapshot_v1:get_blocks`, the other is from CLI. I figured that it would be better to leave the manual call an option to do whatever it wants, while restricting the automated call (which is here).

Vagabond: The problem is that a manual snap is what got us into trouble last time.

xandkar: @Vagabond I switched it to check within `blockchain_ledger_snapshot_v1:get_blocks`.

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -849,11 +853,14 @@ get_blocks(Chain) ->
     %% whichever is lower.
     LoadBlockStart = min(DHeight, ElectionHeight - GraceBlocks),
 
-    [begin
-         {ok, B} = blockchain:get_raw_block(N, Chain),
-         B
-     end
-     || N <- lists:seq(max(?min_height, LoadBlockStart), Height)].
+    HeightRange = lists:seq(max(?min_height, LoadBlockStart), Height),
+    Blocks = [get_ok(blockchain:get_raw_block(H, Chain)) || H <- HeightRange],
```
Vagabond, line 857: Would be nice to throw() as soon as you encounter a rescue block vs doing a double traverse

Vagabond (approved):
Approved pending review comment

PR #1198, opened 2022-01-13
memoize witness validation

evanmcc:
When we make rewards, we assess witness validity twice for each receipt transaction.  The results are the same each time, so we might as well memoize it locally to make it cheaper.  It would be more complex to thread the results through, so we instead use the pdict.

### DIFF
--- a/src/transactions/v2/blockchain_txn_rewards_v2.erl
+++ b/src/transactions/v2/blockchain_txn_rewards_v2.erl
@@ -321,6 +321,7 @@ calculate_rewards_metadata(Start, End, Chain) ->
         %% We only want to fold over the blocks and transaction in an epoch once,
         %% so we will do that top level work here. If we get a thrown error while
         %% we are folding, we will abort reward calculation.
+        PerfTab = ets:new(rwd_perf, [named_table]),
         Results0 = fold_blocks_for_rewards(Start, End, Chain,
                                            Vars, Ledger, AccInit),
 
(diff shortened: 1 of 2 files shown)


Vagabond:
How many times do we normally inspect a witness for validity?

Vagabond:
Does the pdict need cleaning afterwards?

src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -1451,7 +1502,17 @@ legit_witnesses(Txn, Chain, Ledger, Elem, StaticPath, RegionVars, Version) ->
...
+                ElemHash = erlang:phash2(Elem),
+                ValidWitnesses =
+                    case get({KeyHash, ElemHash}) of
+                        undefined ->
+                            VW = blockchain_txn_poc_receipts_v1:valid_witnesses(Elem, WitnessChannel, RegionVars, Ledger),
+                            put({KeyHash, ElemHash}, VW),
+                            VW;
+                        VW -> VW
+                    end,
+                lager:info("challenge witness ~p", [erlang:phash2(ValidWitnesses)]),
```
Vagabond, line 1515: This is very spammy and unhelpful

Vagabond:
Ok, I added some hit/miss logging:

```
^andrew@tower:: grep -e MISS _build/validator/rel/miner/log/console.log | grep 07:46 | wc -l                                                        
13424
^andrew@tower:: grep -e HIT _build/validator/rel/miner/log/console.log | grep 07:46 | wc -l                                                         
13424
```

So this saves us 1 witness recalc per epoch

Vagabond (approved):
Overall this seems to do the job, if comments are addressed.

PR #1200, opened 2022-01-13
If snapshot file on disk is not deserializeable, remove it

Vagabond:
Prevents a loop in a node like this:

```
2022-01-13 14:25:29.519 21780 [error] <0.17615.0>@blockchain_worker:start_snapshot_sync:1072 snapshot download or loading failed because {badmatch,eof}: [{blockchain_ledger_snapshot_v1,find_pairs_in_file,4,[{file,"blockchain_ledger_snapshot_v1.erl"},{line,1472}]},{blockchain_ledger_snapshot_v1,deserialize,2,[{file,"blockchain_ledger_snapshot_v1.erl"},{line,431}]},{blockchain_worker,attempt_load_snapshot_from_disk,3,[{file,"blockchain_worker.erl"},{line,1195}]},{blockchain_worker,'-start_snapshot_sync/4-fun-0-',6,[{file,"blockchain_worker.erl"},{line,1065}]}]
```

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -1062,7 +1062,7 @@ start_snapshot_sync(Hash, Height, Peer,
                                       %% if the file doesn't deserialize correctly, it will
                                       %% get deleted, so we can redownload it on some other
                                       %% attempt
-                                      attempt_load_snapshot_from_disk(Filename, ConfigHash, Chain);
+                                      ok = attempt_load_snapshot_from_disk(Filename, ConfigHash, Chain);
                                   _ ->
                                       %% don't do anything
                                       ok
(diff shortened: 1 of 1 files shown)


src/blockchain_worker.erl
```diff
@@ -1062,7 +1062,7 @@ start_snapshot_sync(Hash, Height, Peer,
                                       %% if the file doesn't deserialize correctly, it will
                                       %% get deleted, so we can redownload it on some other
                                       %% attempt
-                                      attempt_load_snapshot_from_disk(Filename, ConfigHash, Chain);
+                                      ok = attempt_load_snapshot_from_disk(Filename, ConfigHash, Chain);
```
madninja, line 1065: So attempt_load_snapshot_from_disk can still return an error tuple right? Do you want it to crash here?

madninja: ah intentional, ok

madninja (approved):

PR #1201, opened 2022-01-13
Include optional height with commit hooks

syuan100:
- Update hook record to include request for height
- Grab ledger height from `commit_context` ledger
- Return height in hook calls if requested


src/ledger/v1/blockchain_ledger_v1.hrl
```diff
@@ -4,7 +4,8 @@
          predicate :: undefined | fun(),
          hook_inc_fun :: fun(),  %% fun called for each incremental/partial update for the relevant CF
          hook_end_fun :: fun(),  %% fun called after all incremental/partial updates are complete for the relevant cf
-         ref :: undefined | reference()
+         ref :: undefined | reference(),
+         include_height :: undefined | boolean()
```
madninja, line 8: make false the default and set type to just boolean()

i.e. `include_height=false :: boolean()`

evanmcc (approved):

PR #1206, opened 2022-01-15
don't add an unstable map to the snapshot

evanmcc:


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -4927,17 +4927,15 @@ load_state_channels(SCs, Ledger) ->
...
-                  HexMap)));
+    case blockchain_ledger_v1:get_hexes_list(Ledger) of
+        {ok, Hexes} ->
+            lists:foldl(
+              fun({HexAddr, _Ct}, Acc) ->
+                      {ok, Hex} = get_hex(HexAddr, Ledger),
+                      [{HexAddr, Hex} | Acc]
+              end,
+              [{list, Hexes}],
+              Hexes);
```
Vagabond, line 4938: worth reversing the list here, or does it matter?

evanmcc:
I was thinking about it, and we do it for the raws, for whatever reason, but only the stability matters, rather than ordering, so I don't think we need to?  we could remove it from the raw stuff too

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: I don't think that it matters, as long as it's the same.

Vagabond: but if the hex ordering is scrambled, because of previous issues with the map, will these come out in any reasonable order?

Vagabond (approved):

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -4927,23 +4932,23 @@ load_state_channels(SCs, Ledger) ->
...
         {error, not_found} ->
             []
     end.
 
 load_hexes(Hexes0, Ledger) ->
-    lists:foreach(fun({list, Hexes}) ->
+    lists:foreach(fun({list, Hexes}) when is_list(Hexes) ->
+                          ok = set_hexes_list(Hexes, Ledger);
+                     ({list, Hexes}) when is_map(Hexes) ->
```
xandkar, line 4951: Late comment, but this would've been better to normalize in the snap module during deserialization.

PR #1208, opened 2022-01-18
Check s3 latest-snap.json in another place

Vagabond:


src/blockchain_worker.erl
```diff
@@ -633,8 +637,8 @@ handle_info({'DOWN', SyncRef, process, _SyncPid, Reason},
                     {noreply, schedule_sync(State)};
                 _ when Mode == snapshot ->
                     lager:info("snapshot sync down reason ~p", [Reason]),
-                    {Hash, Height} = State#state.snapshot_info,
-                    {noreply, snapshot_sync(Hash, Height, State)};
+                    %% nested records make your eyes bleed but...
```
evanmcc, line 640: huh?

src/blockchain_worker.erl
```diff
@@ -1030,105 +1028,100 @@ grab_snapshot(Height, Hash) ->
...
+get_latest_snap_data(URL, #snapshot_info{etag=Etag} = SnapInfo) ->
+    Headers0 = [
+               {"user-agent", "blockchain-worker-3"}
               ],
+    Headers = case Etag of
+                  undefined -> Headers0;
+                  _ -> [ {"if-none-match", "\"" ++ Etag ++ "\""} | Headers0 ]
+               end,
     HTTPOptions = [
                    {timeout, 900000}, % milliseconds, 900 sec overall request timeout
```
evanmcc, line 1098: these seem really long for a blocking call in the blockchain_worker

src/blockchain_worker.erl
jadeallenx: Originally I had a nested record update here but I refactored it out. Need to delete that comment.

src/blockchain_worker.erl
jadeallenx: I guess maybe we should asynchronously handle this...

evanmcc (approved):

PR #1212, opened 2022-01-20
continuous block gc

Vagabond:
Problem to solve: Even if we GC bytes on startup, we may run so long we still fill up the disk. This change attempts to keep the blockchain size roughly constant if the user has asked for GC by monitoring new blocks being added and every 10 mb, by default, GCing 10mb more data.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -3582,7 +3582,8 @@ context_cache(Cache, GwCache, Ledger) ->
           {ok, blockchain_block:block()} | {error, any()}.
 get_block(Height, #ledger_v1{blocks_db = DB,
                              heights_cf = HeightsCF} = Ledger) when is_integer(Height) ->
-    case {ok, Height} > current_height(Ledger) of
+    {ok, LedgerHeight} = current_height(Ledger),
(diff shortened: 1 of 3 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3581,28 +3581,45 @@ context_cache(Cache, GwCache, Ledger) ->
 -spec get_block(integer(), ledger()) ->
           {ok, blockchain_block:block()} | {error, any()}.
 get_block(Height, #ledger_v1{blocks_db = DB,
-                             blocks_cf = BlocksCF,
-                             heights_cf = HeightsCF} = Ledger) ->
-    case Height > current_height(Ledger) of
+                             heights_cf = HeightsCF} = Ledger) when is_integer(Height) ->
+    case {ok, Height} > current_height(Ledger) of
```
evanmcc, line 3585:
```
> {ok, 1} > {error, not_found}.
true
```
maybe match this out first?  don't want to get bit by weird erlang comparison semantics.

Vagabond: Fair enough

evanmcc (approved):

PR #1213, opened 2022-01-21
cache height in the poc so gc doesn't need to deserialize

evanmcc:
try to make poc gc a lot more efficient by pre-caching the height somewhere it's cheap to decode.


evanmcc:
closing for now

PR #1214, opened 2022-01-24
Adt/gc h3dex andymck additions

andymck:
- Fixes issues in targeting v4.
- Gates targeting version in use behind a chain var, distinct from poc_version
- Provides a uniform API to add/remove a gateway to hexing rather than individual calls to hex and h3dex
- Adds chain var to determine which hexing type to use ( hexes, h3dex or both ), allows olds hexes support to be stopped when required

Linked miner PR: https://github.com/helium/miner/pull/1387

### DIFF
--- a/src/transactions/v1/blockchain_txn_vars_v1.erl
+++ b/src/transactions/v1/blockchain_txn_vars_v1.erl
@@ -974,10 +974,15 @@ validate_var(?polyfill_resolution, Value) ->
     validate_int(Value, "polyfill_resolution", 0, 15, false);
 validate_var(?h3dex_gc_width, Value) ->
   validate_int(Value, "h3dex_gc_width", 1, 10000, false);
-validate_var(?poc_targeting_version, Value) ->
-  validate_int(Value, "poc_targeting_version", 1, 1000, false);
 validate_var(?poc_target_pool_size, Value) ->
   validate_int(Value, "poc_target_pool_size", 1, 1000000, false);
+validate_var(?poc_targeting_version, Value) ->
+    case Value of
+        3 -> ok;
+        4 -> ok;
+        _ ->
+            throw({error, {invalid_poc_targeting_version, Value}})
+    end;
 
 %% score vars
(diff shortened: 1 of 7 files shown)


src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -972,6 +972,12 @@ validate_var(?check_snr, Value) ->
     end;
 validate_var(?polyfill_resolution, Value) ->
     validate_int(Value, "polyfill_resolution", 0, 15, false);
+validate_var(?h3dex_gc_width, Value) ->
+  validate_int(Value, "h3dex_gc_width", 1, 10000, false);
+validate_var(?poc_targeting_version, Value) ->
+  validate_int(Value, "poc_targeting_version", 1, 1000, false);
```
evanmcc, line 978: should we limit this to existing versions?

Vagabond: Yes

andymck: @evanmcc @Vagabond  not sure i understand the request here but if we limit the possible values to the current version how are we going to bump it ?

vihu: @andymck any future `poc_targeting_version` bump would mandate a code change (this would be by design to not allow differing versions of _valid_ code out in the wild). Refer to `validator_version`/`snapshot_version` or any other *_version style var.

andymck: thanks @vihu.  that makes sense.

evanmcc:
I think that this is a good addition to the other PR, I think they can be merged together for final refinement.  is hexing type validated in the vars?

Vagabond (approved):

evanmcc (approved):

andymck:
@evanmcc the hexing type was not being validated in the txn vars module.  Fixed now....good catch

PR #1216, opened 2022-01-25
Andymck/grpc txn submit

andymck:
linked miner pr: https://github.com/helium/miner/pull/1732
linked proto pr: https://github.com/helium/proto/pull/116

Provides GPRC handler and associated APIs to submit a txn over GRPC.  The GRPC API returns the txn key and the received height to the client.

The txn key can then be used by a client to query the txn and get updated data on the txn's position in the sidecar queue.

In order to accommodate the ability to query the sidecar queue this PR also adds a new txn protocol ( TX_PROTOCOL_V3 ) between the dialer and the CG acceptor.  This utilizes two new PBs ( blockchain_txn_request_v1_pb & blockchain_txn_info_v1_pb ).   The request pb is used by client dialer when submitting a request to the dialed acceptor and can be one of two types, submit or update.

When the request type is `submit`, the dialer is submitting a new txn to sidecar, when the request type is `update` it is querying the status of a previously submitted txn and its position in the sidecar queue/buffer.

The server then returns a response as the PB `blockchain_txn_info_v1_pb`.  This PB is then propagated back to the txn mgr which now only expects responses from dialers using this new format.

The V3 protocol will be used when both client and server support v3, if one side does not then the next highest protocol version will be used. 

The v3 protocol also enables the acceptor/server to return error details to the dialer/client should the txn crash out for example during `is_valid` or `absorb`.  This provides more visibility on validators to identify why a txn fails when handled by a remote acceptor.  An example of how this looks from the client/dialer perspective is below. ( enforced error example ):

2022-06-16 12:49:28.103 [info] <0.402.0>@blockchain_txn_mgr:handle_info:{322,5} txn: type=payment, payer="112kT4PeJEK6Y8gDfr2mavzfoQpcm7r4aNZbnnU2azEJUpuDcrgS", payee="11YSf3N8pSM9BHaQdGc5iM5QaJoWCxkFxifgS3BYyjVGTkkDFUb", amount=1000, fee=0, nonce=1, signature=124GJCVPdHdmaXAGomQ8EX1pWjr5Zn5DLzDeXGAsHLWRmEuDtXhntHuzu9E9J7H63cUNWgnUvcr9hxQQmitJFEuPLH9TpYs7JCr1idDL, failed with reason: <<"case_clause">>, member: <<0,71,100,200,146,108,231,95,32,62,103,230,31,208,254,116,186,202,5,91,102,76,195,140,3,147,205,116,203,51,86,81,245>> Dialer: <0.874.0> Trace [{blockchain_txn_handler,handle_data,3,[{file,"/Users/andymck/helium/miner/_build/default/lib/blockchain/src/handlers/blockchain_txn_handler.erl"},{line,98}]},{libp2p_framed_stream,dispatch_handle_data,3,[{file,"/Users/andymck/helium/miner/_build/default/lib/libp2p/src/libp2p_framed_stream.erl"},{line,405}]},{gen_server,try_dispatch,4,[{file,"gen_server.erl"},{line,695}]},{gen_server,handle_msg,6,[{file,"gen_server.erl"},{line,771}]},{libp2p_multistream_server,handle_msg,2,[{file,"/Users/andymck/helium/miner/_build/default/lib/libp2p/src/libp2p_multistream_server.erl"},{line,61}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,226}]}]


config/grpc_server_gen.config
```diff
@@ -6,11 +6,14 @@
 
 {grpc, [
     {proto_files, [
-        "_build/default/lib/helium_proto/src/service/state_channel.proto"
+        "_build/default/lib/helium_proto/src/service/state_channel.proto",
```
jeffgrunewald, line 9: did i read correctly somewhere (in the proto repo perhaps) that the router and state_channel proto functionality had been consolidated into the gateway service? is this state_channel client needed here any more?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -40,16 +45,35 @@
          terminate/2,
          code_change/3
         ]).
-
+%% TODO: use rejector_pb record
 -type rejection() ::
+    {
+        Member :: libp2p_crypto:pubkey_bin(),
+        Height :: non_neg_integer() | undefined,
```
jeffgrunewald, line 52: some of these will need to be marked optional if we want them to be able to handle `undefined` on the erlang side

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -112,17 +152,35 @@ set_chain(Chain) ->
 txn_list() ->
     gen_server:call(?MODULE, txn_list, infinity).
 
--spec txn_status(blockchain_txn:hash()) -> {ok, map()} | {error, not_found}.
-txn_status(Hash) ->
-    gen_server:call(?MODULE, {txn_status,Hash}, infinity).
+-spec txn_status(txn_key()) -> {ok, pending, map()} | {error, not_found}.
+txn_status(TxnKey) ->
```
jeffgrunewald, line 156: with a protected table this is probably a better approach to accessing this data than running it through a `gen_server:call/3` and avoiding the txn mgr's mailbox; i wonder if we should do the same to some of these other accessors like `txn_list/0`?

andymck: it would def be worth it if those are high volume APIs...i dont think txn_list is used other than perhaps in the CLI.

Any API which is hit by a sibyl handler and thus part of a GRPC request should def use this approach, have it run in the caller's context rather than that of the txn mgr.  Otherwise we risk creating a bottleneck

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -17,16 +21,19 @@
...
+         make_ets_tables/0,
+         current_height/0
         ]).
 
 %% Testing backdoors for CT
 -export([
     force_process_cached_txns/0,
-    get_rejections_deferred/0
+    get_rejections_deferred/0,
+    signatory_rand_members/5
```
jeffgrunewald, line 36: i reverted the wrapper around these "testing" functions b/c they're used for testing in applications downstream of core so when they're wrapped in the testing `ifdef` they cause "function unused" compile errors in those downstream apps. i also moved `signatory_rand_members/5` back since it looked like it was only moved to expose it to miner and was still only intended for testing

jeffgrunewald:
PR is now dependent on https://github.com/helium/erlang-libp2p/pull/443 in order to properly compile (proto compilation under the 2.15.0 version of rebar_gpb_plugin has broken of a sudden for some unknown reason). The libp2p dep needs to be reverted to `master` branch in core's rebar.config before final merge

src/blockchain.app.src
```diff
@@ -29,7 +29,6 @@
         base64url,
         jsx,
         erbloom,
-        grpc_client,
```
jadeallenx, line 32: why is this no longer needed?

src/blockchain_utils.erl
```diff
@@ -312,6 +313,90 @@ addr2name(Addr) ->
...
+            case maps:get(<<"grpc_address">>, libp2p_peer:signed_metadata(Peer), undefined) of
+                undefined ->
+                    %% sort listen addrs, ensure the public ip is at the head
+                    case libp2p_peer:cleared_listen_addrs(Peer) of
+                        [] -> {error, no_listen_addrs};
+                        ClearedListenAddrs ->
+                            case libp2p_transport:sort_addrs_with_keys(SwarmTID, ClearedListenAddrs) of
+                                [] -> {error, no_listen_addrs};
+                                [H | _] ->
+                                    case has_addr_public_ip(H) of
```
jadeallenx, line 345: does this make testing harder if the tests have no public IP?

src/handlers/blockchain_txn_handler.erl
```diff
@@ -51,50 +85,153 @@ init(client, _Conn, [Path, Parent, TxnHash]) ->
...
+encode_response(?TX_PROTOCOL_V2, txn_failed, Details, _Trace, _PosInQueue, _QueueLen, _Height)  ->
+    Details;
+encode_response(?TX_PROTOCOL_V2, txn_rejected, _Details, _Trace, _PosInQueue, _QueueLen, Height) ->
+    <<"error", Height/integer>>;
+%% marshall v3 response format
+encode_response(?TX_PROTOCOL_V3, Resp, Details, Trace, QueuePos, QueueLen, Height)  ->
+    Msg = #blockchain_txn_info_v1_pb{
+        result = atom_to_binary(Resp, utf8),
+        details = atom_to_binary(Details, utf8),
+        trace = term_to_binary(Trace),
```
jadeallenx, line 188: should there be some kind of maximum size here? just to ensure we don't have like some runaway stack trace send out a ginormous response?

src/handlers/blockchain_txn_handler.erl
```diff
@@ -51,50 +85,153 @@ init(client, _Conn, [Path, Parent, TxnHash]) ->
...
+    v2_to_v3(txn_rejected, <<"error">>, Height);
+%% v3 -> v3
+decode_response(?TX_PROTOCOL_V3, Resp) when is_binary(Resp) ->
+    blockchain_txn_handler_pb:decode_msg(Resp, blockchain_txn_info_v1_pb).
+
+v1_to_v3(Status, Details) when is_atom(Status)->
+    #blockchain_txn_info_v1_pb{
+        result = atom_to_binary(Status, utf8),
+        details = Details,
+        height = 0,
```
jadeallenx, line 220: would -1 be a better signal here? maybe the original protocol version should be part of the message?

src/handlers/blockchain_txn_handler.erl
```diff
@@ -51,50 +85,153 @@ init(client, _Conn, [Path, Parent, TxnHash]) ->
...
+        queue_pos = 0,
+        queue_len = 0
+    }.
+
+v2_to_v3(Status, Details, Height) when is_atom(Status)->
+    #blockchain_txn_info_v1_pb{
```
jadeallenx, line 226: same comment? Maybe we should include the original protocol version?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -385,6 +468,8 @@ handle_add_block_event({add_block, BlockHash, Sync, _Ledger}, State=#state{chain
             %% only update the current block height if its not a sync block
             NewCurBlockHeight = maybe_update_block_height(CurBlockHeight, BlockHeight, Sync),
             lager:debug("received block height: ~p,  updated state block height: ~p", [BlockHeight, NewCurBlockHeight]),
+            %% cache the current height
+            ets:insert(?CACHE, {?CUR_HEIGHT, NewCurBlockHeight}),
```
jadeallenx, line 472: Are you at all concerned that the cached height will be out of sync with the actual ledger height?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -813,8 +947,6 @@ reject_f(NumMembers)->
     %% 2F+1
     (trunc((NumMembers) div 3) * 2) + 1.
 
--spec get_txn_key()-> txn_key().
+-spec get_txn_key()-> blockchain_txn_mgr:txn_key().
 get_txn_key()->
-    %% define a unique value to use as they cache key for the received txn, for now its just a mono increasing timestamp.
-    %% Timestamp is a poormans key but as txns are serialised via a single txn mgr per node, it satisfies the need here
-    erlang:monotonic_time().
+    integer_to_binary(erlang:monotonic_time()).
```
jadeallenx, line 952: I want to believe this but the assertion it's "good enough" makes my distributed systems heart sad

src/transactions/blockchain_txn_mgr.erl
andymck: The cache height should be being updated at the same point we were already updating the state height.  So it cannot be any more out of sync that it has been.  The height is derived from the block derived from the block hash taken from the block event

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -763,10 +877,30 @@ dial_members(Members, Chain, TxnKey, Txn)->
...
+maybe_query_acceptors([], _TxnKey, _Txn, _CurBlockHeight) ->
+    ok;
+maybe_query_acceptors([{M, H, _QPos, _QLen} | Rest], TxnKey, Txn, CurBlockHeight)->
+    %% NOTE: we deliberately are not saving dialers for updates
+    %% ie we dont add them to the txn data's list of dialers
+    %% as doing so would throw off the calculations to
+    %% determine if we have submitted to sufficient CG members
+    %% we also dont want to retry failed dials for update requests
+    %% they will be auto run again in N blocks
+    case (CurBlockHeight - H) rem 5 == 0 of
```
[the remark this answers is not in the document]
andymck: Yeah def should be set at least as a macro.  A chain var would be nice but maybe overkill

5 was an arbitrary value, if you were waiting a txn to be accepted 5 feels like a reasonable period of time in which the user should get an update on the txns latest status

src/blockchain.app.src
andymck: it is not required by core, core acts as a grpc server, never as a client.  The only times its required by core is as part of CT runs and for this it is added as a test dep

src/transactions/blockchain_txn.erl
```diff
@@ -615,9 +615,10 @@ is_valid(Txn, Chain) ->
                 Res ->
                     Res
             catch
-                What:Why:Stack ->
+                _What:Why:Stack ->
                     lager:warning("crash during validation: ~p ~p", [Why, Stack]),
-                    {error, {Type, What, {Why, Stack}}}
+                    %% TODO: make error reason more descriptive
+                    {error, validation_failed}
```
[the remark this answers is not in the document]
andymck: addressed here: 14eff380e114fe26c79214249656a67dd2a91fb4

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -398,10 +483,10 @@ process_deferred_rejections(
         rejections_deferred = Deferred0,
         reject_f            = RejectF,
         cur_block_height    = CurBlockHeight
-    }=State
+    } = State
 ) ->
-    IsPast    = fun({_, _, _, _, H}) -> H   < CurBlockHeight end,
-    IsCurrent = fun({_, _, _, _, H}) -> H =:= CurBlockHeight end,
+    IsPast    = fun({_, _, _, _, #blockchain_txn_info_v1_pb{height = H}}) -> H < CurBlockHeight end,
```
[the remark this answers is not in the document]
andymck: I cant really do that, these are funs which iterate over a list of tuples ( of type deferred_rejection )
I did however add some details which should make it easier to follow down the road: f8ca4b687fbe53157c0e094edd25513f46881eae

src/transactions/blockchain_txn_mgr.erl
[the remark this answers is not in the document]
andymck: added macro: f97dd8f29c188cf21a309b84049f3d09765046a2

src/blockchain_utils.erl
andymck: it would but there are no tests in core which encounter it.  Miner tho has tests which require a public ip and for that it fakes a public IP for each spawned node

src/handlers/blockchain_txn_handler.erl
andymck: the height value is used to determine if the remote val is current or not and as such a height of zero is prob more representative here.

Added the protocol version as part of this commit: eb6b0b831ddf33f4149d60b762bdb426a9ab4327

src/handlers/blockchain_txn_handler.erl
andymck: added: eb6b0b831ddf33f4149d60b762bdb426a9ab4327

src/transactions/blockchain_txn.erl
```diff
@@ -58,8 +58,8 @@
 -callback json_type() -> binary() | atom().
 -callback hash(State::any()) -> hash().
 -callback sign(txn(), libp2p_crypto:sig_fun()) -> txn().
--callback is_valid(txn(), blockchain:blockchain()) -> ok | {error, any()}.
--callback absorb(txn(),  blockchain:blockchain()) -> ok | {error, any()}.
+-callback is_valid(txn(), blockchain:blockchain()) -> ok | {error, atom()} | {error, {atom(), any()}}.
+-callback absorb(txn(),  blockchain:blockchain()) -> ok | {error, atom()} | {error, {atom(), any()}}.
```
jeffgrunewald, line 62:
```suggestion
-callback is_valid(txn(), blockchain:blockchain()) -> ok | {error, atom() | {atom(), any()}}.
-callback absorb(txn(),  blockchain:blockchain()) -> ok | {error, atom() | {atom(), any()}}.
```

src/transactions/blockchain_txn.erl
```diff
@@ -560,7 +560,7 @@ absorb_block(Block, Rescue, Chain) ->
 %% @doc
 %% @end
 %%--------------------------------------------------------------------
--spec absorb(txn(),blockchain:blockchain()) -> ok | {error, any()}.
+-spec absorb(txn(),blockchain:blockchain()) -> ok | {error, atom()} | {error, {atom(), any()}}.
```
jeffgrunewald, line 563:
```suggestion
-spec absorb(txn(),blockchain:blockchain()) -> ok | {error, atom() | {atom(), any()}}.
```

src/transactions/blockchain_txn.erl
```diff
@@ -606,7 +606,7 @@ print(Txn, Verbose) ->
 %% @doc
 %% @end
 %%--------------------------------------------------------------------
--spec is_valid(txn(), blockchain:blockchain()) -> ok | {error, any()}.
+-spec is_valid(txn(), blockchain:blockchain()) -> ok | {error, atom()} | {error, {atom(), any()}}.
```
jeffgrunewald, line 609:
```suggestion
-spec is_valid(txn(), blockchain:blockchain()) -> ok | {error, atom() | {atom(), any()}}.
```

src/transactions/blockchain_txn.erl
jeffgrunewald: take 'em or leave 'em but i believe nesting these union types is the convention i see in the official docs most often

src/handlers/blockchain_txn_handler.erl
```diff
@@ -43,6 +46,37 @@ server(Connection, Path, _TID, Args) ->
...
+decode_request(?TX_PROTOCOL_V1 = _Path, Bin) ->
+    Txn = blockchain_txn:deserialize(Bin),
+    #blockchain_txn_request_v1_pb{type = submit, txn = Txn};
+decode_request(?TX_PROTOCOL_V2 = _Path, Bin) ->
+    Txn = blockchain_txn:deserialize(Bin),
+    #blockchain_txn_request_v1_pb{type = submit, txn = Txn};
+decode_request(?TX_PROTOCOL_V3 = _Path, Bin) ->
+    #blockchain_txn_request_v1_pb{txn = WrappedTxn} =
+        Msg = blockchain_txn_handler_pb:decode_msg(Bin, blockchain_txn_request_v1_pb),
+    Msg#blockchain_txn_request_v1_pb{txn = blockchain_txn:unwrap_txn(WrappedTxn)}.
```
jeffgrunewald, line 78: are these really only exported for completeness? it looks like `encode_request/4` is the only function here that's ever called outside of the context of the the handler module? it seems weirdly one-sided to have the abstraction leak outside of this module just for sending messages but i don't see a good reason to `decode_response/2` on the receiving side (in the dialer) when you can handle it here in the client handler. mostly just thinking out loud i guess...

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -222,55 +273,86 @@ handle_call(_Msg, _From, State) ->
...
-handle_info({send_failed, {Dialer, TxnKey, Txn, Member}}, State) ->
+handle_info({send_failed, {submit, {Dialer, TxnKey, Txn, Member}}}, State) ->
     lager:debug("txn: ~s, send_failed: ~p, Dialer: ~p", [blockchain_txn:print(Txn), Member, Dialer]),
     ok = retry(TxnKey, Txn, Dialer),
     {noreply, State};
 
-handle_info({accepted, {Dialer, TxnKey, Txn, Member}}, State) ->
-    lager:debug("txn: ~s, accepted_by: ~p, Dialer: ~p", [blockchain_txn:print(Txn), Member, Dialer]),
-    ok = accepted(TxnKey, Txn, Member, Dialer),
+%% dialed CG member related failures
```
jeffgrunewald, line 293: does this comment need to move up a few clauses or else down to the `when Status == <<"txn_failed">>` one?

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -8,25 +8,33 @@
...
 -export([
          start_link/1,
          submit/2, submit/3,
+         grpc_submit/1, grpc_submit/2,
          set_chain/1,
          txn_list/0,
          txn_status/1,
-         make_ets_table/0
+         make_ets_tables/0,
+         current_height/0
```
evanmcc, line 30: not sure why this is exported, it doesn't seem to be used.   I'm also not sure why we couldn't just cache this in the mgr state rather than using an ets table?

evanmcc (approved):
I won't pretend to a super-comprehensive review of this one, but it's looking good so far.  let's get some more devnet/testnet milage on it before we merge, though.

src/transactions/blockchain_txn_mgr.erl
jeffgrunewald: https://github.com/helium/blockchain-core/blob/e2fb08052b05865ff796d63578de20a72640b25c/src/transactions/blockchain_txn_mgr.erl#L146
i believe it's to avoid a blocking call to get the state when sending back height, particularly with the grpc api

andymck: Correct, client can enquire as to the status of a txn via GRPC.  The response to that needs to include the height.  By including the height in ETS, the response can be fulfilled from within the stream context.

PR #1217, opened 2022-01-25
Fix some tests and restore OTP 22 support

Vagabond:


src/blockchain_term.erl
```diff
@@ -494,7 +494,7 @@ atom_ext(<<Bin/binary>>) ->
 -spec bin_to_atom(frame(), binary(), binary()) -> result_internal(atom()).
 bin_to_atom(F, <<AtomName/binary>>, <<Rest/binary>>) ->
     try
-        {ok, {binary_to_atom(AtomName), Rest}}
+        {ok, {list_to_atom(unicode:characters_to_list(AtomName)), Rest}}
```
jeffgrunewald, line 497: is it not possible to restrict this to `list_to_existing_atom/1`?

jeffgrunewald: and is the otp 23 version preferable to this legacy solution such that we would want to wrap it in a macro that checks if the otp version >= 23 and then the macro would better identify a cleanup spot once we're no longer supporting otp 22?

evanmcc (approved):

PR #1222, opened 2022-01-28
conditionally include the common_test lib

lthiery:
I was trying a source build and apparently the erlang release I have doesn't have the test libs. But we should be able to remove this when not doing a test build?


evanmcc (approved):

src/transactions/v2/blockchain_txn_assert_location_v2.erl
```diff
@@ -4,7 +4,9 @@
 %% @end
 %%%-------------------------------------------------------------------
 -module(blockchain_txn_assert_location_v2).
+-ifdef(TEST).
 -include_lib("common_test/include/ct.hrl").
+-endif
```
xandkar, line 9: Missing a period after `endif`.

PR #1232, opened 2022-02-02
Check file hash before downloading snap

jadeallenx:
Problem to solve: We need to be more careful about downloading snapshots because they are very large.

Solution: Check the file hash (not the same as the snapshot hash, which is internal to the snapshot data structures) before we download another snap image.

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -1,4 +1,4 @@
-%%%-------------------------------------------------------------------
+% %%-------------------------------------------------------------------
 %% @doc
 %% == Blockchain Core Worker ==
 %% @end
@@ -76,6 +76,7 @@
 -define(SERVER, ?MODULE).
 -define(READ_SIZE, 16 * 1024). % read 16 kb chunks
 -define(WEEK_OLD_SECONDS, 7*24*60*60). %% a week's worth of seconds
+-define(MAX_ATTEMPTS, 3).
 
 -ifdef(TEST).
 -define(SYNC_TIME, 1000).
@@ -88,7 +89,11 @@
 -record(snapshot_info, {
           hash = <<>> :: snap_hash(),
           height = 0 :: integer(),
-          etag = undefined :: undefined | string()
+          etag = undefined :: undefined | string(),
+          file_hash = undefined :: undefined | binary(),
+          file_size = undefined :: undefined | pos_integer(),
+          last_success = 0 :: non_neg_integer(), %% posix time of last successful download
+          download_attempts = 0 :: non_neg_integer()
          }).
 
 -type snapshot_info() :: #snapshot_info{}.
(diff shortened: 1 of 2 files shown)


src/blockchain_worker.erl
```diff
@@ -1,4 +1,4 @@
-%%%-------------------------------------------------------------------
+% %%-------------------------------------------------------------------
```
jeffgrunewald, line 1:
```suggestion
%%%-------------------------------------------------------------------
```

src/blockchain_worker.erl
```diff
@@ -615,8 +620,33 @@ handle_info(snapshot_timer_tick, State) ->
...
+    %% Schedule a sync "as usual" in "normal" mode
+    lager:info("snapshot process completed normally; scheduling normal sync."),
+    {noreply, schedule_sync(State#state{mode=normal,
+                       snapshot_info = SnapInfo#snapshot_info{ last_success = erlang:system_time(seconds) }})};
+
+%% snapshot attempt failed, so we will retry upto MAX ATTEMPTS
+handle_info({'DOWN', SyncRef, process, _SyncPid, Reason},
+            #state{sync_ref = SyncRef, mode = snapshot,
+                   snapshot_info = #snapshot_info{ download_attempts = Attempts } = SnapInfo} = State) when Attempts < ?MAX_ATTEMPTS ->
+    lager:error("Snapshot attempt ~p exited with: ~p; retrying", [Attempts, Reason]),
```
jeffgrunewald, line 637:
```suggestion
    lager:warning("Snapshot attempt ~p exited with: ~p; retrying", [Attempts, Reason]),
```

src/blockchain_worker.erl
```diff
@@ -615,8 +620,33 @@ handle_info(snapshot_timer_tick, State) ->
...
     {noreply, maybe_sync(State)};
+
+handle_info({'DOWN', SyncRef, process, _SyncPid, normal},
+            #state{sync_ref = SyncRef, mode = snapshot, snapshot_info = SnapInfo} = State) ->
+    %% snapshot process completed normally;
+    %%
+    %% Schedule a sync "as usual" in "normal" mode
+    lager:info("snapshot process completed normally; scheduling normal sync."),
+    {noreply, schedule_sync(State#state{mode=normal,
+                       snapshot_info = SnapInfo#snapshot_info{ last_success = erlang:system_time(seconds) }})};
```
jeffgrunewald, line 631: we should probably reset the number of `download_attempts` here, too so you don't fail once or twice, then succeed before the max attempts which looks like it'd put the snap info back into the sync queue with a running balance of download attempts so the next time it would start with fewer than 3.

looks like you could safely do this in the `schedule_sync/1` since you're calling that in both cases where we're done attempting to download the latest snap and passing in the state

jadeallenx: yeah good catch!

src/blockchain_worker.erl
jeffgrunewald: not strictly necessary since we're only trying a max of 3 times but if we ever changed this value doing all the failures at error level could get a bit noisy as opposed to warning on individual failed attempts and then erroring when we finally give up

jeffgrunewald (approved):

evanmcc (approved):
ideally the errors would be changed to warnings

jeffgrunewald (approved):

PR #1234, opened 2022-02-04
exclude infos, like blocks, from the internal hash

evanmcc:
they contain maps, so the ordering is going to be different, and they are not properly part of the ledger.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -380,10 +380,18 @@ serialize_v6(#{version := v6}=Snapshot0, BlocksOrNoBlocks) ->
             noblocks ->
                 EmptyListBin
         end,
+    Infos =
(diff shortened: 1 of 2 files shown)


xandkar (approved):

Vagabond:
You need a clause in hash() line 810 to handle it as well.

Vagabond:
I tested this by doing the following:

* Booting a validator from the latest blessed snap
* Taking the same snapshot again and checking the hash matched the sys.config hash
* Let the node sync a bit and then take a second snap on OTP22
* Rebuild the release with OTP24 and take the snap again at the same height
* Compare the hashes between OTP22 and OTP24

One thing I'm not sure about: can older nodes load a snap that is hashed differently? How did this node load a blessed snap that was hashed under the old rules? Does downloading from s3 bypass hash checking?

jadeallenx:
> downloading from s3 bypass hash checking?

It uses whatever internal hash is encoded in latest-snap.json - that hash comes from the snapshot info cmd line output

PR #1235, opened 2022-02-04
Fix badly formed 'endif'

macpie:
Fix for
```
===> Compiling _build/default/lib/blockchain/src/transactions/v2/blockchain_txn_assert_location_v2.erl failed
_build/default/lib/blockchain/src/transactions/v2/blockchain_txn_assert_location_v2.erl:11:1: badly formed 'endif'
_build/default/lib/blockchain/src/transactions/v2/blockchain_txn_assert_location_v2.erl:705:1: unterminated '-ifdef'
```
when running `./rebar3 as test compile`

### DIFF
--- a/src/transactions/v2/blockchain_txn_assert_location_v2.erl
+++ b/src/transactions/v2/blockchain_txn_assert_location_v2.erl
@@ -4,13 +4,10 @@
 %% @end
 %%%-------------------------------------------------------------------
 -module(blockchain_txn_assert_location_v2).
--ifdef(TEST).
--include_lib("common_test/include/ct.hrl").
--endif
 
 -behavior(blockchain_txn).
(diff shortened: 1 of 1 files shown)


jeffgrunewald:
looks like it's just missing a period; wouldn't this crash something in the tests in that module if that header needs to be referenced or is that an old include that's not used anymore?

macpie:
> looks like it's just missing a period; wouldn't this crash something in the tests in that module if that header needs to be referenced or is that an old include that's not used anymore?

Could not see the dot thing thanks but then I realized that CT is not needed in there I ran the test without it fine.

jeffgrunewald (approved):

evanmcc (approved):

PR #1236, opened 2022-02-08
Finer-grained validation PR 1 of 3: break-up interface into stages

xandkar:
with currently no-op implementations.

Step 2 will insert possibly-redundant implementations.
Step 3 will remove some redundancy from `Txn:is_valid/2` implementations.

### DIFF
--- a/src/transactions/blockchain_txn.erl
+++ b/src/transactions/blockchain_txn.erl
@@ -50,14 +50,29 @@
 
 -type before_commit_callback() :: fun((blockchain:blockchain(), blockchain_block:hash()) -> ok | {error, any()}).
 -type txns() :: [txn()].
--export_type([hash/0, txn/0, txns/0]).
+-export_type([hash/0, txn/0, txns/0, is_prompt/0]).
 
 -callback fee(txn()) -> non_neg_integer().
 -callback fee_payer(txn(), blockchain_ledger_v1:ledger()) -> libp2p_crypto:pubkey_bin() | undefined.
 -callback json_type() -> binary() | atom().
 -callback hash(State::any()) -> hash().
 -callback sign(txn(), libp2p_crypto:sig_fun()) -> txn().
+
+%% Check the transaction has the required fields and they're well formed and
+%% in-bounds. Use contracts heavily here.
+-callback is_well_formed(txn()) -> ok | {error, {contract_breach, any()}}.
+
+%% Check the txn has the right causal information (nonce, block height, etc) to
+%% be absorbed.  This should be quick.
+-callback is_prompt(txn(), blockchain:blockchain()) ->
+    {ok, is_prompt()} | {error, term()}.
+-type is_prompt() ::
+    yes | no | {not_yet, Delta :: pos_integer()}.
+
+%% Final heavy-weight validity checks, including signature verification and
+%% other complex calculations:
 -callback is_valid(txn(), blockchain:blockchain()) -> ok | {error, any()}.
+
 -callback absorb(txn(),  blockchain:blockchain()) -> ok | {error, any()}.
 -callback print(txn()) -> iodata().
 -callback print(txn(), boolean()) -> iodata().
@@ -595,9 +610,24 @@ is_valid(Txn, Chain) ->
     Type = ?MODULE:type(Txn),
     case lists:keysearch(Type, 1, ?ORDER) of
         {value, _} ->
-            try Type:is_valid(Txn, Chain) of
-                Res ->
-                    Res
+            try
+                case Type:is_well_formed(Txn) of
+                    {error, _}=Err ->
+                        Err;
+                    ok ->
+                        case Type:is_prompt() of
+                            {ok, yes} ->
+                                Type:is_valid(Txn, Chain);
+                            {ok, no} ->
+                                {error, txn_too_late_or_too_early};
+                            {ok, {not_yet, _Delta}} ->
+                                % TODO Bound delta?
+                                % TODO Anything more interesting that can be done here?
+                                {error, txn_too_early};
+                            {error, _}=Err ->
+                                Err
+                        end
+                end
             catch
                 What:Why:Stack ->
                     lager:warning("crash during validation: ~p ~p", [Why, Stack]),
--- a/src/transactions/v1/blockchain_txn_add_gateway_v1.erl
+++ b/src/transactions/v1/blockchain_txn_add_gateway_v1.erl
@@ -34,6 +34,8 @@
     is_valid_payer/1,
     is_valid_staking_key/2,
     is_valid/2,
+    is_well_formed/1,
+    is_prompt/2,
     absorb/2,
     calculate_fee/2, calculate_fee/5, calculate_staking_fee/2, calculate_staking_fee/5,
     print/1,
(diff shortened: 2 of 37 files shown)


src/transactions/blockchain_txn.erl
```diff
@@ -595,9 +610,24 @@ is_valid(Txn, Chain) ->
...
+                        Err;
+                    ok ->
+                        case Type:is_prompt() of
+                            {ok, yes} ->
+                                Type:is_valid(Txn, Chain);
+                            {ok, no} ->
+                                {error, txn_too_late_or_too_early};
+                            {ok, {not_yet, _Delta}} ->
+                                % TODO Bound delta?
+                                % TODO Anything more interesting that can be done here?
```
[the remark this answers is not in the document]
xandkar: Agreed.

src/transactions/blockchain_txn.erl
```diff
@@ -50,14 +50,29 @@
 
 -type before_commit_callback() :: fun((blockchain:blockchain(), blockchain_block:hash()) -> ok | {error, any()}).
 -type txns() :: [txn()].
--export_type([hash/0, txn/0, txns/0]).
+-export_type([hash/0, txn/0, txns/0, is_prompt/0]).
```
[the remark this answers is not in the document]
Vagabond: Prompt is one of those overloaded words, I guess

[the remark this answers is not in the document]
Vagabond: `punctual` or `timely`?

src/transactions/blockchain_txn.erl
```diff
@@ -590,14 +605,33 @@ print(Txn, Verbose) ->
 %% @doc
 %% @end
 %%--------------------------------------------------------------------
--spec is_valid(txn(), blockchain:blockchain()) -> ok | {error, any()}.
+-spec is_valid(txn(), blockchain:blockchain()) ->
+    ok | {error, Reason} when
+    Reason
+        :: txn_too_late
+        |  {txn_too_early, Delta :: pos_integer()}
+        |  any() % TODO Spec other errors explicitly.
```
Vagabond, line 613: Can/should we do this?

src/transactions/blockchain_txn.erl
```diff
@@ -590,14 +605,33 @@ print(Txn, Verbose) ->
...
-                    Res
+            try
+                case Type:is_well_formed(Txn) of
+                    {error, _}=Err ->
+                        Err;
+                    ok ->
+                        case Type:is_prompt() of
+                            {ok, yes} ->
+                                Type:is_valid(Txn, Chain);
+                            {ok, no} ->
```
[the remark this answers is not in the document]
xandkar: Ah, so these are layered abstractions. Demarcating the inherent difference between a fact and its delivery to the caller.

The ok|error indicates whether the operation was able to be completed at all, and if yes - the ok carries the resulting value of the operation.

Since implementations are expected to interact with the chain - they may fail to complete at all.

Even if ultimately it won't matter much for which class of reasons the validation fails (semantic or IO) in most cases, just having a de-conflated API clears the way for more-accurate and less-surprising testing and general reasoning about the system.

As an analogue, I claim that a database lookup should ideally have this signature:

```erl
-spec fetch(connection(), key()) ->
    {ok, none | {some, value()}} | {error, io_error()}.
```

src/transactions/blockchain_txn.erl
xandkar: It is certainly possible and would enhance reasoning.

xandkar: But isn't necessary to be done for _this_ set of PRs.

src/transactions/blockchain_txn.erl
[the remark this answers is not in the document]
xandkar: These are near-synonyms with less-fitting meanings, in my less-than-native understanding of the English tongue :)

"punctual" seems to describe a function's tendency to produce prompt outputs, while "timely" is closer, yet looser in meaning. "prompt" suggests an acceptable position on the timeline, which is what we're asking about here.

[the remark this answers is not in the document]
Vagabond: Yeah, I concede the point.

src/transactions/blockchain_txn.erl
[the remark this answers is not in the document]
Vagabond: Right, ok, this makes sense now.

Vagabond:
```
=== === Reason: no match of right hand side value 
                 {error,
                     {blockchain_txn_transfer_hotspot_v1,error,
                         {undef,
                             [{blockchain_txn_transfer_hotspot_v1,is_prompt,
                                  [],[]},
```

Vagabond:
```
=== === Reason: {assertMatch,
                  [{module,blockchain_price_oracle_SUITE},
                   {line,1303},
                   {expression,
                    "blockchain_txn : is_valid ( SignedPayerAddGatewayTx02 , Chain )"},
                   {pattern,"{ error , payer_invalid_staking_key }"},
                   {value,
                    {error,
                     {blockchain_txn_add_gateway_v1,error,
                      {undef,
                       [{blockchain_txn_add_gateway_v1,is_prompt,[],[]},
                        {blockchain_txn,is_valid,2,
```

src/transactions/blockchain_txn.erl
```diff
@@ -592,14 +607,33 @@ print(Txn, Verbose) ->
...
         {value, _} ->
-            try Type:is_valid(Txn, Chain) of
-                Res ->
-                    Res
+            try
+                case Type:is_well_formed(Txn) of
+                    {error, _}=Err ->
+                        Err;
+                    ok ->
+                        case Type:is_prompt() of
```
Vagabond, line 626: This needs to be called with arguments?

xandkar: Gah. Fixed.

src/transactions/blockchain_txn.erl
```diff
@@ -592,14 +607,33 @@ print(Txn, Verbose) ->
...
         {value, _} ->
-            try Type:is_valid(Txn, Chain) of
-                Res ->
-                    Res
+            try
+                case Type:is_well_formed(Txn) of
+                    {error, _}=Err ->
+                        Err;
+                    ok ->
+                        case Type:is_prompt(Txn, Chain) of
```
Vagabond, line 626: It would be much better to pass the ledger here? We want to avoid chain access during txn checking (you can get at blocks through the ledger now).

xandkar: Yeah, that was the initial goal, but there are uses of the chain in some transactions that were not easy to remove. I don't remember exactly what off the top of my head, but can find it in a few.

Vagabond: Would prefer we fix them than avoid it, and ideally fix is_valid as well in the later PRs.

Vagabond: The only safe way to use the chain is via the handles/apis in the ledger.

xandkar: That is for `is_prompt` implementations. `is_valid` ones are hairier.

xandkar: Switched `is_prompt` to work with the ledger instead of chain :heavy_check_mark:

PR #1238, opened 2022-02-09
State Channel  Re-close test

michaeldjeffrey:
Ensure that submitting a non-dispute state channel close will not make it through when a state channel has already been legitimately closed.

Also update the types of txns a state_channel_client will consider disputable to only those from the original state channel owner. To prevent filing disputes against other disputers.

### DIFF
--- a/src/state_channel/blockchain_state_channels_client.erl
+++ b/src/state_channel/blockchain_state_channels_client.erl
@@ -275,7 +275,10 @@ handle_info({blockchain_event, {add_block, BlockHash, false, Ledger}},
             Block ->
                 lists:foldl(
                     fun(T, Acc) ->
-                        case blockchain_txn:type(T) == blockchain_txn_state_channel_close_v1 of
+                        case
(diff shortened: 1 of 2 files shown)


Vagabond:
What happens if the client has a newer version of the state channel that is legitimate?

macpie:
> What happens if the client has a newer version of the state channel that is legitimate?

That's the ultimate issue because at that point it means that server side will start "lying"

Vagabond:
> > What happens if the client has a newer version of the state channel that is legitimate?
> 
> That's the ultimate issue because at that point it means that server side will start "lying"

No, that's not true, the router may just be offline.

Imagine this scenario:

Router A is offline, but before it went offline it paid B with a nonce of 20 and also paid C with a nonce of 21.

The state channel expiry time has arrived, A is offline so it cannot close. B sends the latest copy of the state channel with nonce 20. Then C sends its close with a nonce of 21 which should *supercede* the nonce 20 close, but not conflict with it.

Vagabond:
@michaeldjeffrey ping

PR #1253, opened 2022-02-24
Update blockchain_worker.erl

joecaswell:
get_quick_sync_height_and_hash should return a tuple

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -1346,6 +1346,7 @@ get_blessed_snapshot_height_and_hash() ->
             {undefined, undefined}
     end.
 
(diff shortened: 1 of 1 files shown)


src/blockchain_worker.erl
```diff
@@ -1346,14 +1346,16 @@ get_blessed_snapshot_height_and_hash() ->
             {undefined, undefined}
     end.
 
+-spec get_quick_sync_height_and_hash(atom()) -> undefined | {non_neg_integer(), non_neg_integer()}.
```
Vagabond, line 1349: the first element of the tuple is a binary, no?

joecaswell: good catch, fixed.

Vagabond (approved):

jadeallenx (approved):
lgtm

PR #1254, opened 2022-02-25
Download and deserialize compressed snapshots

jadeallenx:
Problem to solve: Snapshots are really big. We can save bandwidth costs by compressing the files.

Solution: Preferentially download compressed snapshots if available. Make sure that we can deserialize compressed files appropriately.


src/blockchain_worker.erl
```diff
@@ -1158,10 +1160,21 @@ get_latest_snap_data(URL, SnapInfo) ->
...
+                                  end,
             NewEtag = get_etag(RespHeaders),
-            lager:debug("new latest-json data: previous etag: ~p; height: ~p, internal snapshot hash: ~p, file hash: ~p, file size: ~p, new etag: ~p",
-                        [Etag, Height, B64Hash, MaybeFileHash, MaybeFileSize, NewEtag]),
-            SnapInfo#snapshot_info{height=Height, hash=Hash, file_hash=MaybeFileHash, file_size=MaybeFileSize, etag=NewEtag};
+            lager:debug("new latest-json data: previous etag: ~p; height: ~p, internal snapshot hash: ~p, file hash: ~p, file size: ~p, compressed size: ~p, compressed hash: ~p, new etag: ~p",
+                        [Etag, Height, B64Hash, MaybeFileHash, MaybeFileSize, MaybeCompressedSize, MaybeCompressedHash, NewEtag]),
+            SnapInfo#snapshot_info{height=Height, hash=Hash,
+                                   file_hash=MaybeFileHash, file_size=MaybeFileSize,
+                                   compressed_size=MaybeCompressedSize, compressed_hash=MaybeCompressedHash,
```
Vagabond, line 1176: I'd suggest, if the json contains the compressed data, just putting that into the record and not adding new fields? Maybe simply have a boolean field for if it's compressed or not?

Vagabond (approved):

src/blockchain.erl
```diff
@@ -1997,6 +2000,42 @@ save_bin_snapshot(DestFilename, BinSnap) when is_binary(BinSnap); is_list(BinSna
     ok = filelib:ensure_dir(DestFilename),
     file:write_file(DestFilename, BinSnap).
 
+-spec save_compressed_bin_snapshot(file:filename_all(), blockchain_ledger_snapshot:snapshot()) ->
+   ok | {error, Error :: term()}.
+save_compressed_bin_snapshot(DestFilename, {file, Filename}) ->
```
Vagabond, line 2005: A version of this that handles a binary snapshot as well?

madninja: wouldn't that just be a version of save_bin_snapshot then?

src/blockchain.erl
```diff
@@ -1997,6 +2000,59 @@ save_bin_snapshot(DestFilename, BinSnap) when is_binary(BinSnap); is_list(BinSna
...
+            end,
+   file:close(In),
+   file:close(Out),
+   RetVal;
+save_compressed_bin_snapshot(DestFilename, BinSnap) ->
+   ok = filelib:ensure_dir(DestFilename),
+   CompressedFile = DestFilename ++ ".gz",
+   Z = zlib:open(),
+   ok = zlib:deflateInit(Z, best_compression, deflated, 16 + 15, 8, default),
+   {ok, Out} = file:open(CompressedFile, [raw, write, binary, append]),
```
Vagabond, line 2024: I don't think you need/want append here?

src/blockchain.erl
```diff
@@ -1997,6 +2000,59 @@ save_bin_snapshot(DestFilename, BinSnap) when is_binary(BinSnap); is_list(BinSna
...
+            (Data) ->
+               CData = zlib:deflate(Z, Data),
+               file:write(Out, CData)
+         end,
+   RetVal = blockchain_utils:streaming_transform_iolist(BinSnap, Fun),
+   file:close(Out),
+   RetVal.
+
+do_compress(In, Out, Pos, Z) ->
+   case file:pread(In, Pos, ?BLOCK_READ_SIZE) of
```
Vagabond, line 2038: If you are just reading a file sequentially you can just keep calling read(), the position gets advanced automatically. pread will do a seek and then a read.

src/blockchain_utils.erl
```diff
@@ -651,6 +652,18 @@ do_hash(FD, Pos, HashState) ->
         {error, _E} = Err -> Err
     end.
 
+streaming_transform_iolist(L, Fun) when is_list(L) and length(L) =< 32000 ->
```
Vagabond, line 655: According to zlib:deflate() it can take `iodata` so you probably don't need to iolist_to_binary anywhere in these functions?

jadeallenx: ah missed that

PR #1258, opened 2022-02-26
Correct expected fee to align with new size of txn

joecaswell:


Vagabond (approved):

Vagabond:
Longer term it'd be good to calculate the fee dynamically, but this will do for now.

PR #1259, opened 2022-02-28
add a tuning knob for h3dex gc res

evanmcc:
This allows the gc area to be increased so we then have two tuning knobs:  the number of requests we'd like to consider, and the area we'd like each of them to cover.


andymck (approved):
lgtm

Vagabond (approved):

PR #1260, opened 2022-03-01
add fixedwidth bitvector encoding and var to control its use in miner

evanmcc:
this code initially assumed that all votes would be present, but we failed to enforce that, so we need to provide a new version that can only be used safely.   while we don't use this function in core, a var is required because otherwise we will produce halting blocks during the upgrade process.


evanmcc:
closing in favor of #1263

PR #1264, opened 2022-03-01
fix snapshot info for v6 snapshot

joecaswell:


src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -1256,6 +1257,21 @@ kv_stream_to_list(Next0) when is_function(Next0) ->
         {K, V, Next1} -> [{K, V} | kv_stream_to_list(Next1)]
     end.
 
+-spec blocks_info(snapshot()) -> {ok, {non_neg_integer(), non_neg_integer(), non_neg_integer()}}.
```
xandkar, line 1260: Great idea abstracting this! It maybe even better to stick it into a map instead of the triple - I begin to doubt my memory of what is what beyond doubles.

Vagabond: I think 3 is ok, but certainly would be better to use a map beyond that.

Vagabond (approved):

PR #1266, opened 2022-03-03
Adt/add snapshot grab peer option

Vagabond:


src/cli/blockchain_cli_snapshot.erl
```diff
@@ -123,29 +123,37 @@ snapshot_load(Filename) ->
...
-        %% NOTE: grab_snapshot returns a deserialized snapshot
-        file:write_file(Filename, blockchain_ledger_snapshot_v1:serialize(Snapshot))
+        Res = case proplists:get_value(peer, Args, undefined) of
+            undefined -> blockchain_worker:grab_snapshot(Height, Hash);
+            Peer -> blockchain_worker:grab_snapshot(Height, Hash, Peer)
+        end,
+        case Res of
+            {ok, Snapshot} ->
+                %% NOTE: grab_snapshot returns a deserialized snapshot
+                file:write_file(Filename, blockchain_ledger_snapshot_v1:serialize(Snapshot));
```
jadeallenx, line 147: there are functions in blockchain to write snapshots now (compressed and uncompressed...)

jadeallenx (approved):

src/cli/blockchain_cli_snapshot.erl
joecaswell: switched to save_bin_snapshot

src/blockchain_worker.erl
```diff
@@ -1022,40 +1023,44 @@ start_block_sync(SwarmTID, Chain, Peer, Heights, GossipedHash) ->
...
+                cancel ->
+                    lager:info("snapshot sync cancelled"),
+                    _ = libp2p_framed_stream:close(Stream),
+                    {error, snap_sync_cancel};
+                {'DOWN', Ref1, process, Stream, normal} ->
+                    {error, down};
+                {'DOWN', Ref1, process, Stream, Reason} ->
+                    lager:info("snapshot sync failed with error ~p", [Reason]),
+                    {error, down, Reason}
+            after timer:minutes(15) ->
```
jadeallenx, line 1059: Huh, TIL :+1:

src/cli/blockchain_cli_snapshot.erl
```diff
@@ -123,29 +123,39 @@ snapshot_load(Filename) ->
...
-        %% NOTE: grab_snapshot returns a deserialized snapshot
-        file:write_file(Filename, blockchain_ledger_snapshot_v1:serialize(Snapshot))
+        Res = case proplists:get_value(peer, Args, undefined) of
+            undefined -> blockchain_worker:grab_snapshot(Height, Hash);
+            Peer -> blockchain_worker:grab_snapshot(Height, Hash, Peer)
+        end,
+        case Res of
+            {ok, Snapshot} ->
+                %% NOTE: grab_snapshot returns a deserialized snapshot
+                ok = file:write_file(Filename, blockchain_ledger_snapshot_v1:serialize(Snapshot)),
```
jadeallenx, line 148: This should _probaby_ be the same `save_snapshot` call in `blockchain` - also tiny lil preference to save the snapshot as compressed but... meh

src/cli/blockchain_cli_snapshot.erl
```diff
@@ -165,8 +175,10 @@ snapshot_diff(_, _, _) ->
     usage.
 
 snapshot_diff(AFilename, BFilename) ->
-    {ok, A} = blockchain_ledger_snapshot_v1:deserialize({file, AFilename}),
-    {ok, B} = blockchain_ledger_snapshot_v1:deserialize({file, BFilename}),
+    {ok, ABin} = file:read_file(AFilename),
```
jadeallenx, line 178: this will require the entire snapshot to be in memory right? Unclear if we need/want snapshot diffs to be "streaming" the same way that loading them would be...

jadeallenx (commented):
Overall looks good - would appreciate a look at the couple comments

src/cli/blockchain_cli_snapshot.erl
joecaswell: Yes, it does.  Comment on blockchain_ledger_snapshot_v1:deserialize suggests that snapshots < version 6 must be read from disk all at once to be converted to erl terms.  Assuming the keys in the kv_stream() are always in the same order, we may be able to pre-test for version and use a streaming diff if both files are v6, otherwise they will need to be read into memory first, but that seems like a lot of work to save a few megabytes of temporary memory usage.

src/cli/blockchain_cli_snapshot.erl
joecaswell: Done.

jadeallenx (approved):

PR #1267, opened 2022-03-03
Madninja/base64url snapshot

madninja:
Allows binary and base64url hashes in sys.config for `assumed_valid_block_hash` and `blessed_snapshot_block_hash`


Vagabond:
Would be good to do snapshot grab as well, if #1266 lands

PR #1272, opened 2022-03-04
Andymck/poc grpc v2 plus vals as chall

andymck:
Core side of https://github.com/helium/miner/pull/1469

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -152,6 +152,10 @@
 %%       and so this value must be sufficient as to give time for absorb to occur
 -define(poc_receipts_absorb_timeout, poc_receipts_absorb_timeout).
 
+%% Number of blocks for which a validator ephemeral key will remain valid : pos_integer
+%% after this period it will be GCed
+%% NOTE the minimum value here should be greater than that of validator_liveness_interval
+-define(poc_validator_ephemeral_key_timeout, poc_validator_ephemeral_key_timeout).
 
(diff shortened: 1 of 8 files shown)


include/blockchain_vars.hrl
```diff
@@ -152,6 +152,13 @@
 %%       and so this value must be sufficient as to give time for absorb to occur
 -define(poc_receipts_absorb_timeout, poc_receipts_absorb_timeout).
 
+%% Number of POC ephemeral keys to be generated and included in a validator heartbeat: pos_integer
+-define(poc_validator_ephemeral_key_count, poc_validator_ephemeral_key_count).
```
Vagabond, line 156: is it possible to calculate this from the # of active validators and the "per block poc rate" instead?

Vagabond: I'm worried about having to manually change this value if the # of active validators changes

andymck: yeah that makes for a better approach

andymck: fixed here: 
https://github.com/helium/blockchain-core/pull/1272/commits/67480e4307e3e1bfaca1fb8a64a847639ce0d9ea
https://github.com/helium/miner/pull/1469/commits/cf381c6c4b2878d7d6747660ea49dc3036605853
https://github.com/helium/sibyl/pull/23/commits/f1a49b2908efa4aa217165e0aebd6d31ab7ad9b2

PR #1277, opened 2022-03-07
Work-around that hackney doesn't send a message on timeouts when async

Vagabond:


jadeallenx (approved):
Would prefer as a macro or adding a comment

PR #1280, opened 2022-03-11
Jc/snapshot grab stream file

joecaswell:
Stream snapshot data when server from a file


src/handlers/blockchain_snapshot_handler.erl
```diff
@@ -121,3 +121,40 @@ handle_data(server, Data, #state{chain = Chain} = State) ->
 handle_info(_Type, _Msg, State) ->
     lager:info("unhandled message ~p ~p", [_Type, _Msg]),
     {noreply, State}.
+
+%% ------------------------------------------------------------------
+%% internal functions
+%% ------------------------------------------------------------------
+
+% convert integer to varint encoding for protocol buffer
```
Vagabond, line 129: We have a varint library in the dep tree, I think its called `small_ints` or something

src/handlers/blockchain_snapshot_handler.erl
```diff
@@ -121,3 +121,40 @@ handle_data(server, Data, #state{chain = Chain} = State) ->
...
+            fun() ->
+                    {mk_file_stream_fun(File), Data}
+            end;
+        eof ->
+            file:close(File),
+            fun() -> ok end
+    end;
+mk_file_stream_fun(FileName) ->
+    {ok, #file_info{size = Bytes}} = file:read_file_info(FileName),
+    {ok, File} = file:open(FileName, [read, binary]),
```
Vagabond, line 149: Would suggest `raw` mode here for performance reasons

src/handlers/blockchain_snapshot_handler.erl
joecaswell: TIL

src/handlers/blockchain_snapshot_handler.erl
joecaswell: Done

PR #1281, opened 2022-03-15
Jc/sync final flag

### LINKED ISSUE #1273: When sending final block sync bundle, indicate if the last batch of blocks is 'final'
Currently the block sender does not indicate if the final batch of blocks sent is all the block the far end has. Adding a flag to the protocol and having the client check for that to inform further sync behaviour would be useful. Currently the node waits for a new block hash gossip to try to sync again, which may take several minutes.

joecaswell:
Add a flag to indicate the final batch of blocks in a sync, client skips eager gossip for non-final batch.  Fixes #1273 Depends on helium/proto#125


joecaswell:
This has commits from 2 different PRs, need to rebase

PR #1282, opened 2022-03-15
add tag for the final batch of blocks in a sync

### LINKED ISSUE #1273: When sending final block sync bundle, indicate if the last batch of blocks is 'final'

joecaswell:
deduplicate block send code
fixes #1273 depends on helium/proto#125

### DIFF
--- a/src/handlers/blockchain_sync_handler.erl
+++ b/src/handlers/blockchain_sync_handler.erl
@@ -125,7 +125,7 @@ handle_data(client, Data0, #state{blockchain=Chain, path=Path, gossiped_hash=Gos
             ?SYNC_PROTOCOL_V1 -> Data0;
             ?SYNC_PROTOCOL_V2 -> zlib:uncompress(Data0)
         end,
-    #blockchain_sync_blocks_pb{blocks=BinBlocks} =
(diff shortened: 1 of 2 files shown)


src/handlers/blockchain_sync_handler.erl
```diff
@@ -139,8 +139,12 @@ handle_data(client, Data0, #state{blockchain=Chain, path=Path, gossiped_hash=Gos
             %% nothing was plausible, see if it has anything else
             {noreply, State, blockchain_sync_handler_pb:encode_msg(#blockchain_sync_req_pb{msg={response, true}})};
         HighestPlausible ->
-            lager:info("Eagerly re-gossiping ~p", [blockchain_block:height(HighestPlausible)]),
-            blockchain_gossip_handler:regossip_block(HighestPlausible, SwarmTID),
+            case Final of
```
Vagabond, line 142: So one thing we'd know now is that if we see the Final flag but we haven't seen the block with `GossipedHash` we know that this peer has more blocks (or that more blocks exist generally) and we can keep syncing immediately?

PR #1283, opened 2022-03-15
validate file size and hash after downloading

### LINKED ISSUE #1276: Snapshot download does not verify file size or hash
The JSON contains the hash and the size of the file to download, but it's currently not checked.

The result of `blockchain_worker:do_snap_source_download` should be checked before attempting to load it.

joecaswell:
fixes #1276


src/blockchain_worker.erl
```diff
@@ -1240,17 +1242,24 @@ attempt_fetch_snap_source_snapshot(BaseUrl, #snapshot_info{height = Height, file
...
             end;
+        %% no file? no problem.
+        false -> ok
+    end,
+    case Retry of
+        true ->
+            %% on the first check, try to download or resume the file
+            _ = do_snap_source_download(build_url(BaseUrl, Filename), Filepath),
+            %% we just downloaded, recurse to check size and hash of the new download
+            attempt_fetch_snap_source_snapshot(BaseUrl, SnapInfo#snapshot_info{retry_failure = false});
```
jeffgrunewald, line 1259: i think i'm missing something here; when will this ever succeed if it always retries once on `Retry = true` and then recurses, setting `Retry = false` at which point it throws an error?

jeffgrunewald: it looks to me like you need to be nesting this retry case statement in the main one so you return that `{ok, FilePath}` on the occasions where the file `is_regular` andalso the hashes match

joecaswell: After do_snap_source_download returns, the new snapshot file will stored, so on the retry, it should successfully pass the size and hash check to hit the `{true, true}` branch at https://github.com/helium/blockchain-core/blob/0c1503fe3f63dbf0f645f11da8b92528dbc3ab46/src/blockchain_worker.erl#L1241
If it gets back to this `Retry` case on the second pass, that would mean that the downloaded file was not intact, so this function should throw to break out of the spawn_monitor started here: https://github.com/helium/blockchain-core/blob/0c1503fe3f63dbf0f645f11da8b92528dbc3ab46/src/blockchain_worker.erl#L1241
which will land in the retry attempt at https://github.com/helium/blockchain-core/blob/0c1503fe3f63dbf0f645f11da8b92528dbc3ab46/src/blockchain_worker.erl#L1241

jeffgrunewald: Right, but don’t you need to bind the result of that case statement (the first one beginning on line 1232) as the return value of the function since the blockchain_worker in the spawn_monitor is expecting that to return the ok tuple? Right now it’s not binding the result of that case to anything and then it’s going immediately into the case statement that evaluates the Retry variable which is outside of and at the same level as the first case clause so that should be the last clause of the function and what it ultimately returns, no?

src/blockchain_worker.erl
```diff
@@ -1204,7 +1204,16 @@ set_filename(BaseFilename, true) ->
 set_filename(BaseFilename, false) ->
     BaseFilename.
 
-attempt_fetch_snap_source_snapshot(BaseUrl, #snapshot_info{height = Height, file_hash = Hash,
+attempt_fetch_snap_source_snapshot(BaseUrl, SnapInfo) ->
+    case validate_snapshot_file(SnapInfo) of
+        {ok, Filename} ->
+            {ok, Filename}
+        {invalid, Filename, Filepath ->
```
jeffgrunewald, line 1211:
```suggestion
        {invalid, Filename, Filepath} ->
```

src/blockchain_worker.erl
```diff
@@ -1204,7 +1204,16 @@ set_filename(BaseFilename, true) ->
...
+            _ = do_snap_source_download(build_url(BaseUrl, Filename), Filepath),
+            validate_snapshot_file(SnapInfo)
```
jeffgrunewald, line 1213: this makes sense but i think you need to either do a short recursive loop here and countdown however many download attempts we want to allow or just nest another case statement here and turn a second `{invalid, Filename, Filepath}` into a `throw({error, something_bad})`.

in the main spawn_monitor function where `attempt_fetch_snap_source_snapshot/2` is called this is wrapped inside a `try/catch` and all of the old code paths always resulted in the download being either an `{ok, Filepath}` or `throw` ing some error related to the download. In this case, if something went really bad with retrieving the snap from S3 and we failed on both attempts to validate the snap file the second attempt would return the `invalid` triple which would result in a match exception on 1073

jeffgrunewald, line 1213:
```suggestion
            case validate_snapshot_file(SnapInfo) of
                {ok, Filename} -> {ok, Filename};
                {invalid, _Filepath, _Filename} -> throw({error, invalid_snap_retrieved})
            end
```
something like this seems inelegant but should work assuming we want to limit the number of attempts to validate the snap file to just two

joecaswell: The download attempts are limited to 3 and counted in the blockchain_worker gen_server state.  This function represents a single download attempt.

The 3-tuple return will cause a bad match on `{ok, Filename} = attempt_fetch_snap_source_snapshot(BaseUrl, NewSnapInfo)`.  We could make that more explicit here with `{ok, _} = validate_snapshot_file(SnapInfo)`, or do you think the additional case statement would be better?

jeffgrunewald: i guess the bad match operation will still work here with the `catch` clause, it'll just wrap the `{invalid, Filepath, Filename}` error in an additional `{badmatch, {invalid, Filepath, Filename}}` which should still be pretty clear what's going wrong and where. i still get tripped up by erlang `throw/1` since Elixir has a distinction between control flow with `try/catch` and failure exceptions with `raise/rescue` but under the hood those exceptions are just a special type of `throw`

all that to say, using an explicit `throw/1` when the download is unrecoverable would make the `lager:error` on 1087 and the `exit(Error)` on 1088 _a little_ more to the point but i don't know how useful that actually is.

joecaswell: If I'm reading that right, the sole purpose of the try/catch is to make sure we log _something_ before going back for a retry on 637 or note the exceeded limit on 645

src/blockchain_worker.erl
```diff
@@ -1240,17 +1249,15 @@ attempt_fetch_snap_source_snapshot(BaseUrl, #snapshot_info{height = Height, file
                             lager:info("Already have snapshot file for height ~p with hash ~p", [Height, Hash]),
                             ok = file:write_file(HashStateFile, Hash),
                             {ok, Filepath};
-                        {smaller, _} ->
-                            %% see if this is a a resumable download
-                            _ = do_snap_source_download(build_url(BaseUrl, Filename), Filepath);
                         _ ->
-                            %% file is bigger than it should be, or the hash is wrong, scrap it
+                            %% file size or the hash is wrong, scrap it, 
+                            %% we'll try to resume from the scratch file, if there is one
```
Vagabond, line 1254: I'm not sure this doesn't break download resume?

src/blockchain_worker.erl
```diff
@@ -1204,7 +1204,16 @@ set_filename(BaseFilename, true) ->
...
     BaseFilename.
 
-attempt_fetch_snap_source_snapshot(BaseUrl, #snapshot_info{height = Height, file_hash = Hash,
+attempt_fetch_snap_source_snapshot(BaseUrl, SnapInfo) ->
+    case validate_snapshot_file(SnapInfo) of
+        {ok, Filename} ->
+            {ok, Filename}
+        {invalid, Filename, Filepath} ->
+            _ = do_snap_source_download(build_url(BaseUrl, Filename), Filepath),
+            {ok, _} = validate_snapshot_file(SnapInfo)
```
Vagabond, line 1213: Arguably you can just return the result of `validate_snapshot_file` here and not cause a badmatch?

joecaswell: Works either way, the return value is immediately matched with {ok, Filename} so it would badmatch either way.

src/blockchain_worker.erl
joecaswell: The resume login in do_snap_source_download only considers the .scratch file when checking for a partial download, is there another place to check for issues?

src/blockchain_worker.erl
```diff
@@ -1206,7 +1206,16 @@ set_filename(BaseFilename, true) ->
 set_filename(BaseFilename, false) ->
     BaseFilename.
 
-attempt_fetch_snap_source_snapshot(BaseUrl, #snapshot_info{height = Height, file_hash = Hash,
+attempt_fetch_snap_source_snapshot(BaseUrl, SnapInfo) ->
+    case validate_snapshot_file(SnapInfo) of
+        {ok, Filename} ->
+            {ok, Filename}
```
vihu, line 1212:
```suggestion
            {ok, Filename};
```

PR #1285, opened 2022-03-16
HIP-46: facilitate routing by netid-to-oui

dpezely:
This facilitates functionality that will be required if [HIP-46](https://github.com/helium/HIP/blob/main/0046-lorawan-netid-routing.md) passes:

- Create new chain var, `routers_by_netid_to_oui` that contains sorted list of tuples
- Each tuple associates a NetID to OUI for a roaming partner
- New functionality is primarily contained within State Channel client code


dpezely:
@evanmcc Here's some context, since you've been added as reviewer.  Tests pass for others yet fail on my local system.  Not sure if it's because they're running macos and I'm on Linux.  

e.g., `full_test()` fails for me in `init_per_testcase()` when running via:
```bash
SC_DIR=/tmp/sc/ ./rebar3 ct --suite test/blockchain_state_channel_SUITE --group sc_libp2p --case test_full
```
Once we can get beyond that, there are a few blocks of commented-out code (marked with FIXME) in that same suite, `netid_to_oui_test()` that needs to be finished.  It started life as a copy of `full_test()`.

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -196,7 +220,13 @@ handle_info(post_init, #state{chain=undefined}=State) ->
...
             {noreply, State};
         Chain ->
-            {noreply, State#state{chain=Chain}}
+            Ledger = blockchain:ledger(Chain),
+            Routers =
+                case blockchain_ledger_v1:config(?routers_by_netid_to_oui, Ledger) of
+                    {ok, Bin} -> binary_to_term(Bin);
+                    _ -> []
+                end,
+            {noreply, State#state{chain=Chain, routers=Routers}}
```
evanmcc, line 229: caching this in the state is fine, but it means that you need to scan incoming blocks (see the add_block event) for var changes and update your local state when that changes.

dpezely: Addressed in commit 0132da3

dpezely:
Within `blockchain_simple_SUITE.erl`, two existing tests are failing for reasons that seem unrelated to the new code: `reload_test()` and `restart_test()`.  Since they're unrelated, I would have approved this PR if GitHub's UI would allow me.  Because I created the initial PR, someone else will have to approve.

test/blockchain_simple_SUITE.erl
```diff
@@ -1343,6 +1345,96 @@ routing_test(Config) ->
...
+    Region = 'US915',
+    Now = erlang:system_time(seconds),
+    %% FIXME: may need to strip "/p2p/1" prefix
+    DefaultRouters = ["/p2p/11w77YQLhgUt8HUJrMtntGGr97RyXmot1ofs5Ct2ELTmbFoYsQa",
+                      "/p2p/11afuQSrmk52mgxLu91AdtDXbJ9wmqWBUxC3hvjejoXkxEZfPvY"],
+
+    %% Send packet using devaddr outside any of our test netids.
+
+    %% FIXME: using Network Byte Order returns entire roaming list, and
+    %% without that crashes `find_routing_via_devaddr' with bad arg.
```
dpezely, line 1399: We can remove these `FIXME` comments.

```suggestion
```

test/blockchain_simple_SUITE.erl
```diff
@@ -1343,6 +1345,96 @@ routing_test(Config) ->
...
+    RoamingRouters =
+        case blockchain_ledger_v1:config(?routers_by_netid_to_oui, Ledger) of
+            {ok, Bin} -> binary_to_term(Bin);
+            _ -> []
+        end,
+    ?assertNotEqual([], RoamingRouters),
+    State = blockchain_state_channels_client:set_routers(RoamingRouters, Chain),
+    Region = 'US915',
+    Now = erlang:system_time(seconds),
+    %% FIXME: may need to strip "/p2p/1" prefix
```
dpezely, line 1392: We can remove these `FIXME` comments.

```suggestion
```

macpie: You could use `libp2p_crypto:generate_keys` (look at Router how it does it) and then use the related `p2p` function to create the p2p address so you don't have to hard code them

dpezely: I went with this after a bit of discovery and based upon our screen-sharing conversation a while back:
```Erlang
DefaultRouters =
        [libp2p_crypto:pubkey_bin_to_p2p(binary:list_to_bin(blockchain_ct_utils:randname(16)))
         || _X <- lists:seq(1, 3)],
```

dpezely:
For anything directly related to `blockchain-core`, my confidence in whether this is complete enough (or at least sufficient for this piece) is far less than with `router` repo.

With that caveat, I think this PR is ready for final review.

TODO: There still needs to be an official means for populating the new chain var, `routers_by_netid_to_oui`.

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -32,6 +32,7 @@
 -include("blockchain.hrl").
 -include("blockchain_rocks.hrl").
 -include("blockchain_vars.hrl").
+-include("../grpc/autogen/server/state_channel_pb.hrl").
```
Vagabond, line 35: I think there's a better include_lib or something you can do here

dpezely: Fixed in commit 656f5b5d with:
```Erlang
include_lib("grpc/autogen/server/state_channel_pb.hrl").
```

src/state_channel/blockchain_state_channels_client.erl
```diff
@@ -32,6 +32,7 @@
 -include("blockchain.hrl").
 -include("blockchain_rocks.hrl").
 -include("blockchain_vars.hrl").
+-include_lib("grpc/autogen/server/state_channel_pb.hrl").
```
Vagabond, line 35: but this comes from helium_proto, right? why isn't it something like `-include_lib("helium_proto/include/state_channel_pb.hrl").`?

jadeallenx (approved):

jadeallenx:
please squash before merge

dpezely:
Superseded by [PR 1412](https://github.com/helium/blockchain-core/pull/1412)

PR #1287, opened 2022-03-18
filter bad netids from sync targets

evanmcc:
somehow, some bad peers are getting into the peerbook.  we need to fix that issue, but for now at least don't try to sync with them.


jadeallenx (commented):
Will this affect seed nodes selecting peers to gossip with? Or do we simply need to assign seeds a network id so they only gossip with their own network?

jadeallenx (approved):

PR #1291, opened 2022-03-30
assign challenger rewards to validators

jeffgrunewald:
under validator-led challenges, the `poc_challengers` reward type needs to be tagged as being for validators in order to be correctly handled within the map:fold that assigns rewards to accounts. this will require a chain reset.

~~a messier alternative, but would be slightly more backward compatible could be to add a clause such that we attempt to lookup a "gateway" as a validator, failing that as a gateway, and failing _that_ return the accumulator and move on, but that seems more hacky and likely to leave a footgun waiting for us. it might look something like:~~

this is a terrible idea; please disregard

### DIFF
--- a/src/transactions/v2/blockchain_txn_rewards_v2.erl
+++ b/src/transactions/v2/blockchain_txn_rewards_v2.erl
@@ -946,7 +946,7 @@ normalize_challenger_rewards(ChallengerRewards, #{epoch_reward := EpochReward,
         fun(Challenger, Challenged, Acc) ->
             PercentofReward = (Challenged*100/TotalChallenged)/100,
             Amount = erlang:round(PercentofReward * ChallengersReward),
-            maps:put({gateway, poc_challengers, Challenger}, Amount, Acc)
+            maps:put({validator, poc_challengers, Challenger}, Amount, Acc)
         end,
         #{},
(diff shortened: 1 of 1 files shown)


src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -946,7 +946,7 @@ normalize_challenger_rewards(ChallengerRewards, #{epoch_reward := EpochReward,
         fun(Challenger, Challenged, Acc) ->
             PercentofReward = (Challenged*100/TotalChallenged)/100,
             Amount = erlang:round(PercentofReward * ChallengersReward),
-            maps:put({gateway, poc_challengers, Challenger}, Amount, Acc)
```
evanmcc, line 949: I'm not sure the right var here, but we need a chain var switch here.

andymck: Chain var is poc_challenger_type,  if value is validator then use validator tagging else use gateway

jeffgrunewald: Done

evanmcc (approved):

andymck (approved):

PR #1294, opened 2022-04-06
validate heartbeat key list length

evanmcc:
- [x] add validator count to ledger
- [x] move len calc to txn util fun
- [x] miner PR to use util fun
- [x] validate using len calc

### DIFF
--- a/src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
+++ b/src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
@@ -120,6 +120,7 @@ is_valid(Txn, Chain) ->
     TxnHeight = height(Txn),
     {ok, Height} = blockchain_ledger_v1:current_height(Ledger),
     Proposals = poc_key_proposals(Txn),
+    Reactivated = reactivated_gws(Txn),
     case is_valid_sig(Txn) of
         false ->
             {error, bad_signature};
@@ -160,6 +161,19 @@ is_valid(Txn, Chain) ->
                     true -> ok;
                     false ->
                         throw({bad_proposal_length, TargetLen, length(Proposals)})
+                end,
+                ReactivationLimit =
+                    case blockchain_ledger_v1:config(?poc_challenger_type, Ledger) of
+                        {ok, validator} ->
+                            {ok, RL} = blockchain_ledger_v1:config(?validator_hb_reactivation_limit, Ledger),
+                            RL;
+                        _ ->
+                            0
+                    end,
+                case length(Reactivated) =< ReactivationLimit of
+                    true -> ok;
+                    false ->
+                        throw({bad_reactvation_length, ReactivationLimit, length(Reactivated)})
                 end,
                 ok
             catch throw:Cause ->
--- a/src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
+++ b/src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
@@ -173,7 +173,7 @@ is_valid(Txn, Chain) ->
                 case length(Reactivated) =< ReactivationLimit of
                     true -> ok;
                     false ->
-                        throw({bad_reactvation_length, ReactivationLimit, length(Reactivated)})
+                        throw({bad_reactivation_length, ReactivationLimit, length(Reactivated)})
                 end,
                 ok
             catch throw:Cause ->
--- a/src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
+++ b/src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
@@ -150,26 +150,20 @@ is_valid(Txn, Chain) ->
                     true -> ok;
                     false -> throw({bad_version, Version})
                 end,
-                TargetLen =
+                {TargetLen, ReactivationLimit} =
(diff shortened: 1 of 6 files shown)


src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
```diff
@@ -243,6 +258,16 @@ reactivate_gws(GWAddrs, Height, Ledger) ->
             end
         end, GWAddrs).
 
+proposal_length(Ledger) ->
+    case blockchain_ledger_v1:validator_count(Ledger) of
+        {ok, NumVals} when NumVals > 0 ->
+            {ok, ChallengeRate} = blockchain_ledger_v1:config(?poc_challenge_rate, Ledger),
+            {ok, HBInterval} = blockchain_ledger_v1:config(?validator_liveness_interval, Ledger),
+            round((ChallengeRate / (NumVals * 0.8)) * HBInterval);
```
jeffgrunewald, line 266: where does 0.8 come from?

jeffgrunewald (commented):
so we're going to "wipe" the validator count on each chain var hook and recalculate the number from the staked validators in the ledger and in between these we'll increment and decrement individually as validators join or leave the staking pool?

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
jeffgrunewald: i see the comment in the miner PR; very least i'd port that comment over here but feels like it should be a config or chain var value

andymck: yeah should pull over the comment here...gives some context on the 0.8

andymck: yeah should pull over the comment here...gives some context on the 0.8

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
```diff
@@ -243,6 +258,16 @@ reactivate_gws(GWAddrs, Height, Ledger) ->
             end
         end, GWAddrs).
 
+proposal_length(Ledger) ->
```
andymck, line 261: maybe add specs?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -717,7 +717,23 @@ var_hook(?poc_hexing_type, hex_h3dex, Ledger) ->
 %% we want to clear out the pocs CF
 %% we dont care about its value, if its been
 %% updated then we wipe all POCs
-var_hook(?poc_challenger_type, _, Ledger) ->
+var_hook(?poc_challenger_type, Type, Ledger) ->
+    case Type of
+        validator ->
```
andymck, line 722: why redo val counts upon var change here?

andymck, line 722: why redo val counts upon var change here?

evanmcc: they start to matter here, they don't currently exist in the ledger, so upon activation, we get a solid count. thereafter the activation and deactivation code keeps the number up to date.

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
evanmcc: I'll grab the comment and make the factor a var

src/transactions/v1/blockchain_txn_vars_v1.erl
jeffgrunewald: I would also see this drifting over time, like there were crashes mid-way through incrementing or decrementing the count so this could be a sort of periodic “audit” of the count

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
```diff
@@ -146,6 +150,31 @@ is_valid(Txn, Chain) ->
...
+                        {ok, validator} ->
+                            {ok, RL} = blockchain_ledger_v1:config(?validator_hb_reactivation_limit, Ledger),
+                            RL;
+                        _ ->
+                            0
+                    end,
+                case length(Reactivated) =< ReactivationLimit of
+                    true -> ok;
+                    false ->
+                        throw({bad_reactvation_length, ReactivationLimit, length(Reactivated)})
```
jeffgrunewald, line 176:
```suggestion
                        throw({bad_reactivation_length, ReactivationLimit, length(Reactivated)})
```

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
```diff
@@ -146,6 +150,31 @@ is_valid(Txn, Chain) ->
                     true -> ok;
                     false -> throw({bad_version, Version})
                 end,
+                TargetLen =
+                    case blockchain_ledger_v1:config(?poc_challenger_type, Ledger) of
```
jeffgrunewald, line 154: rather than checking twice in a row if the `?poc_challenger_type` is validator or not does it make more sense to just have a  single
```
case blockchain_ledger_v1:config(?poc_challenger_type, Ledger) of
    {ok, validator} ->
        all of the checks and reactivation validation
    _ -> ok
end
```

PR #1296, opened 2022-04-06
WIP: Cleanup and rework poc proposals

vihu:
- Remove poc_keys from block
- Update proto for removing keys from block
- Store poc proposal directly in the ledger on heartbeat txn
- Remove unnecessary cruft and address some review cmts
~~- Add poc_proposal_processing_count chain var~~
- publishes blockchain event containing the selected poc keys per block, to be consumed by miner and sibyl

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -2542,6 +2542,8 @@ save_block(Block, Batch, #blockchain{default=DefaultCF, blocks=BlocksCF, heights
     ok = rocksdb:batch_put(Batch, HeightsCF, <<Height:64/integer-unsigned-big>>, Hash),
     Info = mk_block_info(Hash, Block),
     Ledger = ?MODULE:ledger(Chain),
+    %% TODO: confirm if this is the best place for generating the key proposals
+    %% and sending the associated event
     BlockPOCs = blockchain_ledger_v1:process_poc_proposals(Height, Hash, Ledger),
     ok = blockchain_worker:notify({poc_keys, Height, Hash, false, BlockPOCs}),
     ok = rocksdb:batch_put(Batch, InfoCF, <<Height:64/integer-unsigned-big>>, term_to_binary(Info)).
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -2545,7 +2545,7 @@ save_block(Block, Batch, #blockchain{default=DefaultCF, blocks=BlocksCF, heights
     %% TODO: confirm if this is the best place for generating the key proposals
     %% and sending the associated event
     BlockPOCs = blockchain_ledger_v1:process_poc_proposals(Height, Hash, Ledger),
-    ok = blockchain_worker:notify({poc_keys, Height, Hash, false, BlockPOCs}),
+    ok = blockchain_worker:notify({poc_keys, {Height, Hash, false, BlockPOCs}}),
     ok = rocksdb:batch_put(Batch, InfoCF, <<Height:64/integer-unsigned-big>>, term_to_binary(Info)).
 
 save_temp_block(Block, #blockchain{db=DB, temp_blocks=TempBlocks, default=DefaultCF}=Chain) ->
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -2531,7 +2531,7 @@ save_block(Block, Chain = #blockchain{db=DB}) ->
 
 -spec save_block(blockchain_block:block(), rocksdb:batch_handle(), blockchain()) -> ok.
 save_block(Block, Batch, #blockchain{default=DefaultCF, blocks=BlocksCF, heights=HeightsCF,
-                                     info=InfoCF}) ->
+                                     info=InfoCF} = Chain) ->
     Height = blockchain_block:height(Block),
     Hash = blockchain_block:hash_block(Block),
     ok = rocksdb:batch_put(Batch, BlocksCF, Hash, blockchain_block:serialize(Block)),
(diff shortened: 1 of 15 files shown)


src/blockchain.erl
```diff
@@ -796,43 +796,23 @@ get_block_info(Height, Chain = #blockchain{db=DB, info=InfoCF}) ->
...
-        case PoCRequests of
-            M when map_size(M) == 0 ->
-                PoCKeys;
-            _ ->
-                PoCRequests
-        end,
+    PoCs = lists:flatmap(
+             fun(Txn) ->
+                     case blockchain_txn:type(Txn) of
+                         blockchain_txn_poc_request_v1 ->
```
jeffgrunewald, line 804: after the chain var switch over to validator-led PoC this will _always_ be an empty list since we no longer use the `blockchain_txn_poc_request_v1` (or any sort of txn_poc_request as i understand it) which is why the comment on the previous version of the code. is it worth keeping this field around as an empty map in the block info?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2080,33 +2081,59 @@ delete_poc(OnionKeyHash, Challenger, Ledger) ->
...
+                {ok, K} ->
+                    RandState = blockchain_util:rand_state(BlockHash),
+                    {_, POCSubset} = blockchain_utils:deterministic_subset(K, RandState, ProposedPOCs),
+                    lists:foreach(fun({_Proposal, POC}) ->
+                                          %% TODO: Do something with the proposal...?
+
+                                          ActivePOC = blockchain_ledger_poc_v3:status(POC, active),
+                                          OnionKeyHash = blockchain_ledger_poc_v3:onion_key_hash(ActivePOC),
+                                          Challenger = blockchain_ledger_poc_v3:challenger(ActivePOC),
+                                          save_public_poc_(OnionKeyHash, Challenger, BlockHash, BlockHeight, Ledger)
```
jeffgrunewald, line 2114:
```suggestion
                                          save_public_poc(OnionKeyHash, Challenger, BlockHash, BlockHeight, Ledger)
```

vihu: This is obsolete and fixed by @andymck in 033f7e2f5187f8f9984888476d3cbd5fe3758617

src/blockchain.erl
```diff
@@ -2561,6 +2541,11 @@ save_block(Block, Batch, #blockchain{default=DefaultCF, blocks=BlocksCF, heights
     ok = rocksdb:batch_put(Batch, HeightsCF, Hash, <<Height:64/integer-unsigned-big>>),
     ok = rocksdb:batch_put(Batch, HeightsCF, <<Height:64/integer-unsigned-big>>, Hash),
     Info = mk_block_info(Hash, Block),
+    Ledger = ?MODULE:ledger(Chain),
+    %% TODO: confirm if this is the best place for generating the key proposals
+    %% and sending the associated event
+    BlockPOCs = blockchain_ledger_v1:process_poc_proposals(Height, Hash, Ledger),
+    ok = blockchain_worker:notify({poc_keys, {Height, Hash, false, BlockPOCs}}),
```
jeffgrunewald, line 2548:
```suggestion
    ok = blockchain_worker:notify({poc_keys, {Height, Hash, BlockPOCs}}),
```
any way we can get rid of this `sync` boolean? i see both sibyl and miner in their respective PRs are ignoring it and from the call site here it just looks like a magic value but not sure if there's any intention to use it in future by another consumer

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -705,7 +705,7 @@ load_blocks(Ledger0, Chain, Snapshot) ->
               Ht = blockchain_block:height(Block),
               %% since hash and block are written at the same time, just getting the
               %% hash from the height is an acceptable presence check, and much cheaper
-              BlockHash =
+              _BlockHash =
```
jeffgrunewald, line 708: this and those commented out lines still needed?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2080,32 +2081,61 @@ delete_poc(OnionKeyHash, Challenger, Ledger) ->
...
+                    RandState = blockchain_utils:rand_state(BlockHash),
+                    {_, POCSubset0} = blockchain_utils:deterministic_subset(K, RandState, ProposedPOCs),
+                    POCSubset = lists:flatten(POCSubset0),
+                    L1 = ?MODULE:new_context(Ledger),
+                    lists:foreach(
+                        fun({_Key, POC}) ->
+                            ActivePOC0 = blockchain_ledger_poc_v3:status(active, POC),
+                            ActivePOC1 = blockchain_ledger_poc_v3:block_hash(BlockHash, ActivePOC0),
+                            ActivePOC2 = blockchain_ledger_poc_v3:start_height(BlockHeight, ActivePOC1),
+                            update_public_poc(ActivePOC2, L1)
```
jeffgrunewald, line 2115: i'm having a hard time tracing what happens in `update_public_poc/2` and whether or not this `cache_put` is a direct write or not but in theory what happens here if there's a crash some time when the node is running through this `lists:foreach` and its marked half of the proposal subset as `active` but hasn't yet committed the new ledger context? when it picks up again and re-runs `find_public_poc_proposals/1` again won't it get a different subset b/c some number of the previous group has already been marked `active` and `find_public_poc_proposals/1` only returns proposals in the `proposed` state?

src/blockchain.erl
andymck: whilst after the switchover this field will be empty, it does have to remain in order to support receipts v1s as it is checked during validation

src/ledger/v1/blockchain_ledger_snapshot_v1.erl
andymck: fixed

src/ledger/v1/blockchain_ledger_v1.erl
andymck: if it crashes before completing the foreach run, the context will never be committed and thus there will be no changes to the ledger..ie no public pocs will have been updated either ( either on original ledger rocks or its associated ETS cache )

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
```diff
@@ -177,12 +182,8 @@ absorb(Txn, Chain) ->
...
                     case blockchain:config(poc_challenger_type, Ledger) of
                         {ok, validator} ->
-                            case application:get_env(blockchain, poc_mgr_mod) of
-                                {ok, POCMgrMod} ->
-                                    POCMgrMod:save_poc_key_proposals(Validator, POCKeyProposals, TxnHeight);
-                                _ ->
-                                    ok
-                            end;
+                            %% POCKeyProposals = [{onion_key_hash, ledger_poc_v3}, {}...]
+                            blockchain_ledger_v1:save_public_poc_proposals(POCKeyProposals, Validator, LedgerHash, LedgerHt, Ledger);
```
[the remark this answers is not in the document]
andymck: possibly but moving to the save_public_poc_proposals/5 function would require passing in the Chain also ( in order to get the block hash ).  And then getting the ledger from the chain in that function....so its prob a lil better to just keep as is as we already have derived the ledger from the chain in the txn implementation

src/blockchain.erl
andymck: Will revisit this when settled on final trigger point for firing the event

PR #1303, opened 2022-04-15
Introduce var to change how xor filter update txn fees are calculated

macpie:


src/transactions/v1/blockchain_txn_routing_v1.erl
```diff
@@ -168,8 +168,50 @@ calculate_fee(Txn, Chain) ->
...
+                    %% Find out current size at index, get new size, calculate diff
+                    OUI = Txn0#blockchain_txn_routing_v1_pb.oui,
+                    OldFilter = 
+                        case blockchain_ledger_v1:find_routing(OUI, Ledger) of
+                            {ok, Routing} ->
+                                Filters = blockchain_ledger_routing_v1:filters(Routing),
+                                %% We do +1 here because Index is "0 indexed"
+                                %% and lists:nth/2 does not support 0
+                                lists:nth(Index+1, Filters);
+                            _Error ->
```
vihu, line 195: nit: Any use in handling `{error, not_found}` separately? Also, worth logging error here (not sure how noisy it would get though)

vihu (commented):
lgtm, one question is all

src/transactions/v1/blockchain_txn_routing_v1.erl
```diff
@@ -168,8 +168,50 @@ calculate_fee(Txn, Chain) ->
...
+                        end,
+                    SizeDiff = erlang:byte_size(OldFilter) - erlang:byte_size(Filter),
+                    %% If diff < 0, meaning that old filter is bigger than new one we set new filter to be empty
+                    %% If diff > 0, we calculate fees based on a random binary of same size
+                    case SizeDiff < 0 of
+                        true ->
+                            Txn2 = Txn1#blockchain_txn_routing_v1_pb{update= {update_xor, Index, <<>>}},
+                            ?calculate_fee(Txn2, Ledger, DCPayloadSize, TxnFeeMultiplier);
+                        false ->
+                            Txn2 = Txn1#blockchain_txn_routing_v1_pb{update= {update_xor, Index, crypto:strong_rand_bytes(SizeDiff)}},
```
evanmcc, line 206: do we need to use entropy here (calls a nif, takes a lock, uses up system entropy)? can we not just make a 0 binary of the appropriate size, like `<<0:SizeDiff/bytes>>` or something?

evanmcc: probably needs to be spelled `<<0:(8*SizeDiff)>>` actually

src/transactions/v1/blockchain_txn_routing_v1.erl
macpie: Error log could be ok here I think

src/transactions/v1/blockchain_txn_routing_v1.erl
macpie: Yes, you are right, just got the habit of using `crypto:strong_rand_bytes/1`, will fix.

include/blockchain_vars.hrl
```diff
@@ -392,6 +392,9 @@
 %%  - 1 :: accept first dispute, drop all DC from opener, no rewards
 -define(sc_dispute_strategy_version, sc_dispute_strategy_version).
 
+%% Txn Routing Xor Filter Fee calculation var HIP-XXX
```
andymck, line 395: nit: add permitted values here ie "not set or 1"

andymck (approved):
lgtm

src/transactions/v1/blockchain_txn_routing_v1.erl
```diff
@@ -168,8 +168,51 @@ calculate_fee(Txn, Chain) ->
...
+                                %% and lists:nth/2 does not support 0
+                                lists:nth(Index+1, Filters);
+                            _Error ->
+                                lager:error("we failed to get routing info for OUI=~p : ~p", [OUI, _Error]),
+                                <<>>
+                        end,
+                    SizeDiff = erlang:byte_size(OldFilter) - erlang:byte_size(Filter),
+                    %% If diff < 0, meaning that old filter is bigger than new one we set new filter to be empty
+                    %% If diff > 0, we calculate fees based on a random binary of same size
+                    case SizeDiff < 0 of
```
anthonyra, line 202: What should happen if the SizeDiff is equal to 0? I'd assume the same if < 0?

Vagabond: Yeah, this might be better as =< 0

macpie: Good catch fixing

src/transactions/v1/blockchain_txn_routing_v1.erl
```diff
@@ -584,4 +632,68 @@ to_json_test() ->
...
+
+            %% Testing with smaller xor
+            ?assertEqual(108, calculate_fee(Txn1, ledger, 1, 1, true)),
+
+            Xor2 = crypto:strong_rand_bytes(102),
+            Txn2 = blockchain_txn_routing_v1:update_xor(OUI, Owner, Index, Xor2, Nonce),
+
+            %% Testing with small update to show price diff with previous version
+            %% 112 compare to > 300
+            ?assertEqual(112, calculate_fee(Txn2, ledger, 1, 1, true)),
```
vihu, line 691:
```suggestion
            ?assertEqual(112, calculate_fee(Txn2, ledger, 1, 1, true)),
            
            Xor3 = crypto:strong_rand_bytes(102),
            Txn3 = blockchain_txn_routing_v1:update_xor(OUI, Owner, Index, Xor3, Nonce),

            %% Xor3 is same size as Xor2, no change in fee.
            ?assertEqual(112, calculate_fee(Txn3, ledger, 1, 1, true)),
```

vihu (approved):

evanmcc (approved):

PR #1308, opened 2022-04-18
Don't use context in a block absorb hook

Vagabond:


vihu (approved):

jeffgrunewald (approved):

andymck (approved):
Tested on testval1 which was halted on block 3544 due to hitting the "Not implemented: Not supported operation in read only mode." error

With this change in place the node is catching up, so lgtm

PR #1311, opened 2022-04-22
make each var do one thing

evanmcc:
in the previous version of this code we had some level of confusion as to what vars are actually activating hex targeting.  I think that separating this out like this improves the situation, as we can separately enable and disable hex targeting and incremental updated and etc.

### DIFF
--- a/src/transactions/v1/blockchain_txn_vars_v1.erl
+++ b/src/transactions/v1/blockchain_txn_vars_v1.erl
@@ -690,24 +690,10 @@ process_hooks(Vars, Unsets, Ledger) ->
 var_hook(?poc_targeting_version, 6, Ledger) ->
     %% purge any active POCs
     purge_pocs(Ledger),
-    %% build the h3dex lookup
-    {ok, Res} = blockchain_ledger_v1:config(?poc_target_hex_parent_res, Ledger),
-    blockchain_ledger_v1:build_random_hex_targeting_lookup(Res, Ledger),
     ok;
 var_hook(?poc_targeting_version, 5, Ledger) ->
     %% purge any active POCs
     purge_pocs(Ledger),
-    %% v3 targeting enabled, remove the h3dex lookup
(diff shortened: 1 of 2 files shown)


src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -690,24 +690,10 @@ process_hooks(Vars, Unsets, Ledger) ->
...
-    {ok, Res} = blockchain_ledger_v1:config(?poc_target_hex_parent_res, Ledger),
-    blockchain_ledger_v1:build_random_hex_targeting_lookup(Res, Ledger),
     ok;
 var_hook(?poc_targeting_version, 5, Ledger) ->
     %% purge any active POCs
     purge_pocs(Ledger),
-    %% v3 targeting enabled, remove the h3dex lookup
-    blockchain_ledger_v1:clean_random_hex_targeting_lookup(Ledger),
-    ok;
-var_hook(?poc_targeting_version, 4, Ledger) ->
```
andymck, line 703: i think we should keep support for version 3 and 4, and purge the POCs upon the value being changed to that, given we are testing moving from the version 3 and forward

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -720,6 +706,9 @@ var_hook(?poc_hexing_type, h3dex, Ledger) ->
     ok;
 var_hook(?poc_hexing_type, hex_h3dex, Ledger) ->
```
andymck, line 707: Maybe we should perform same actions for poc_hexing_type  with a value of hex_h3dex or h3dex.  hex_h3dex is an interim value which allows for both the old hexes and the new h3dexs to be maintained but ultimately this var will move to a value of h3dex.  Its probably fine if we move to that value *after* previously moving to a value of hex_h3dex but it does no harm to rebuilt the lookups in either scenario

PR #1312, opened 2022-04-25
verify poc key pair

andymck:
This verifies the presented secret ( a key pair ) as part of a receipt v2 txn are valid, ie the public key was derived from the private key.  It also checks the presented poc key hash was derived from the same public key.  This ties the secret to the active/public POC on the ledger and guards against spurious secrets being presented

### DIFF
--- a/src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
+++ b/src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
@@ -187,7 +187,6 @@ check_is_valid_poc(POCVersion, Txn, Chain) ->
     Ledger = blockchain:ledger(Chain),
     Challenger = ?MODULE:challenger(Txn),
     POCOnionKeyHash = ?MODULE:onion_key_hash(Txn),
-    BlockHash = ?MODULE:block_hash(Txn),
     POCID = ?MODULE:poc_id(Txn),
     StartPre = maybe_start_duration(),
     case blockchain_ledger_v1:find_public_poc(POCOnionKeyHash, Ledger) of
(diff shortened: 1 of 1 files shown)


src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
```diff
@@ -197,8 +197,20 @@ check_is_valid_poc(POCVersion, Txn, Chain) ->
...
             Secret = ?MODULE:secret(Txn),
-            case blockchain_ledger_poc_v3:verify(PoC, Challenger, BlockHash) of
+            Keys = libp2p_crypto:keys_from_bin(Secret),
+            #{public := PubKey, secret := PrivKey} = Keys,
+            OnionHash = crypto:hash(sha256, libp2p_crypto:pubkey_to_bin(PubKey)),
+            SigFun = libp2p_crypto:mk_sig_fun(PrivKey),
+            SignedPayload = SigFun(OnionHash),
+            case blockchain_ledger_poc_v3:verify(PoC, Challenger, BlockHash)
+                andalso POCOnionKeyHash == OnionHash
+                andalso libp2p_crypto:verify(OnionHash, SignedPayload, PubKey) of
```
jeffgrunewald, line 213: do we ever want to know why this is failing to verify if it's not since we catch all three possible validations in a single error tuple?

andymck: it might be nice but i dont think its a necessity at this point

vihu (approved):

evanmcc (approved):

PR #1313, opened 2022-04-25
h3dex test issues

andymck:
1. Avoid attempt to GC h3dex if on genesis block

2. Prevent infinite loop in scenario where we hit hex with no gateways, in theory this should not be possible but as we handle the scenario we should handle it correctly


src/poc/blockchain_poc_target_v6.erl
```diff
@@ -116,51 +153,82 @@ target_(
...
+                                Acc
+                        end;
+                    false ->
+                        %% If the GW is inactive then remove it from the h3dex
+                        %% so as to avoid having to iterate over it again later
+                        %% if it becomes active again it will be re-added
+                        %% via its durable validator heartbeat reactivation mechanism
+                        %% NOTE: the GW will eventually be GCed irrespective of what we do here
+                        %% but GCing it here keeps the h3dex more current
+                        _ = blockchain_ledger_v1:remove_gw_from_h3dex(Hex, A, Res, L1),
```
Vagabond, line 224: I don't think this is safe to do here because target() might be called in situations where its not safe to modify the ledger.

Vagabond: yeah but different machines might run target() at different times, so even using a context isn't safe

src/poc/blockchain_poc_target_v5.erl
```diff
@@ -219,4 +219,4 @@ is_active(_, _Gateway, _Height, _Vars, _Ledger) ->
 
 -spec max_activity_age(Vars :: map()) -> pos_integer().
 max_activity_age(Vars) ->
-    maps:get(poc_v4_target_challenge_age, Vars).
```
evanmcc, line 222: I think the ship has sailed on this change with 1.9.1.  we should probably use `target_challenge_age` for both?

evanmcc (commented):
I think that this is ready to merge, but I don't think that we can change the var we're using, since we've already shipped 1.9.1 and we don't want to go to 1.9.2 right now.  Ideally we make both use the same var, and then we can run the transition again.

src/poc/blockchain_poc_target_v5.erl
andymck: good point.  Settle on target_challenge_age then but lets set the value to the same as that of hip17_interactivity_blocks when enabling validator challenges

src/poc/blockchain_poc_target_v6.erl
```diff
@@ -116,51 +143,61 @@ target_(
...
+    Ledger :: blockchain_ledger_v1:ledger()
 ) -> [libp2p_crypto:pubkey_bin()].
-filter(AddrList, Ledger, Height, Vars) ->
+filter(AddrList, Height, Ledger) ->
     ActivityFilterEnabled =
         case blockchain:config(poc_activity_filter_enabled, Ledger) of
             {ok, V} -> V;
             _ -> false
         end,
+    {ok, MaxActivityAge} = blockchain:config(poc_v4_target_challenge_age, Ledger),
```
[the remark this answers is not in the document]
Vagabond: Oh, I see the discussion about this, disregard

PR #1317, opened 2022-04-27
move to pid-local ets

evanmcc:
Using a global name for this ets table opens up a race when a node is in the consensus group.  It's possible for the node to be creating the rewards and validating them at the same time if we're a little slow, which will cause one of the processes to crash, which can bring down the service unexpectedly.

This patch makes the perf recording entirely optional, and make it so that this race will not happen, even when it is enabled.


andymck (approved):
lgtm

andymck:
good find btw

PR #1319, opened 2022-04-28
sync_peer option to specify preferred nodes

joecaswell:
adds blockchain sync_peers configuration option, accepts a list of string multi-addresses

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -879,21 +879,29 @@ reset_ledger_to_snap(Hash, Height, State) ->
     snapshot_sync(State1#state{snapshot_info=SnapInfo}).
 
 start_sync(#state{blockchain = Chain, swarm_tid = SwarmTID} = State) ->
-    case get_random_peer(SwarmTID) of
+    Peer = get_configured_or_random_peer(SwarmTID),
+    case Peer of
         no_peers ->
             %% try again later when there's peers
             schedule_sync(State);
-        RandomPeer ->
-            {Pid, Ref} = start_block_sync(SwarmTID, Chain, RandomPeer, [], <<>>),
+        Peer ->
+            {Pid, Ref} = start_block_sync(SwarmTID, Chain, Peer, [], <<>>),
             lager:info("new block sync starting with Pid: ~p, Ref: ~p, Peer: ~p",
-                       [Pid, Ref, RandomPeer]),
+                       [Pid, Ref, Peer]),
             State#state{sync_pid = Pid, sync_ref = Ref}
+    end.
+
+get_configured_or_random_peer(SwarmTID) ->
+    case get_configured_sync_peer(SwarmTID) of
+        undefined-> get_random_peer(SwarmTID);
+        P -> P
     end.
 
 -spec get_random_peer(SwarmTID :: ets:tab()) -> no_peers | string().
 get_random_peer(SwarmTID) ->
+    lager:debug("Get random peer"),
     Peerbook = libp2p_swarm:peerbook(SwarmTID),
-    %% limit peers to random connections with public addresses
+    %% limit peers to connections with a public address
     F = fun(Peer) ->
                 case application:get_env(blockchain, testing, false) of
                     false ->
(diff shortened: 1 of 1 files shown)


src/blockchain_worker.erl
```diff
@@ -879,21 +879,29 @@ reset_ledger_to_snap(Hash, Height, State) ->
     snapshot_sync(State1#state{snapshot_info=SnapInfo}).
 
 start_sync(#state{blockchain = Chain, swarm_tid = SwarmTID} = State) ->
-    case get_random_peer(SwarmTID) of
+    Peer = get_configured_or_random_peer(SwarmTID),
+    case Peer of
```
jeffgrunewald, line 883:
```suggestion
    case get_configured_or_random_peer(SwarmTID) of
```

jeffgrunewald: not sure the reuse of `Peer` in the assignment, the `case`, and again in the case match clause below works here since erlang doesn't let you shadow variables.

src/blockchain_worker.erl
```diff
@@ -879,21 +879,29 @@ reset_ledger_to_snap(Hash, Height, State) ->
...
+            {Pid, Ref} = start_block_sync(SwarmTID, Chain, Peer, [], <<>>),
             lager:info("new block sync starting with Pid: ~p, Ref: ~p, Peer: ~p",
-                       [Pid, Ref, RandomPeer]),
+                       [Pid, Ref, Peer]),
             State#state{sync_pid = Pid, sync_ref = Ref}
     end.
 
+get_configured_or_random_peer(SwarmTID) ->
+    case get_configured_sync_peer(SwarmTID) of
+        undefined-> get_random_peer(SwarmTID);
```
jeffgrunewald, line 896:
```suggestion
        undefined -> get_random_peer(SwarmTID);
```

src/blockchain_worker.erl
joecaswell: Since the value would be the same, it would work, but I like the simplicity of this suggestion.

src/blockchain_worker.erl
joecaswell: Since `Peer` would have the same value, it would work, but I do like the simplicity of not setting it in the first place

src/blockchain_worker.erl
```diff
@@ -906,9 +913,35 @@ get_random_peer(SwarmTID) ->
...
+            libp2p_crypto:pubkey_bin_to_p2p(Addr)
+    end.
+
+-spec get_configured_sync_peer(ets:tab()) -> string() | undefined.
+get_configured_sync_peer(SwarmTID) ->
+    case application:get_env(blockchain, sync_peers, []) of
+        [] ->
+            lager:debug("No sync_peers configured"),
+            undefined;
+        ConfiguredPeers ->
```
jeffgrunewald, line 925:
```suggestion
        ConfiguredPeers when is_list(ConfiguredPeers) ->
```

src/blockchain_worker.erl
```diff
@@ -906,9 +913,35 @@ get_random_peer(SwarmTID) ->
...
+            Peerbook = libp2p_swarm:peerbook(SwarmTID),
+            {Left, Right} = lists:split(rand:uniform(length(ConfiguredPeers)), ConfiguredPeers),
+            get_configured_sync_peer(SwarmTID, Peerbook, Right ++ Left)
```
jeffgrunewald, line 928: is this "cutting the deck" of configured peers so to speak? just a cheaper version of shuffling the list?

joecaswell: Yes, exactly.  We only need one peer, the rest of the list is only used if the chosen one fails to connect.

jeffgrunewald: slick solution 👍

src/blockchain_worker.erl
joecaswell: As long as we're validating, we should check `[P | _]` when is_list(P)` to make sure it is a list of strings, also should gracefully fail back to random so the node will still try to sync if the config is pooched.

jeffgrunewald (approved):

PR #1322, opened 2022-04-29
removing old 'dialing' stream handling block

jeffgrunewald:
Because of the way we add the `dialing` status to a stream state (via the `add_stream/3`) and don't ever set the value of that state map directly, we never actually have a state of `dialing`. The only possible states are `undefined`, `{unverified, dialing}`, `{unverified, pid()}`, or `pid()`.

### DIFF
--- a/src/state_channel/blockchain_state_channels_client.erl
+++ b/src/state_channel/blockchain_state_channels_client.erl
@@ -53,7 +53,7 @@
 
 -type state() :: #state{}.
 -type stream_key() :: non_neg_integer() | string().
--type stream_val() :: undefined | dialing | {unverified, pid()} | pid().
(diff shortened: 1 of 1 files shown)


src/blockchain.erl
```diff
@@ -3349,7 +3349,7 @@ block_info_upgrade_test() ->
     ExpV2BlockInfo = #block_info_v2{height = 1,
                                     time = 1,
                                     hash = <<"blockhash">>,
-                                    pocs = [],
+                                    pocs = #{},
```
jeffgrunewald, line 3352: this was breaking an test and looks like we're creating it via `maps:from_list/1` so the spec likely changed without the test being updated

PR #1323, opened 2022-04-29
blockchain_aux_ledger:new/2 missed reading and populating proposed_pocs

marcsugiyama:
`blockchain_aux_ledger:new/2` did not expect a `ProposedPoCsCF` value which causes a bad match error decoding the return from blockchain_ledger_v1:open_db/5. Found by `blockchain_aux_ledger_SUITE`.


vihu (approved):
nice find!

PR #1334, opened 2022-05-10
Parallelize `build_hash_chain`

xandkar:
For a full chain resync, this reduces the `build_hash_chain` running time from ~65 minutes to ~11 minutes on my machine (16-core Ryzen 9 with 32 GB RAM).

The main idea is to
1. asynchronously: enumerate all K->V pairs in the blocks CF;
2. in-parallel:
    2.1 deserialize blocks;
    2.2 lookup parents;
    2.3 accumulate `{Parent, Child}` pairs;
3. in-serial:
    3.1 build graph from above-built pairs;
    3.2 walk backwards from the youngest given hash and trace its
        longest possible lineage up to the oldest given hash.

The last step, 3.2, is essentially what the previous implementation (that this one replaces) used to do, but this time all the expensive deserialization has already been done, in parallel.


src/data/data_stream.erl
```diff
@@ -0,0 +1,235 @@
...
+                            Y = F(X),
+                            SchedPid ! {SchedID, consumer_output, Y},
+                            Work();
+                        {SchedID, done} ->
+                            ok
+                    end
+                end,
+            Producer =
+                fun () ->
+                    ok = iter(fun (X) -> SchedPid ! {SchedID, producer_output, X} end, T)
```
xandkar, line 87: It hasn't (yet) caused a problem, but in theory this is bad: producer is pouring into the scheduler's queue as fast as possible, potentially faster than consumers can pull from it.

One solution is to have the scheduler call the producer whenever more work is asked for, but ... that can block the scheduler, starving consumers.

Another is to produce in (configurable size) batches.

PR #1335, opened 2022-05-10
make gc interval changes safe for replay post-poc-v3 activation

evanmcc:
see comments for more information

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -2371,7 +2371,7 @@ maybe_gc_pocs(_Chain, Ledger, validator) ->
         case blockchain:config(?poc_proposal_gc_window_check, Ledger) of
             {ok, true} ->
                 %% make any needed interval changes here
-                101; % <---- this one
+                application:get_env(blockchain, poc_gc_interval_size, 101); % <---- this one
 
             %% don't change this one!
             _ -> 101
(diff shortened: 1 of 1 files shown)


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2363,7 +2363,20 @@ maybe_gc_pocs(_Chain, Ledger, validator) ->
     %% to be absorbed
     {ok, CurHeight} = current_height(Ledger),
     %%  set GC point off by one so it doesnt align with so many other gc processes
-    case CurHeight rem 101 == 0 of
```
andymck, line 2366: i dont get why we would need to change this ?  Or are you putting this forward as an improvement that allows us to readily change the trigger for a GC scan ?

If so keep in mind the GC scan is run over two CFs, each with different deletion points

evanmcc: this is the heart of the bug.  if we change the interval of GC later (in code, since it's not a var) it potentially breaks replay.  This locks us to 101 in the period where we're active but the GC period selection fix is inactive, which makes these interval changes dangerous.

andymck: Right but the 101 interval has always been there and has not changed.  I am still not seeing the issue here...

evanmcc:
confirmed these changes in a conversation with @Vagabond

PR #1336, opened 2022-05-10
Blockchain absorb metrics

jeffgrunewald:
Add basic block production metrics via telemetry to export the absorption times and block height at each new block produced.


jadeallenx (approved):
Nice, we'll want to instrument rewards with more detail at some point

PR #1337, opened 2022-05-10
Identify connected nodes instead of forcing connections.

marcsugiyama:
Many tests randomly fail to run when `blockchain_state_channel_SUITE:init_per_testcase/2` fails. `init_per_testcase` tries to make sure the first two nodes set up by `blockchain_ct_utils:init_per_testcase/2`, the `?config(nodes, Config)` value, are connected to each other as many tests expect this. If the two nodes are not connected to each other, `blockchain_state_channel_SUITE:init_per_testcase` tries to force a connection. The set up code uses `ct_rpc:call(Node, blockchain_swarm, gossip_peers, [], 500)` to see if a node is connected to another node and `ct_rpc:call(Node, libp2p_swarm, connect, [Swarm, libp2p_crypto:pubkey_bin_to_p2p(AddrToConnectToo)], 500)` to force a connection. Even if `libp2p_swarm:connect` succeeds, it's not clear that the `gossip_peers` returns the new connection. As a result, the setup always fails if the first to nodes are not connected.

Rather than change how setup forces a connection, this PR inspects the list of nodes and finds two nodes that are connected to each other for the tests to use.

For better clarity, the router and gateway nodes are added explicitly to `Config`. Also, a new config is added for a connected pair of nodes.

With these changes, all tests in `blockchain_state_channel_SUITE` are consistently running, however, not all tests are passing consistently. Getting all the tests to pass can be addressed in other PRs.

### DIFF
--- a/test/blockchain_ct_utils.erl
+++ b/test/blockchain_ct_utils.erl
@@ -28,7 +28,8 @@
          join_packet/3,
          ledger/2,
          destroy_ledger/0,
-         download_serialized_region/1
+         download_serialized_region/1,
+         find_connected_node_pair/1
         ]).
 
 -ifdef(EQC).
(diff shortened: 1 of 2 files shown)


evanmcc (approved):

test/blockchain_ct_utils.erl
```diff
@@ -582,3 +583,53 @@ download_serialized_region(URL) ->
     {ok, Data} = file:read_file(FPath),
     Data.
 
+
+% [{Node, PubKey}]
```
xandkar, line 587: Use a spec instead. We should eventually move to also Dialyzing the test code - there's no convincing reason not to.

test/blockchain_ct_utils.erl
```diff
@@ -582,3 +583,53 @@ download_serialized_region(URL) ->
...
+                Acc#{AddrStr => Node}
+            end, #{}, NodeAddrList),
+    ct:pal("Node map: ~p", [AddrMap]),
+    find_connected_node_pair(maps:next(maps:iterator(AddrMap)), AddrMap, #{}).
+
+find_connected_node_pair(none, _, NetworkMap) ->
+    ct:pal("Final Network Map~n~p", [NetworkMap]),
+    disjoint_network;
+
+% NetworkMap - #{ConnectedToNode => [ConnectedFromNode]}
```
xandkar, line 602: Same here and other similar places - spec instead of comment.

test/blockchain_state_channel_SUITE.erl
```diff
@@ -233,13 +212,18 @@ init_per_testcase(Test, Config) ->
...
 
-    ok = check_genesis_block(InitConfig, GenesisBlock),
-    ConsensusMembers = get_consensus_members(InitConfig, ConsensusAddrs),
-    [{consensus_members, ConsensusMembers}, {master_key, MasterKey} | InitConfig].
+    ok = check_genesis_block(Nodes, GenesisBlock),
+    ConsensusMembers = get_consensus_members(Nodes, ConsensusAddrs),
+    [
+        {connected_nodes, [RouterNode, GatewayNode]},
+        {routernode, RouterNode},
+        {gatewaynode, GatewayNode},
```
xandkar, line 222: `s/routernode/router_node/g`
`s/gatewaynode/gateway_node/g`
for readability.

test/blockchain_state_channel_SUITE.erl
```diff
@@ -2791,4 +2792,4 @@ get_active_state_channel(RouterNode, SCID) ->
     end.
 
 unwrap_msg(#blockchain_state_channel_message_v1_pb{msg={Type, Msg}}) ->
-    {Type, Msg}.
\ No newline at end of file
+    {Type, Msg}.
```
xandkar, line 2795: YES! POSIX compliance FTW :)

PR #1339, opened 2022-05-11
reject poc requests txns if chainvar set

andymck:
Once val challengers are enabled, we dont want to be accepting poc request txns from hotspots, they just create noise.


src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1036,7 +1036,12 @@ validate_var(?poc_proposal_gc_window_check, Value) ->
         false -> ok;
         _ -> throw({error, {poc_proposal_gc_window_check, Value}})
     end;
-
+validate_var(?poc_reject_requests, Value) ->
+    case Value of
+        true -> ok;
+        false -> ok;
```
jeffgrunewald, line 1042:
```suggestion
    case Value of
        Bool when is_boolean(Bool) -> ok;
```

jeffgrunewald: personally, I like
```
validate_var(?poc_reject_requests, Value) when is_boolean(Value) -> ok;
validate_var(?poc_reject_requests, Value) -> throw({error, {poc_reject_requests, Value}}).
```
but I know that multi-head style is controversial

Vagabond (approved):

Vagabond:
We're going to hold this for 1.11.x and do the filtering in the sidecar until then see helium/miner#1600

PR #1341, opened 2022-05-11
ensure construct compressed snap filename

jeffgrunewald:
we're now _only_ producing gzip compressed snapshots and writing them out with the `.gz` extension; the `blockchain:get_snap/2` needs to ensure it's looking for the correct filename now that we're not producing both compressed and uncompressed snapshots

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -66,8 +66,7 @@
     add_snapshot/2, add_bin_snapshot/4,
     have_snapshot/2, get_snapshot/2, find_last_snapshot/1,
     find_last_snapshots/2,
-    save_bin_snapshots/2,
-    save_bin_snapshot/2,  hash_bin_snapshot/1, size_bin_snapshot/1,
+    hash_bin_snapshot/1, size_bin_snapshot/1,
     save_compressed_bin_snapshot/2, maybe_get_compressed_snapdata/1,
 
     add_implicit_burn/3,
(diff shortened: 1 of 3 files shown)


src/blockchain.erl
```diff
@@ -2177,7 +2177,7 @@ get_snapshot(<<Hash/binary>>, #blockchain{db=DB, dir=Dir, snapshots=SnapshotsCF}
             {error, sentinel};
         {ok, <<"file:", SnapFile/binary>>} ->
             SnapDir = filename:join(Dir, "saved-snaps"),
-            Filename = filename:join(SnapDir, SnapFile),
+            Filename = filename:join(SnapDir, SnapFile ++ ".gz"),
```
xandkar, line 2180: `.gz` should not be added here, but returned as an updated filename from the function which makes the decision to add the extension: `save_bin_snapshots/2`.

`save_bin_snapshots/2` makes the decision based on the value of `snapshot_compression_mode` option, which can be set to `both`, in which case both, the uncompressed (with original filename) and the compressed (original filename with `.gz` appended), versions are stored and thus 2 filenames need to be accounted for.

My suggestion is to modify `save_bin_snapshots/2` from:

```erl
-spec save_bin_snapshots(file:filename_all(), blockchain_ledger_snapshot:snapshot()) ->
    ok | {error, term()}.
```

to

```erl
-spec save_bin_snapshots(F, blockchain_ledger_snapshot:snapshot()) ->
    {ok, [F]} | {error, term()} when F :: file:filename_all().
```

then store the list of filenames in rocks, as `<<"files:", (term_to_binary(Filenames))/binary>>`.

To account for the existing, dangling `.gz` files, GC should check for presence of both (with and without `.gz`) in case of `file:` and just walk the list in case of `files:`.

Also, snap tests should help flag if this is working, since currently they're failing on this lack of `.gz` extension.

Snap URL fix:

```diff
-            "cd ~s && wget -c https://snapshots.helium.wtf/mainnet/~s",
+            "cd ~s && wget -c https://snapshots-dev.helium.wtf/tests/~s",
```

xandkar (changes requested):

src/blockchain.erl
xandkar: That said, if compressed is the only thing that's gonna be supported, then `snapshot_compression_mode` option handling needs to be updated. In either case `save_bin_snapshots` should return whatever filename it decided to use.

jeffgrunewald: that all makes sense; i had asked if we were saving the snap filename with the extension when i probably should have asked if we _should_ be saving the filename with the extension. as for returning any filenames written to, i don't see this being consumed anywhere unless it's just for the user calling into the cli command to save a snap. do you have anything else in mind that would use this return value?

xandkar: Line 1975 calls `save_bin_snapshots` and a couple of lines later stores `SnapFile` in rocks.

src/blockchain.erl
```diff
@@ -2070,7 +2026,7 @@ save_compressed_bin_snapshot(DestFilename, BinSnap) ->
          end,
    RetVal = blockchain_utils:streaming_transform_iolist(BinSnap, Fun),
    file:close(Out),
-   RetVal.
+   {RetVal, CompressedFile}.
```
xandkar, line 2029:
```suggestion
   {ok, CompressedFile}.
```

src/blockchain.erl
```diff
@@ -2070,7 +2026,7 @@ save_compressed_bin_snapshot(DestFilename, BinSnap) ->
          end,
    RetVal = blockchain_utils:streaming_transform_iolist(BinSnap, Fun),
```
xandkar, line 2027:
```suggestion
   ok = blockchain_utils:streaming_transform_iolist(BinSnap, Fun),
```

xandkar (approved):

PR #1343, opened 2022-05-13
More verbose ledger_poc_v3:verify error

vihu:
Problem
----
When ledger POC verification fails we can't really tell whether it failed due to incorrect challenger or the block hash.

Solution
----
This switches the return type for `poc_v3:verify` to be an `ok | {error, ...}` and passes it down.


src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
```diff
@@ -1202,8 +1202,8 @@ verify_poc_details(Txn, PoC, Keys) ->
     SigFun = libp2p_crypto:mk_sig_fun(PrivKey),
     SignedPayload = SigFun(OnionHash),
     case blockchain_ledger_poc_v3:verify(PoC, Challenger, BlockHash) of
-        false -> {error, mismatched_poc};
-        true ->
+        {error, R} -> {error, {mismatched_poc, R}};
```
xandkar, line 1205: Need to update the spec of `verify_poc_details/3` with this new error.

src/ledger/v1/blockchain_ledger_poc_v3.erl
```diff
@@ -82,9 +82,25 @@ start_height(PoC) ->
 start_height(Height, PoC) ->
     PoC#poc_v3{start_height=Height}.
 
--spec verify(poc(), libp2p_crypto:pubkey_bin(), binary()) -> boolean().
+-spec verify(poc(), libp2p_crypto:pubkey_bin(), binary()) -> ok | {error, any()}.
```
xandkar, line 85:
```suggestion
-spec verify(poc(), libp2p_crypto:pubkey_bin(), binary()) -> ok | {error, Reason}
    when
        Reason ::
            [
                {onion_key_hash   , binary()},
                {txn_challenger   , string()},
                {ledger_challenger, string()},
                {txn_blockhash    , string()},
                {ledger_blockhash , string()},
                {equal_challenger , boolean()},
                {equal_blockhash  , boolean()}
            ].
```

xandkar: Even better to give this reason a named type and then reference it both here and in the updated spec of `verify_poc_details`.

xandkar (changes requested):

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
vihu: Updated in https://github.com/helium/blockchain-core/pull/1343/commits/12860ea434228e17e99255453013f3482adfbd82

vihu: Did it slightly differently though and made the error return a tuple of tuples.

xandkar: Sensible. So close to a record!

xandkar (approved):

PR #1348, opened 2022-05-16
WIP to lookup challenge region once and thread the result through

Vagabond:
This is incomplete and it's not compatible with "old" receipts apis.


PaulVMo:
Additional thoughts on optimizing POC receipts. First, tagged_witnesses is checking both that the witness frequency is valid for the source region AND that the witness frequency matches the beacon frequency. The latter should be sufficient, especially since the regions are being checked to be the same already. At minimum, consider threading the SourceRegion frequencies through to the witness validation since it doesn't change.

Second, because the frequency check is happening and witnesses are capped to 100 km, there is not a lot of opportunity for witnessing across regions. thus, it might not even been necessary to check that the witness region is the same as the beaconer region. This would obviously be a larger change however but one that might be worth it to reduce the validation complexity.

PR #1353, opened 2022-05-18
close dangling zreference

jeffgrunewald:
Closes the dangling zlib reference when saving compressed snapshots to files


xandkar (approved):
Good catch!

I wonder if these are ever considered by GC?

PR #1354, opened 2022-05-18
Add support for multitokens

vihu:
Summary
----
- Introduce `token_version`, `ledger_entry_version` and `deprecate_security_exchange_v1` chain variables
- Add support for multi token payment transaction
- Add support for multi token coinbase transaction
- Update payment_v2 suite with ct_group support (`with_token_version` and `without_token_version`)
- Add `blockchain_token_SUITE`
- Add support for migrating ledger entries to new style ledger entries
- Deprecate security exchange v1 txn behind `deprecate_security_exchange_v1` chain variable

Requirements
----
Proto: https://github.com/helium/proto/pull/133

TODO
----

- [x] Payment nonces are borked 
- [x] 3 Payment v2 are broken
- [x] Resync testing
- [x] Add logic for payment_v1 and sec_exchange deprecation
- [x] Add support for migration ledger entries (regular and securities) to new style entries_v2
- [x] Add support and testing with aux ledger


src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -80,6 +80,16 @@ payees(Txn) ->
 amounts(Txn) ->
     [blockchain_payment_v2:amount(Payment) || Payment <- ?MODULE:payments(Txn)].
 
+-spec total_amounts_v2(txn_payment_v2()) -> #{blockchain_token_type_v1:token_type() => non_neg_integer()}.
```
madninja, line 83: no need to v2 here.. it's already disambiguated by the `s`

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -303,6 +318,75 @@ do_is_valid_checks(Txn, Chain, MaxPayments) ->
...
+        false ->
+            {error, bad_signature};
+        true ->
+            LengthPayments = length(Payments),
+            case LengthPayments == 0 of
+                true ->
+                    %% Check that there are payments
+                    {error, zero_payees};
+                false ->
+                    case blockchain_ledger_v1:find_entry_v2(Payer, Ledger) of
```
madninja, line 344: surely there's new and exciting ways to avoid the crazy ladder?

src/transactions/v2/blockchain_txn_payment_v2.erl
vihu: updated in d848c54

src/transactions/v2/blockchain_txn_payment_v2.erl
vihu: Yeah sure, just wanted to have something quick and dirty to get going

include/blockchain_vars.hrl
```diff
@@ -614,3 +614,12 @@
 
 %% Block size limit variable (in bytes). Set to 25 * 1024 * 1024.
 -define(block_size_limit, block_size_limit).
+
+%% ------------------------------------------------------------------
+%% Protocol version (aka support multiple tokens)
+%% Testing for now, set to 2
+-define(protocol_version, protocol_version).
```
jeffgrunewald, line 621: this name seems very generic/ambiguous; can it further clarify what domain/protocol it's referring to? (`helium_protocol_version`, `token_protocol_version`, etc)

vihu: Sure, I'll change it to `token_protocol_version`

vihu: Updated 9de2be3

src/transactions/v2/blockchain_payment_v2.erl
```diff
@@ -52,21 +54,33 @@ new(Payee, max) ->
 -spec new(Payee :: libp2p_crypto:pubkey_bin(),
           Amount :: non_neg_integer() | max ,
           Memo :: non_neg_integer()) -> payment().
-new(Payee, Amount, Memo) when is_integer(Amount) ->
+new(Payee, Amount, Memo) when is_integer(Memo) andalso is_integer(Amount) ->
```
jeffgrunewald, line 57: needs the spec updated if new/3 will also support the TT as an alternate third arg

src/transactions/v2/blockchain_payment_v2.erl
```diff
@@ -52,21 +54,33 @@ new(Payee, max) ->
...
-       memo=Memo,
-       max=true
+       amount=Amount,
+       memo=0,
+       max=false,
+       token_type=TT
       }.
 
+-spec new(Payee :: libp2p_crypto:pubkey_bin(),
+          Amount :: non_neg_integer(),
```
jeffgrunewald, line 74: what happens or what is our desired behavior for `max`/balance-clearing txns in multi-coin land? do we expect the user to always supply a token type and have to perform one txn for each token type supported in order to fully liquidate an account or do we need to support it returning a collection of payments that aggregates all token totals and the `max` setting would apply to all token simultaneously? cc: @abhay

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -357,6 +473,39 @@ memo_check(Txn, Ledger) ->
...
+
+    PayerHasEnoughTTBalance =
+    lists:all(
+      fun(TT) ->
+              PayerTTBalance = blockchain_ledger_entry_v2:balance(PayerEntry, TT),
+              PayerTTBalance >= maps:get(TT, TotAmounts, 0)
+      end,
+      blockchain_token_type_v1:supported_tokens()),
+
+    case blockchain:config(?allow_zero_amount, Ledger) of
```
jeffgrunewald, line 492: this block seems to imply we don't allow zero-amount txns no matter what? can we skip the var check and just validate it has all positive amounts or else throw an error?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -296,6 +302,8 @@
                                     blockchain_ledger_state_channel_v1:state_channel()
                                     | blockchain_ledger_state_channel_v2:state_channel_v2()}.
 -type h3dex() :: #{h3:h3_index() => [libp2p_crypto:pubkey_bin()]}. %% these keys are gateway addresses
+-type tagged_cf() :: tagged_cf().
```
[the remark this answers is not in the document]
vihu: Oh yeah, good catch. Probably got messed up in a rebase.

src/blockchain_token_type_v1.erl
```diff
@@ -0,0 +1,16 @@
...
+
+-type token_type() :: hnt | hst | hlt | hgt.
+-type token_types() :: [token_type()].
+
+-export_type([token_type/0, token_types/0]).
+-export([supported_tokens/0]).
+
+-spec supported_tokens() -> token_types().
+supported_tokens() ->
+    [hnt, hst, hlt, hgt].
```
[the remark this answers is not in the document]
vihu: I'll hold on this before we get some more confirmation(s). Also this probably needs an update on the proto branch first.

src/ledger/v2/blockchain_ledger_entry_v2.erl
```diff
@@ -0,0 +1,183 @@
...
+    Entry0 = new(),
+    nonce(credit(Entry0, Balance, hst), Nonce).
+
+%% ==================================================================
+%% Internal Functions
+%% ==================================================================
+
+-spec credit_hnt(Entry :: entry(), Amount :: non_neg_integer()) -> entry().
+credit_hnt(Entry, Amount) ->
+    Entry#blockchain_ledger_entry_v2_pb{hnt_balance = hnt_balance(Entry) + Amount}.
```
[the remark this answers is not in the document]
vihu: it has its quirks

src/transactions/v2/blockchain_payment_v2.erl
```diff
@@ -88,6 +102,14 @@ memo(Payment, Memo) ->
 max(Payment) ->
     Payment#payment_pb.max.
 
+-spec token_type(Payment :: payment()) -> blockchain_token_v1:type().
+token_type(Payment) ->
+    Payment#payment_pb.token_type.
+
+-spec token_type(Payment :: payment(), TT :: blockchain_token_v1:type()) -> payment().
```
jeffgrunewald, line 109: do we want to allow the token type of a payment to be set after the record has been created? i notice we don't have setter functions for payee, amount (and max b/c it has a direct impact on amount) but we do have a setter for memo. this seems like the type of thing we might want to make immutable

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -357,63 +469,113 @@ memo_check(Txn, Ledger) ->
...
+                                                 {TokenType, MaxPaymentAmt}
+                                             end, MaxPayments),
+                    {MaxPayments1, SpecifiedPayments1};
+                _ ->
+                    %% Prior to token_version 2 there can only be a single balance-clearing
+                    %% `max` payment per txn
+                    {ok, PayerEntry} = blockchain_ledger_v1:find_entry(Payer, Ledger),
+                    Balance = blockchain_ledger_entry_v1:balance(PayerEntry),
+                    SpecifiedAmts = lists:foldl(fun({_, Val}, Acc) -> Val + Acc end, 0, SpecifiedPayments1),
+                    MaxPaymentAmt = Balance - SpecifiedAmts - Fee,
```
jadeallenx, line 551: Shouldn't this be renamed to `RemainingTokenBalance` or something like that? It would be nice if the shape of this data type and the purpose of it was documented in a comment

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2852,6 +2868,90 @@ hnt_to_dc(HNTAmount, Ledger)->
...
+                    EntryV1Balance = blockchain_ledger_entry_v1:balance(EntryV1),
+                    EntryV1Nonce = blockchain_ledger_entry_v1:nonce(EntryV1),
+                    NewEntryV2 = blockchain_ledger_entry_v2:nonce(
+                      blockchain_ledger_entry_v2:credit(
+                        blockchain_ledger_entry_v2:new(), EntryV1Balance, hnt),
+                      EntryV1Nonce),
+                    Bin = blockchain_ledger_entry_v2:serialize(NewEntryV2),
+                    cache_put(Ledger, EntriesV2CF, Address, Bin);
+                _ ->
+                    %% NOTE: This should never happen due to the order
```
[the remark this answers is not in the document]
vihu: Yeah, this code is sus, I'm not really sure how to handle the failing scenario. We definitely need a more robust way to migrate the ledger entries. Let's say one or more validators do crash while doing the migration, will they not just have the right entries as soon as the next block (assuming the CG survives the migration) comes in? Probably need input from @evanmcc about this. I did the absolute simplest thing here for time sakes.

Another way would be to have the CG handle the migration entirely and do a "special" block when the var kicks in which tells everyone else to have the v2-style entries? Maybe?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2966,19 +3066,41 @@ trim_price_list(LastTime, PriceEntries) ->
...
+                  EntriesCF :: tagged_cf(),
+                  Ledger :: ledger()) ->
+    {ok, blockchain_ledger_entry_v1:entry()}
+    | {ok, blockchain_ledger_entry_v2:entry()}
+    | {error, any()}.
+find_entry_(Address, EntryMod, EntriesCF, Ledger) ->
     case cache_get(Ledger, EntriesCF, Address, []) of
         {ok, BinEntry} ->
-            {ok, blockchain_ledger_entry_v1:deserialize(BinEntry)};
+            {ok, EntryMod:deserialize(BinEntry)};
```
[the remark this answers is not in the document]
vihu: Sure, we can do that. I thought about that initially but wanted to keep the ser/deser separate in the modules. Don't really have a strong opinion about either approach though.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -296,6 +301,8 @@
                                     blockchain_ledger_state_channel_v1:state_channel()
                                     | blockchain_ledger_state_channel_v2:state_channel_v2()}.
 -type h3dex() :: #{h3:h3_index() => [libp2p_crypto:pubkey_bin()]}. %% these keys are gateway addresses
+-type tagged_cf() :: {atom(), rocksdb:db_handle(), rocksdb:cf_handle()}.
```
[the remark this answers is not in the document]
vihu: This was just a type spec copied in multiple places so I moved it up here

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2998,32 +3120,90 @@ credit_account(Address, Amount, Ledger) ->
...
+                        true ->
+                            Entry0 = maps:fold(
+                                       fun(TT, Amt, Acc) ->
+                                               blockchain_ledger_entry_v2:debit(Acc, Amt, TT)
+                                       end, Entry, AmountOrAmounts),
+                            Entry1 = blockchain_ledger_entry_v2:nonce(Entry0, Nonce),
+                            Bin = blockchain_ledger_entry_v2:serialize(Entry1),
+                            EntriesCF = entries_v2_cf(Ledger),
+                            cache_put(Ledger, EntriesCF, Address, Bin);
+                        false ->
```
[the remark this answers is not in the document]
vihu: Not quite what you asked but I added a better error msg on txn validation, check eb2106c

include/blockchain_vars.hrl
```diff
@@ -614,3 +614,14 @@
 
 %% Block size limit variable (in bytes). Set to 25 * 1024 * 1024.
 -define(block_size_limit, block_size_limit).
+
+%% ------------------------------------------------------------------
+%% Token version (aka support multiple tokens)
+%% Testing for now, set to 2
```
evanmcc, line 620: might clean up this description

src/ledger/v2/blockchain_ledger_entry_v2.erl
```diff
@@ -0,0 +1,183 @@
...
+    Balance = blockchain_ledger_security_entry_v1:balance(SecEntryV1),
+    Nonce = blockchain_ledger_security_entry_v1:nonce(SecEntryV1),
+    Entry0 = new(),
+    nonce(credit(Entry0, Balance, hst), Nonce).
+
+%% ==================================================================
+%% Internal Functions
+%% ==================================================================
+
+-spec credit_hnt(Entry :: entry(), Amount :: non_neg_integer()) -> entry().
```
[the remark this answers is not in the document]
evanmcc: yeah it'd make more sense to me to have one `credit/3` fun but ultimately I don't care strongly.

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -745,6 +745,10 @@ var_hook(?poc_challenger_type, Type, Ledger) ->
     end,
     purge_pocs(Ledger),
     ok;
+var_hook(?ledger_entry_version, 2, Ledger) ->
+    %% Migrate to new style ledger entries
+    blockchain_ledger_v1:migrate_entries(Ledger),
```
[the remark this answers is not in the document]
evanmcc: has to be sync, sadly, that's how the hooks work.

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1520,6 +1529,21 @@ validate_var(?discard_zero_freq_witness, Value) ->
 validate_var(?block_size_limit, Value) ->
     validate_int(Value, "block_size_limit", 1*1024*1024, 512*1024*1024, false);
 
+validate_var(?token_version, Value) ->
+    case Value of
+        undefined -> ok;
```
evanmcc, line 1534: don't understand why `undefined` is allowed here?  should just unset?

src/transactions/v2/blockchain_payment_v2.erl
evanmcc: I'm not sure that it matters?  the signature will lock it in in any case?

src/transactions/v1/blockchain_txn_vars_v1.erl
vihu: Addressed in [f5520a3](https://github.com/helium/blockchain-core/pull/1354/commits/f5520a3c0dbd7470294febcc441d9ce7a8465ff6)

include/blockchain_vars.hrl
vihu: Fixed in [f5520a3](https://github.com/helium/blockchain-core/pull/1354/commits/f5520a3c0dbd7470294febcc441d9ce7a8465ff6)

src/ledger/v2/blockchain_ledger_entry_v2.erl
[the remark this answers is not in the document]
vihu: Updated in [f5520a3](https://github.com/helium/blockchain-core/pull/1354/commits/f5520a3c0dbd7470294febcc441d9ce7a8465ff6)

src/blockchain_token_v1.erl
```diff
@@ -0,0 +1,16 @@
...
+-module(blockchain_token_v1).
+
+-type type() :: hnt | hst | mobile | iot.
+-type types() :: [type()].
+
+-export_type([type/0, types/0]).
+-export([supported_tokens/0]).
+
+-spec supported_tokens() -> types().
+supported_tokens() ->
```
jeffgrunewald, line 15: is there any way to define this based on the proto enum definition in `blockchain_token_type_v1`?

src/ledger/v1/blockchain_ledger_v1.erl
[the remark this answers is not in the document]
jeffgrunewald: could we delete the v1 entry at the same time we're converting to the v2 version and commit atomically so even if the process is interrupted we're just picking up where we left off on recovery?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2976,71 +3076,161 @@ trim_price_list(LastTime, PriceEntries) ->
...
+debit_account(Address, AmountOrAmounts, Nonce, Ledger) when is_integer(AmountOrAmounts) ->
+    %% debit_account being called with an Amount can occur in two situations:
+    %% - token_version is not set
+    %% - token_version is set but for aux ledger
+    %% - we're in some weird transitionary state
+    %% - payment_v1 is active on chain (and those only work with HNT token)
+    %% It's best to make sure that the right ledger entry mod is being invoked
+
+    {EntryMod, EntriesCF} = versioned_entry_mod_and_entries_cf(Ledger),
     case ?MODULE:find_entry(Address, Ledger) of
```
jeffgrunewald, line 3174: can you save yourself a lookup of the ledger entry version here and use `find_entry_/4` or is that an immaterial cost?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2976,71 +3076,161 @@ trim_price_list(LastTime, PriceEntries) ->
...
+                {error, _}=Error ->
+                    Error
+            end
+    end.
+
+-spec credit_account(Address :: libp2p_crypto:pubkey_bin(),
+                     Amount :: integer(),
+                     TT :: blockchain_token_v1:type(),
+                     Ledger :: ledger()) -> ok | {error, any()}.
+credit_account(Address, Amount, TT, Ledger) ->
```
jeffgrunewald, line 3142: why do we support debits via a map of token type/amount but not credits?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2976,71 +3076,161 @@ trim_price_list(LastTime, PriceEntries) ->
...
+                            Entry0 = maps:fold(
+                                       fun(TT, Amt, Acc) ->
+                                               EntryMod:debit(Acc, Amt, TT)
+                                       end, Entry, AmountOrAmounts),
+                            Entry1 = EntryMod:nonce(Entry0, Nonce),
+                            Bin = EntryMod:serialize(Entry1),
                             cache_put(Ledger, EntriesCF, Address, Bin);
                         false ->
-                            {error, {insufficient_balance, {Amount, Balance}}}
+                            {error, {insufficient_balance, {libp2p_crypto:bin_to_b58(Address), AmountOrAmounts}}}
```
jeffgrunewald, line 3217: should we indicate which token(s) had the insufficient balance?

src/ledger/v1/blockchain_ledger_v1.erl
jeffgrunewald: or perhaps just make `find_entry/4` public and pull the check for the module and cf out of the current private version?

src/ledger/v1/blockchain_ledger_v1.erl
vihu: Didn't want to change `credit` behavior much to still remain in accordance with payment_v1 support (which we really should be looking to deprecate, but it still lingers)

src/ledger/v1/blockchain_ledger_v1.erl
vihu: We still need to get the CF for storing the changes and the actual entry mod for serialization

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3053,12 +3243,8 @@ debit_fee_from_account(Address, Fee, Ledger, TxnHash, Chain) ->
                         false ->
                             ok
                     end,
-                    Entry1 = blockchain_ledger_entry_v1:new(
-                        blockchain_ledger_entry_v1:nonce(Entry),
-                        (Balance - Fee)
-                    ),
-                    EntryBin = blockchain_ledger_entry_v1:serialize(Entry1),
-                    EntriesCF = entries_cf(Ledger),
+                    Entry1 = EntryMod:new(EntryMod:nonce(Entry), (Balance - Fee)),
```
jeffgrunewald, line 3246: does this need to be parameterized by ledger module and cf versions? i don't recall seeing a migration of the DC entries.

vihu: I could be wrong but whenever `debit_fee_from_account` is called, the `fee` getting passed in is being converted to `hnt`, which would imply that the deduction needs to be versioned according token_version.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2976,71 +3076,161 @@ trim_price_list(LastTime, PriceEntries) ->
...
+    {EntryMod, EntriesCF} = versioned_entry_mod_and_entries_cf(Ledger),
+    case ?MODULE:find_entry(Address, Ledger) of
+        {error, _}=Error ->
+            Error;
+        {ok, Entry} ->
+            case Nonce =:= EntryMod:nonce(Entry) + 1 of
+                true ->
+                    case lists:all(
+                      fun(TT) ->
+                              EntryMod:balance(Entry, TT) >= maps:get(TT, AmountOrAmounts, 0)
```
jeffgrunewald, line 3204: maybe too complex in a single function when juggling the two different ledger entry versions but could this  not use the `check_balance` function instead?

src/ledger/v2/blockchain_ledger_entry_v2.erl
```diff
@@ -0,0 +1,156 @@
...
+        nonce = 0,
+        hnt_balance = 0,
+        hst_balance = 0,
+        mobile_balance = 0,
+        iot_balance = 0
+    }.
+
+%% NOTE: This function is for 1:1 correspondence entry_v1
+-spec new(Nonce :: non_neg_integer(), HNTBalance :: non_neg_integer()) -> entry().
+new(Nonce, HNTBalance) when Nonce /= undefined andalso HNTBalance /= undefined ->
```
jeffgrunewald, line 46: should these guards be `is_integer/1` to be more restrictive in what's allowed?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2862,6 +2878,90 @@ hnt_to_dc(HNTAmount, Ledger)->
...
+                    %% This should _always_ occur as we assume no entries_v2 exist
+                    %% on ledger yet
+                    EntryV1Balance = blockchain_ledger_entry_v1:balance(EntryV1),
+                    EntryV1Nonce = blockchain_ledger_entry_v1:nonce(EntryV1),
+                    NewEntryV2 = blockchain_ledger_entry_v2:nonce(
+                      blockchain_ledger_entry_v2:credit(
+                        blockchain_ledger_entry_v2:new(), EntryV1Balance, hnt),
+                      EntryV1Nonce),
+                    Bin = blockchain_ledger_entry_v2:serialize(NewEntryV2),
+                    cache_put(Ledger, EntriesV2CF, Address, Bin);
```
jeffgrunewald, line 2912: can we use the `from_v1/2` functions here to do these conversions? where else would we use those if not here?

src/ledger/v1/blockchain_ledger_v1.erl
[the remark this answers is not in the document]
evanmcc: A partial failure scenario would merit a snapshot reload; we assume that essentially all of the writes in a particular block happen at the same time (in txn order, when there are interrelations).  There is no other code around to manage partial conversions.  While it's not wonderful, it's kind of where we are right now.

src/ledger/v1/blockchain_ledger_v1.erl
jeffgrunewald: i forgot what function this was referenced in and was just remembering the `debit_fee` function

jeffgrunewald: that being said, this only assumes implicit burn of hnt; do we have any plans to support that for other tokens once we have users who maybe only have one type of subnetwork token and no hnt yet or keeping it simple and only assuming hnt as a "reserve token" for implicit burn shortfalls? i can see the queries to price oracles getting pretty complicated if we did and we'd either need to allow the user to do something like designate a token or tokens as implicit burn targets like overdraft protection on a bank account or else do some erlang-style arbitrary sorting order of tokens to try and burn in order until a target is reached.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2862,6 +2878,90 @@ hnt_to_dc(HNTAmount, Ledger)->
...
+-spec migrate_sec_entries(Ledger :: ledger()) -> ok | {error, any()}.
+migrate_sec_entries(Ledger) ->
+    SecuritiesCF = securities_cf(Ledger),
+    EntriesV2CF = entries_v2_cf(Ledger),
+    cache_fold(
+        Ledger,
+        SecuritiesCF,
+        fun({Address, Binary}, ok) ->
+            EntryV1 = blockchain_ledger_security_entry_v1:deserialize(Binary),
+            case ?MODULE:find_entry(Address, Ledger) of
```
evanmcc, line 2932: given the level of duplication here, would it be possible to fold these together just for clarity?

src/ledger/v1/blockchain_ledger_v1.erl
vihu: Oh good catch, this is precisely where I intended to use the `from_v1` funs, but had a brain fart clearly. Fixed [7cb5fb5](https://github.com/helium/blockchain-core/pull/1354/commits/7cb5fb5d1d61b6a5b85da70b726442717ea9234d)

src/ledger/v1/blockchain_ledger_v1.erl
vihu: I just cleaned these funs up a bit, still has some duplication but they still differ enough to warrant their own funs I think

evanmcc (approved):

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2976,71 +3066,161 @@ trim_price_list(LastTime, PriceEntries) ->
...
+                    Balance = EntryMod:balance(Entry),
+                    case (Balance - AmountOrAmounts) >= 0 of
                         true ->
-                            Entry1 = blockchain_ledger_entry_v1:new(
-                                Nonce,
-                                (Balance - Amount)
-                            ),
-                            Bin = blockchain_ledger_entry_v1:serialize(Entry1),
-                            EntriesCF = entries_cf(Ledger),
+                            Entry1 = EntryMod:new( Nonce, (Balance - AmountOrAmounts)),
```
andymck, line 3173: Same as comment above

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2976,71 +3066,161 @@ trim_price_list(LastTime, PriceEntries) ->
...
+                    {error, {bad_nonce, {payment, Nonce, EntryMod:nonce(Entry)}}}
+            end
+    end;
+debit_account(Address, AmountOrAmounts, Nonce, Ledger) when is_map(AmountOrAmounts) ->
+    %% For consistency sake we'll get the entry mod here as well (it should ideally always be v2)
+    {EntryMod, EntriesCF} = versioned_entry_mod_and_entries_cf(Ledger),
+    case ?MODULE:find_entry(Address, Ledger) of
+        {error, _}=Error ->
+            Error;
+        {ok, Entry} ->
```
andymck, line 3189: Similar to earlier comment, what happens if you hit a non migrated v1 entry ?  I think version tagging the v2s gives you the ability to match these on all cache reads/folds and find_entrys and handle stale v1 appropriately

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2976,71 +3066,161 @@ trim_price_list(LastTime, PriceEntries) ->
...
+                            Entry0 = maps:fold(
+                                       fun(TT, Amt, Acc) ->
+                                               EntryMod:debit(Acc, Amt, TT)
+                                       end, Entry, AmountOrAmounts),
+                            Entry1 = EntryMod:nonce(Entry0, Nonce),
+                            Bin = EntryMod:serialize(Entry1),
                             cache_put(Ledger, EntriesCF, Address, Bin);
                         false ->
-                            {error, {insufficient_balance, {Amount, Balance}}}
+                            {error, {insufficient_balance, {libp2p_crypto:bin_to_b58(Address), AmountOrAmounts}}}
```
andymck, line 3207: any possible side effects from the two debit_account/4 functions returning differing insufficient_balance errors ? i expect not as the last tuple is commonly used for logging purposes only

src/ledger/v1/blockchain_ledger_v1.erl
andymck: confirming the fee passed to debit_fee_from_account/5 is expected to be in HNT

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -2862,6 +2878,80 @@ hnt_to_dc(HNTAmount, Ledger)->
...
 
+%%--------------------------------------------------------------------
+%% @doc
+%% Migrate ledger entries
+%% @end
+%%--------------------------------------------------------------------
+-spec migrate_entries(Ledger :: ledger()) -> ok | {error, any()}.
+migrate_entries(Ledger) ->
+    ok = migrate_reg_entries(Ledger),
+    ok = migrate_sec_entries(Ledger),
```
andymck, line 2889: What if migrate_reg_entries succeeds and subsequently migrate_sec_entries fails?

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -357,63 +469,137 @@ memo_check(Txn, Ledger) ->
...
+      fun(TT, Acc) ->
+              PayerTTBalance = blockchain_ledger_entry_v2:balance(PayerEntry, TT),
+              Amount = maps:get(TT, TotAmounts, 0),
+              IsSufficient = PayerTTBalance >= Amount,
+              [{TT, IsSufficient, PayerTTBalance, Amount} | Acc]
+      end,
+      [],
+      blockchain_token_v1:supported_tokens()),
+
+    case lists:filter(
```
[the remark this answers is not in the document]
vihu: > you wouldn't as easily have the insufficient payment token/amount to bubble up in the error response?
 Yup, that's precisely it. We'd lose the details of the insufficient funds and related token types.

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: txn will fail, node will halt if a retry hits the same issue.

src/ledger/v1/blockchain_ledger_v1.erl
evanmcc: they're in different CFs, though, so presumably we'd drop the old entries and hst one when we're done?.  also historically we've handled upgrades transparently in deserialize.

evanmcc (approved):

PR #1355, opened 2022-05-18
add as923_1b region

mikev:
AS923_1B defines a Region supporting Malaysia. It is illegal in Malaysia to use any frequency above 924 Mhz, hence the reason for this new Region. The AS923_1B defines Uplink frequencies at [922.0, 922.2, 922.4, 922.6, 922.8, 923.0, 923.2, 923.4]

I don't believe this change broke any CT tests, but some CT tests still fail on buildkite and some tests were commented to make further test progress on buildkite.

This change is a prerequisite for - https://github.com/helium/miner/pull/1664

### DIFF
--- a/test/blockchain_region_SUITE.erl
+++ b/test/blockchain_region_SUITE.erl
@@ -195,6 +195,12 @@ init_per_testcase(TestCase, Config) ->
 %%--------------------------------------------------------------------
 %% test cases
 %%--------------------------------------------------------------------
+
+
+%% H3 located from https://github.com/helium/lorawan-h3
+
+%% H3 created from HPlans at https://github.com/dewi-alliance/hplans
+
 region_param_test(Config) ->
     Ledger = ?config(ledger, Config),
     H3 = 631183727389488639,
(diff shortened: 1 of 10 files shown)


test/blockchain_region_SUITE.erl
```diff
@@ -213,6 +217,14 @@ as923_1_test(Config) ->
     false = blockchain_region_v1:h3_in_region(H3, region_us915, Ledger),
     ok.
 
+as923_1b_test(Config) ->
+    Ledger = ?config(ledger, Config),
+    H3 = 631319855840474623,
```
vihu, line 222: Maybe add a comment here stating where this H3 index is located?

vihu:
I think this looks good generally. We need to start consolidating all the necessary chain vars. The ones with `region_*` are the most cumbersome ones as they depend on serialized regions from <https://github.com/helium/lorawan-h3>

test/blockchain_region_SUITE.erl
mikev: Add comment with link to H3 and HPlans repos

Vagabond (approved):
All this adds is symbol lookups for the new region, everything else is no-op or testing related.

vihu (approved):

PR #1358, opened 2022-05-19
Always process gw reactivation list

### LINKED ISSUE #1357: reactivation list within a HB is not processed if the node is in consensus
Since Validator challenges, a val heartbeat txn can contain two additional lists:

1.  A list of POC key proposals
2. A list of gateways to reactivate

When processing a HB txn there is a check to determine if the validator is in consensus.   If true we dont process either of the above lists.  This behaviour is incorrect.  We should always process the activation list ( assuming inactivity is enabled via its chain var ).  The key proposals should continue to be processed only if the node is not in consensus.

By not processing the reactivation list, GWs are not being marked as active and continue to be excluded from POC.
The issue presented itself as extremely apparent on testnet where we only had a total of 5 vals, with a CG size of 4.  This resulted in GWs almost never being reactivated as they were always connecting to a validator which was in consensus.  On mainnet the visibility of this will be minor as the availability of non consensus vals is significant.

jadeallenx:
Problem to solve: Currently, gateway "last activity" is updated only when a validator is _not_ in consensus. But this behavior is incorrect. The reactivation list must always be processed so that gateways are properly marked as "active again" and thus may be considered for POC targeting in the future.

Solution: Always process the reactivation list, even if a node is in consensus.

Fixes #1357

### DIFF
--- a/src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
+++ b/src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
@@ -199,7 +199,14 @@ absorb(Txn, Chain) ->
             {ok, ConsensusAddrs} = blockchain_ledger_v1:consensus_members(Ledger),
             case lists:member(Validator, ConsensusAddrs) of
                 true ->
-                    ok;
+                    %% gateway reactivation must be processed regardless of consensus status
+                    %% see GH#1357
+                    case blockchain_ledger_v1:config(?poc_always_process_reactivations, Ledger) of
+                        {ok, true} ->
+                            maybe_reactivate_gateways(Txn, Ledger),
+                            ok;
+                        _ -> ok
+                    end;
                 false ->
                     case blockchain:config(poc_challenger_type, Ledger) of
                         {ok, validator} ->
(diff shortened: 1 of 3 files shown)


src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
```diff
@@ -199,6 +199,9 @@ absorb(Txn, Chain) ->
             {ok, ConsensusAddrs} = blockchain_ledger_v1:consensus_members(Ledger),
             case lists:member(Validator, ConsensusAddrs) of
                 true ->
+                    %% gateway reactivation must be processed regardless of consensus status
+                    %% see GH#1357
+                    maybe_reactivate_gateways(Txn, Ledger),
```
jeffgrunewald, line 204: do we need this case statement at all anymore? seems like the question of `maybe_reactivate_gateways/2` is always "yes"

jeffgrunewald: okay, nvm the question about the case statement, github was hiding the extra steps we take in the `false` case, but it still seems like the private function should just be called `reactivate_gateways/2`

jeffgrunewald (approved):

andymck:
@jadeallenx sorry i should have noted in the issue but the change will need to be behind a new var, otherwise sync will break due to targeting issues ( ie having an additional list of active hotspots to choose from )

jadeallenx:
> @jadeallenx sorry i should have noted in the issue but the change will need to be behind a new var, otherwise sync will break due to targeting issues ( ie having an additional list of active hotspots to choose from )

OK, I will update this behind a var

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
jeffgrunewald: re-read further; apparently i missed the `poc_activity_filter_enabled` chain that makes this a "maybe" so ignore me.

jeffgrunewald (approved):

andymck (approved):

PR #1359, opened 2022-05-19
var gated bypass of witness validation

jeffgrunewald:
Allows skipping of witness validation when processing V2 PoC receipts, which have a heavy performance burden on receipt processing.

### DIFF
--- a/src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
+++ b/src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
@@ -326,7 +326,13 @@ poc_particpants(Txn, Chain) ->
                                     _ ->
                                         {lists:nth(ElementPos - 1, Path), lists:nth(ElementPos - 1, Channels), lists:nth(ElementPos, Channels)}
                                 end,
-                            Witnesses = valid_witness_addrs(Element, WitnessChannel, Ledger),
+                            Witnesses = case blockchain:config(?poc_receipt_witness_validation, Ledger) of
+                                            {ok, false} ->
+                                                UnvalidatedWitnesses = lists:reverse(blockchain_poc_path_element_v1:witnesses(Element)),
+                                                [W#blockchain_poc_witness_v1_pb.gateway || W <- UnvalidatedWitnesses];
+                                            _ ->
+                                                valid_witness_addrs(Element, WitnessChannel, Ledger)
+                                        end,
                             %% only include the challengee in the poc participants list
                             %% if we have received a receipt
                             ElemParticipants =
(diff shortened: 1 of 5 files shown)


src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
```diff
@@ -855,19 +855,24 @@ valid_receipt(PreviousElement, Element, Channel, Ledger) ->
                       Channel :: non_neg_integer(),
                       Ledger :: blockchain_ledger_v1:ledger()) -> blockchain_poc_witness_v1:poc_witnesses().
 valid_witnesses(Element, Channel, Ledger) ->
-    TaggedWitnesses = tagged_witnesses(Element, Channel, Ledger),
```
andymck, line 858: If you change `valid_witnesses` then when the chain var is activated the witnesses will not be validated during rewards calculation.  Is that intentional ?  The original plan was to only remove validations in the poc_participants function ie 

https://github.com/helium/blockchain-core/blob/7f9c93ee96efce96ed491cf5383e99dcf0755140/src/transactions/v2/blockchain_txn_poc_receipts_v2.erl#L329

The line above could be wrapped in the var check and if enabled then maybe just pull the witnesses via

Witnesses = blockchain_poc_path_element_v1:witnesses(Element)

jeffgrunewald: It was intentional but maybe not necessary or intentionally too broad. I was worried bypassing validation in one area but not everywhere would cause a discrepancy but if that’s not going to have any undesirable consequences then that check is easy to move

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
```diff
@@ -855,8 +861,8 @@ valid_receipt(PreviousElement, Element, Channel, Ledger) ->
...
-    [ W || {true, _, W} <- TaggedWitnesses ].
+    {ok, RegionVars} = blockchain_region_v1:get_all_region_bins(Ledger),
```
andymck, line 864: this is an optimisation change ?

jeffgrunewald: Yeah; that was so I could safely thread valid_witnesses/3 through valid_witnesses/4. Probably not strictly necessary now that I’m not putting the var gate in the valid_witnesses/4.

test/blockchain_simple_SUITE.erl
```diff
@@ -3723,6 +3771,114 @@ receipts_txn_dont_reject_empty_receipt_test(Config) ->
...
+    _ = blockchain_gossip_handler:add_block(Block, Chain, self(), blockchain_swarm:tid()),
+
+    ok = test_utils:wait_until(fun() -> {ok, 4} =:= blockchain:height(Chain) end),
+
+    Ledger = blockchain:ledger(Chain),
+
+    W1Info = blockchain_ledger_v1:find_gateway_info(Witness1, Ledger),
+    ct:pal("Gateway Info: Witness1 = ~p", [W1Info]),
+
+    ?assertEqual({ok, undefined}, blockchain_ledger_v1:find_gateway_last_challenge(Witness1, Ledger)),
```
jeffgrunewald, line 3820: I would prefer to have the "enable receipt validations test" to have the first witness be valid and the second witness be invalid but I wasn't able to find hexes to assert their locations at that didn't result in badmath errors when calculating `blockchain_utils:min_rcv_sig/2` or else failed to pass the `witness_too_close` validation (the intention for the first witness to be valid and the second witness to fail validation because of 0.0 frequency)

evanmcc: I've found that the easiest way to get usable, understandable h3 test locations is to use `h3:from_geo/2` and just extend the lat or the long until you get the result you need.  There's some bitwise stuff going on in the hex encoding that makes manipulating it directly a bit difficult.

andymck (approved):
this lgtm, assuming the tests pass...not had a chance to run locally

evanmcc (approved):

PR #1360, opened 2022-05-20
Add var to harmonize a GWs activity on hip17 interactivity blocks

andymck:
Related PR:  https://github.com/helium/sibyl/pull/51

After 19.1 was cut it was realised that there were two vars in use to determine a GW was active or inactive, these were:

poc_v4_target_challenge_age
hip17_interactivity_blocks

This harmonises the activity determination to use the hip17 var ( which has a higher permitted threshold of 5000 compared to 1000 for the former )

An alternative approach was to add a completely new var to determine the number of blocks before a GW is marked as inactive and replace both the above vars. The only scenario in which I thought that justified was if the max value of 5000 is considered constrictive.  If that is the case then I will rework this....

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -4893,10 +4893,18 @@ maybe_gc_h3dex(Ledger) ->
     %% pick a random h3dex index and remove any inactive hotspots from it
     case ?MODULE:config(?h3dex_gc_width, Ledger) of
         {ok, Width} ->
-            InactivityThreshold =
+            InactivityThreshold0 =
               case ?MODULE:config(?poc_v4_target_challenge_age, Ledger) of
-                {ok, InActV} -> InActV;
+                {ok, V1} -> V1;
                 _ -> 10
+              end,
+            InactivityThreshold1 =
+              case ?MODULE:config(?harmonize_activity_on_hip17_interactivity_blocks, Ledger) of
+                {ok, true} ->
+                    {ok, V2} = ?MODULE:config(?hip17_interactivity_blocks, Ledger),
+                    V2;
+                _ ->
+                    InactivityThreshold0
               end,
             %% we need a fairly deterministic way to choose hexes to be GC'd
             %% that ideally is not tied to internal representations like rocksdb
@@ -4927,7 +4935,7 @@ maybe_gc_h3dex(Ledger) ->
                                           Challengee = blockchain_poc_path_element_v1:challengee(hd(Path)),
                                           case find_gateway_location(Challengee, Ledger) of
                                               {ok, Location} ->
-                                                  gc_h3dex_hex(Location, Height, InactivityThreshold, Ledger);
+                                                  gc_h3dex_hex(Location, Height, InactivityThreshold1, Ledger);
                                               _ ->
                                                   ok
                                           end
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -4914,32 +4914,37 @@ maybe_gc_h3dex(Ledger) ->
             %% from the current block (which are sorted by *challenger* and GC the
             %% hexes the *challengee* is in.
             {ok, Height} = current_height(Ledger),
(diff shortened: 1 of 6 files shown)


src/poc/blockchain_poc_target_v6.erl
```diff
@@ -182,8 +182,11 @@ filter(AddrList, Height, Ledger) ->
             {ok, V} -> V;
             _ -> false
         end,
-    {ok, MaxActivityAge} = blockchain:config(poc_v4_target_challenge_age, Ledger),
-
+    {ok, MaxActivityAge} =
```
madninja, line 185: use `max_challenge_age/1`

src/poc/blockchain_poc_target_v5.erl
```diff
@@ -219,4 +219,7 @@ is_active(_, _Gateway, _Height, _Vars, _Ledger) ->
 
 -spec max_activity_age(Vars :: map()) -> pos_integer().
 max_activity_age(Vars) ->
-    maps:get(poc_v4_target_challenge_age, Vars).
+    case maps:get(harmonize_activity_on_hip17_interactivity_blocks, Vars, false) of
```
madninja, line 222: why is function here the same as challenge_age?

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -4893,11 +4893,19 @@ maybe_gc_h3dex(Ledger) ->
     %% pick a random h3dex index and remove any inactive hotspots from it
     case ?MODULE:config(?h3dex_gc_width, Ledger) of
         {ok, Width} ->
-            InactivityThreshold =
+            InactivityThreshold0 =
```
madninja, line 4896: this looks like it's the same logic as challenge_age and max_activity_age? Why not share that code somehow?

andymck: because this one has a default value of 10 if poc_v4_target_challenge_age not set....it is prob not needed but at this point better safe than sorry

src/poc/blockchain_poc_target_v5.erl
andymck: challenge age in v4 was called that way as it went in prior to validator challenges.  v5 and v6 came with validator challenges and renamed to better suit the new intention based on activity

madninja: so can we remove challenge_age/1 if max_activity_age/` with the new, better name and does the exact same thing?

src/ledger/v1/blockchain_ledger_v1.erl
madninja: sorry ,where's the 10 default?

andymck: https://github.com/helium/blockchain-core/blob/master/src/ledger/v1/blockchain_ledger_v1.erl#L4899

src/poc/blockchain_poc_target_v5.erl
andymck: the v4 module is for h3dex targeting when hotspot challenges enabled.  that module will be redundant shortly and can be deleted then.

vihu:
I think this generally looks okay, but I don't like that there are changes here which are potentially unrelated to the problem at hand. If possible can we limit the scope to just handle the `challenge_age` and `harmonize...` var? The GC stuff is already gated behind `h3dex_gc_width` which is currently not set on chain, unless we plan on activating that with the next release, we should probably do a separate PR to the limit scope of this one. 

Regardless, stamping :heavy_check_mark:

vihu:
I think this generally looks okay, but I don't like that there are changes here which are potentially unrelated to the problem at hand. If possible can we limit the scope to just handle the `challenge_age` and `harmonize...` var? The GC stuff is already gated behind `h3dex_gc_width` which is currently not set on chain, unless we plan on activating that with the next release, we should probably do a separate PR to the limit scope of this one. 

Regardless, stamping :heavy_check_mark:

vihu (approved):
Approving with one general comment.

jeffgrunewald (approved):

PR #1365, opened 2022-05-25
separate out purging of active and proposed POCs

andymck:
This breaks out the purging of active and proposed POCs into two functions allowing each to be called independently.  

Doing so prevents the proposed POCs being wiped when we change targeting.  This avoids all poc proposals being lost when we enabled h3dex on mainnet

The only time we wipe the poc proposals with this fix is when poc_challenger_type is changed or unset.


andymck:
closed as decided to accept the loss of the poc keys when enabling h3dex

PR #1371, opened 2022-05-27
Attempt to send txn immediately

PaulVMo:
Currently, when a transaction is submitted to the blockchain txn manager, it is added to the transaction cache and then only sent to the CG after the next block add event has been processed. This change also attempts to process newly added transaction immediately so they are sent to the CG sooner. This change may help smooth out filling of txn buffers for CG members and avoid large oscillations in txns per block.

### DIFF
--- a/src/transactions/blockchain_txn_mgr.erl
+++ b/src/transactions/blockchain_txn_mgr.erl
@@ -162,7 +162,7 @@ handle_cast({submit, Txn, Key, Callback}, State=#state{chain = undefined}) ->
     end,
     {noreply, State};
 
-handle_cast({submit, Txn, Key, Callback}, State=#state{cur_block_height = H}) ->
+handle_cast({submit, Txn, Key, Callback}, State=#state{cur_block_height = H, chain = Chain, submit_f = SubmitF}) ->
     %% add the txn to the cache
     %% provided key will be utilised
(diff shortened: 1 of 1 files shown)


andymck (approved):

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -162,7 +162,7 @@ handle_cast({submit, Txn, Key, Callback}, State=#state{chain = undefined}) ->
     end,
     {noreply, State};
 
-handle_cast({submit, Txn, Key, Callback}, State=#state{cur_block_height = H}) ->
+handle_cast({submit, Txn, Key, Callback}, State=#state{cur_block_height = H, chain = Chain, submit_f = SubmitF}) ->
```
andymck, line 165: thinking a bit more about this, I dont think we should have the direct submit as the default path.  It means we bypass all the validations for every incoming txn including all external txns being injected by ETL, putting all the load onto the CG to perform validations. This could maybe result in unexpected results such as those txns with an out of sequence nonce...which currently would remain in the txn mgr cache until they become valid or definitively invalid.

We should instead add a flow whereby we can submit a local txn ( ie a txn being submitted by the local node ) and allow it to be processed immediately

PR #1373, opened 2022-05-28
Update multi-coin initial transactions to accommodate balance clearing

jeffgrunewald:
Updates `txn_payments_v2` in order to support the balance clearing functionality of an account.
This change also cleans up some of the balance clearing logic previously implemented.

### DIFF
--- a/test/blockchain_payment_v2_SUITE.erl
+++ b/test/blockchain_payment_v2_SUITE.erl
@@ -528,7 +528,7 @@ invalid_balance_clearing_test(Config) ->
 
     ct:pal("~s", [blockchain_txn:print(SignedTx)]),
 
-    {error, {invalid_txns, [{BadTx, invalid_transaction}]}} = test_utils:create_block(
+    {error, {invalid_txns, [{BadTx, validation_failed}]}} = test_utils:create_block(
         ConsensusMembers,
         [SignedTx]
     ),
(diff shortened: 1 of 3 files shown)


test/blockchain_payment_v2_SUITE.erl
```diff
@@ -528,7 +528,7 @@ invalid_balance_clearing_test(Config) ->
 
     ct:pal("~s", [blockchain_txn:print(SignedTx)]),
 
-    {error, {invalid_txns, [{BadTx, invalid_transaction}]}} = test_utils:create_block(
+    {error, {invalid_txns, [{BadTx, validation_failed}]}} = test_utils:create_block(
```
vihu, line 531: I wanted to update this in the draft PR, but we should match on the exact failure when doing `txn:is_valid`

test/blockchain_payment_v2_SUITE.erl
```diff
@@ -566,7 +566,7 @@ balance_clearing_disabled_test(Config) ->
...
```
vihu, line 569: Same as above, match on the exact failure

test/blockchain_payment_v2_SUITE.erl
jeffgrunewald: bubbling up the full validation result as in the v2 check

PR #1374, opened 2022-05-30
telemetry events around block and txn process

jeffgrunewald:
Introduce additional instrumentation metrics into the blockchain core library related to transaction validation and block absorption, state channel opening and closing, snapshot generation and loading, and the transaction manager process's cache size and processing times

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -1230,10 +1230,11 @@ process_snapshot(ConsensusHash, MyAddress, Signers,
                                 Infos = blockchain_ledger_snapshot_v1:get_infos(Blockchain),
                                 SnapStart = erlang:monotonic_time(millisecond),
                                 case blockchain_ledger_snapshot_v1:snapshot(Ledger, Blocks, Infos) of
-                                    {ok, #{version := SnapVersion} = Snap} ->
+                                    {ok, Snap} ->
+                                        SnapVersion = blockchain_ledger_snapshot_v1:version(Snap),
                                         case blockchain_ledger_snapshot_v1:hash(Snap) of
                                             ConsensusHash ->
-                                                {ok, _, ConsensusHash, Size} = SnapResult = add_snapshot(Snap, ConsensusHash, Blockchain),
+                                                {ok, {_, ConsensusHash, Size}} = SnapResult = add_snapshot(Snap, ConsensusHash, Blockchain),
                                                 telemetry:execute([blockchain, snapshot, generate], #{duration => erlang:monotonic_time(millisecond) - SnapStart, size => Size},
                                                                                                     #{blocks => length(Blocks), mode => delayed, version => SnapVersion}),
                                                 SnapResult;
@@ -1243,7 +1244,7 @@ process_snapshot(ConsensusHash, MyAddress, Signers,
                                                 case application:get_env(blockchain, save_bad_snapshot, false) of
                                                     true ->
                                                         lager:info("saving bad snapshot ~p", [OtherHash]),
-                                                        {ok, _, OtherHash, Size} = BadResult = add_snapshot(Snap, OtherHash, Blockchain),
+                                                        {ok, {_, OtherHash, Size}} = BadResult = add_snapshot(Snap, OtherHash, Blockchain),
                                                         telemetry:execute([blockchain, snapshot, generate], #{duration => erlang:monotonic_time(millisecond) - SnapStart, size => Size},
                                                                                                             #{blocks => length(Blocks), mode => delayed, version => SnapVersion}),
                                                         BadResult;
@@ -1944,7 +1945,7 @@ missing_block(#blockchain{db=DB, default=DefaultCF}) ->
     end.
 
 -spec add_snapshot(blockchain_ledger_snapshot:snapshot(), blockchain()) ->
-    {ok, pos_integer(), binary(), pos_integer()} | {error, any()}.
+    {ok, {pos_integer(), binary(), pos_integer()}} | {error, any()}.
 add_snapshot(Snapshot, Chain) ->
     try
         Hash = blockchain_ledger_snapshot_v1:hash(Snapshot),
@@ -1955,7 +1956,7 @@ add_snapshot(Snapshot, Chain) ->
     end.
 
 -spec add_snapshot(blockchain_ledger_snapshot:snapshot(), binary(), blockchain()) ->
-    {ok, pos_integer(), binary(), pos_integer()} | {error, any()}.
+    {ok, {pos_integer(), binary(), pos_integer()}} | {error, any()}.
 add_snapshot(Snapshot, Hash, #blockchain{db=DB, snapshots=SnapshotsCF}=Chain) ->
     try
         Height = blockchain_ledger_snapshot_v1:height(Snapshot),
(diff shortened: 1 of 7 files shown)


src/transactions/blockchain_txn_mgr.erl
```diff
@@ -735,13 +745,15 @@ rejected(TxnKey, Txn, Member, Dialer, CurBlockHeight, RejectF) ->
 %% txn has exceeded the max number of rejections
 %% delete it and invoke callback
 -spec reject_actions(cached_txn_type(), integer(), integer()) -> ok.
-reject_actions({TxnKey, _Txn, #txn_data{callback = Callback, dialers = Dialers, rejections = _Rejections}},
+reject_actions({TxnKey, Txn, #txn_data{callback = Callback, dialers = Dialers, recv_block_height = RecvBlockHeight, rejections = Rejections}},
```
jadeallenx, line 748: Think this needs to be rebased on #1216

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -751,50 +763,55 @@ reject_actions({TxnKey, Txn, TxnData},
...
 
--spec cached_txns()-> [cached_txn_type()].
-cached_txns()->
+-spec cached_txns() -> [cached_txn_type()].
+cached_txns() ->
     ets:tab2list(?TXN_CACHE).
 
+-spec cache_size() -> non_neg_integer().
+cache_size() ->
+    ets:info(?TXN_CACHE, size).
```
jadeallenx, line 806: Wary of this metric - the units it uses are weird

src/transactions/blockchain_txn_mgr.erl
```diff
@@ -751,50 +763,55 @@ reject_actions({TxnKey, Txn, TxnData},
...
+
 -spec sorted_cached_txns()-> [] | [cached_txn_type()].
-sorted_cached_txns()->
+sorted_cached_txns() ->
```
jadeallenx, line 809: Should we just used an ordered table and avoid the overhead of sorting the list manually?

jadeallenx (approved):
Generally +1 on this but I think it needs a rebase for the new txn submission stuff

src/transactions/blockchain_txn_mgr.erl
jeffgrunewald: Yeah, that’ll be a bit of an effort. Probably lean on getting 1216 merger and then sorting through the rebase in case there are any final changes to that one

evanmcc (approved):

src/blockchain.erl
```diff
@@ -2657,8 +2664,9 @@ init_blessed_snapshot(Blockchain, _HashAndHeight={Hash, Height0}) when is_binary
             case get_snapshot(Hash, Blockchain) of
                {ok, BinSnapOrFile} ->
                   case blockchain_ledger_snapshot_v1:deserialize(BinSnapOrFile) of
-                      {ok, Snap} ->
+                      {ok, #{version := Version} = Snap} ->
```
xandkar, line 2667: Same here - blockchain_ledger_snapshot_v1:version/1` accessor.

src/blockchain.erl
```diff
@@ -1954,7 +1961,7 @@ add_snapshot(Snapshot, Hash, #blockchain{db=DB, snapshots=SnapshotsCF}=Chain) ->
         ok = rocksdb:write_batch(DB, Batch0, []),
         BinSnap = blockchain_ledger_snapshot_v1:serialize(Snapshot),
         case add_bin_snapshot(BinSnap, Height, Hash, Chain) of
-            ok -> {ok, Height, Hash};
+            ok -> {ok, Height, Hash, byte_size(BinSnap)};
```
xandkar, line 1964: 2 things:
1. specifically here - why bundle the size instead of just computing it at the one spot it is needed?
2. generally - better to take the opportunity to move towards a composable result type: `-type result(A, B) :: {ok, A} | {error, B}`, so `{ok, {Height, Hash, Size}}` in this case.

src/blockchain_worker.erl
```diff
@@ -525,6 +526,12 @@ handle_call({install_snapshot, Height, Hash, Snapshot, BinSnap}, _From,
             end,
             blockchain_lock:release(),
             true = ets:insert(?CACHE, {?CHAIN, NewChain}),
+            {SnapSource, ByteSize} = case BinSnap of
+                                         {file, Filename} -> {file, filelib:file_size(Filename)};
+                                         Bin when is_binary(Bin) -> {binary, byte_size(Bin)}
+                                     end,
+            telemetry:execute([blockchain, snapshot, load], #{duration => erlang:monotonic_time(millisecond) - LoadStart, size => ByteSize},
+                                                            #{height => Height, hash => Hash, version => Version, source => SnapSource}),
```
xandkar, line 534: This chunk is exactly the same as the one in `blockchain.erl` - maybe an opportunity to dedup.

src/state_channel/blockchain_state_channels_worker.erl
```diff
@@ -123,6 +123,8 @@ init(Args) ->
     {_, OwnerSigFun} = Owner,
     ok = blockchain_event:add_handler(self()),
     lager:info("started ~p", [blockchain_state_channel_v1:name(SC)]),
+    telemetry:execute([blockchain, state_channel, open], #{amount => Amount, time => erlang:monotonic_time(millisecond)},
+                                                         #{id => ID, sc_version => SCVer}),
```
xandkar, line 127: Are the metadata fields ad hoc for each of these `telemetry:execute` calls or is there a broader design codified somewhere?

xandkar (commented):
Not necessarily this time, but later it would be nice to abstract most of these telemetry calls. The duration one could also be wrapped in a macro which automatically inserts current module and function names. Maybe something like:

```erlang
-define(APPLY_TIMED(F)
    blockchain_metrics:apply_timed(F, ?MODULE, ?FUNCTION_NAME)
).
```
`blockchain_metrics:apply_timed` would manage the starting and subtracting time dance.

xandkar:
> Not necessarily this time, but later it would be nice to abstract most of these telemetry calls. The duration one could also be wrapped in a macro which automatically inserts current module and function names. Maybe something like:
> 
> ```erlang
> -define(APPLY_TIMED(F)
>     blockchain_metrics:apply_timed(F, ?MODULE, ?FUNCTION_NAME)
> ).
> ```
> 
> `blockchain_metrics:apply_timed` would manage the starting and subtracting time dance.

and, `blockchain_metrics` could probably encapsulate handling the cause of the annoying flood of warning messages during unit test runs.

src/blockchain.erl
jeffgrunewald: - for 1 we're already serializing the snapshot within the `add_snapshot/3` function, so rather than doing a potentially expensive binary serialization a second time at the call site I figured I'd save the overhead and computing time and use the already serialized version in the single function.
- the composable type is fine; i didn't really think about it either way since we'd already crossed that line with a 3-tuple

src/state_channel/blockchain_state_channels_worker.erl
jeffgrunewald: this is a first pass of potentially useful metadata to index and categorize metrics on; we'd shared a google doc a couple of times with a request for "what would be useful to further operationalize these services" but I don't know that anyone with more context ever filled it out.

src/transactions/blockchain_txn_mgr.erl
jeffgrunewald: what units does it use? looks like it should just be the count of items inserted into the table which would give us a metrics for the size of the cache and track when it grows beyond a certain threshold that might indicate a backlog building.

PR #1378, opened 2022-06-02
ensure after filtering we have max active gws up to limit

andymck:
This improves the filtering and limit application of h3dex targeting.  The problem is that we were first limiting the pool of GWs in the hex to a max count equal to chain var `poc_witness_consideration_limit` and then filting any inactives out of those.

This could result in a final pool of GWs well below the max limit value, should any inactive GWs be included in the deterministic subset result when applying the limit.

Now in this scenario if after performing the subset and filtering should the count of GWs be less than the limit value we then add additional active GWs to the subset should they be available.

### DIFF
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -11,7 +11,7 @@
 
 -export([
     shuffle_from_hash/2,
-    shuffle/1,
+    shuffle/1, shuffle/2,
     rand_from_hash/1, rand_state/1,
     normalize_float/1,
     challenge_interval/1,
@@ -58,7 +58,9 @@
     var_cache_stats/0,
     teardown_var_cache/0,
     init_var_cache/0,
-    target_v_to_mod/1
+    target_v_to_mod/1,
+    is_gw_active/3,
+    max_activity_age/1
 
 ]).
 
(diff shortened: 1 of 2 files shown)


src/blockchain_utils.erl
```diff
@@ -122,6 +124,10 @@ shuffle_from_hash(Hash, L) ->
 shuffle(Xs) ->
     [X || {_, X} <- lists:sort([{rand:uniform(), X} || X <- Xs])].
 
+-spec shuffle(rand:state(), [A]) -> [A].
+shuffle(RandState, Xs) ->
+    [X || {_, X} <- lists:sort([{rand:uniform_s(RandState), X} || X <- Xs])].
```
evanmcc, line 129: I do not think this works correctly:

```erlang
10> S =  rand:seed(exsss).
{#{bits => 58,jump => #Fun<rand.3.92093067>,
   next => #Fun<rand.0.92093067>,type => exsss,
   uniform => #Fun<rand.1.92093067>,
   uniform_n => #Fun<rand.2.92093067>},
 [246911442441078211|255457969820438583]}
11> [X || {_, X} <- lists:sort([{rand:uniform_s(S), X} || X <- lists:seq(1, 10)])].
[1,2,3,4,5,6,7,8,9,10]
```

a working oneliner is:
```
element(2, lists:unzip(lists:sort(element(1, lists:foldl(fun(X, {A, St}) -> {R, St1} = rand:uniform_s(St), {[{R, X} | A ], St1} end, {[], rand:seed(exsss)}, lists:seq(1, 10)))))).
``` but that's kind of dramatically ugly

andymck: doh, fixed here https://github.com/helium/blockchain-core/pull/1378/commits/d3fe951e5d5b61c7e0d4862bb203b4eb5bcde2de

PR #1379, opened 2022-06-02
use old ledger when determining target

andymck:
During receipt v2 validations, the incorrect ledger was being used when determining the target.  This could have resulted in valid receipts being rejected and is the probable cause for most of the "receipt not in order" errors.

thanks to community member BigEnigma for spotting this

NOTE: i have gated this behind the h3dex targeting chain var.  The fix will come into use once h3dex is enabled


madninja:
> :) me again, sorry i noticed you applied this such that it would only kick in for target V6 and above, was this because prior to that we cannot reliable replay anything should anyone do so? this ones a curiosity question in just how to go about managing versioned logic against version transactions and data models :)

That's correct.. If we don't do this as part of v6 then replaying the chain would cause it to use newer code on older blocks which would be consensus breaking

antonyscerri:
Hi

Did my first question about the variable used to get the targeting version make sense, is that bit ok going forward if we get to v7 but something gets an old txn from before a chain var switch should it not use v6 targeting instead of the current configured v7?

BigE

BigEnigma:
Hi

Did my first question about the variable used to get the targeting version make sense, is that bit ok going forward if we get to v7 but something gets an old txn from before a chain var switch should it not use v6 targeting instead of the current configured v7?

BigE

PR #1380, opened 2022-06-02
Fix the core test suites for great confidence

jeffgrunewald:
- [X] Dialyzer builds green
- [X] Eunit suites green
- [ ] CT suites green


Vagabond (approved):
Propose we merge this just to make things incrementally better

PR #1381, opened 2022-06-03
try to streamline poc v6 targeting

evanmcc:
this could probably be simplified a bit more, but there is time pressure.

### DIFF
--- a/src/poc/blockchain_poc_target_v6.erl
+++ b/src/poc/blockchain_poc_target_v6.erl
@@ -14,7 +14,6 @@
 
 -export([
     target_zone/2,
-    gateways_for_zone/5,
     target/5
 ]).
 
--- a/src/poc/blockchain_poc_target_v6.erl
+++ b/src/poc/blockchain_poc_target_v6.erl
@@ -39,7 +39,7 @@ target_zone(RandState, Ledger) ->
     Vars :: map(),
     HexList :: [h3:h3_index()],
     Attempted :: [{h3:h3_index(), rand:state()}]
-) -> {ok, [libp2p_crypto:pubkey_bin()]} | {error, any()}.
+) -> {ok, libp2p_crypto:pubkey_bin(), rand:state()} | {error, any()}.
 gateways_for_zone(
     ChallengerPubkeyBin,
     Ledger,
@@ -57,7 +57,7 @@ gateways_for_zone(
     HexList :: [h3:h3_index()],
     Attempted :: [{h3:h3_index(), rand:state()}],
     NumAttempts :: integer()
-) -> {ok, [libp2p_crypto:pubkey_bin()]} | {error, any()}.
+) -> {ok, libp2p_crypto:pubkey_bin(), rand:state()} | {error, any()}.
 gateways_for_zone(
     _ChallengerPubkeyBin,
     _Ledger,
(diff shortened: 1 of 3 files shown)


src/poc/blockchain_poc_target_v6.erl
```diff
@@ -131,80 +125,74 @@ target(ChallengerPubkeyBin, InitTargetRandState, ZoneRandState, Ledger, Vars) ->
...
-
-        {ok, []} ->
+    case gateways_for_zone(
+           ChallengerPubkeyBin,
+           Ledger, Vars, HexList,
+           [{InitHex, InitTargetRandState}]) of
+        {error, no_gateways_found} ->
+            {error, no_gateways_found};
+        {error, empty_hex_list} ->
             {error, no_gateways_found};
```
vihu, line 145: change this error msg?

jeffgrunewald (approved):

src/poc/blockchain_poc_target_v6.erl
jeffgrunewald: `no_valid_gateways_found`?

src/poc/blockchain_poc_target_v6.erl
```diff
@@ -75,20 +75,15 @@ gateways_for_zone(
...
     %% Limit max number of potential targets in the zone
-    {HexRandState, AddrList} = limit_addrs(Vars, HexRandState0, AddrList0),
-
-    %% filter the selected hex
-    case filter(AddrList, Height, Ledger, Vars) of
-        FilteredList when length(FilteredList) >= 1 ->
-            lager:debug("*** filtered gateways for hex ~p: ~p", [Hex, FilteredList]),
-            {ok, FilteredList};
-        _ ->
+    case find_active_addr(Vars, HexRandState0, AddrList0, Ledger) of
```
andymck, line 83: I am confused by this change with the use of find_active_addr,  gateways_for_zone should return a list of active gateways in the target hex ( where length of list is determined by the limit value).  This is currently returning a single gateway....Maybe confused this API with selecting the target within the selected hex?

src/poc/blockchain_poc_target_v6.erl
```diff
@@ -131,80 +125,74 @@ target(ChallengerPubkeyBin, InitTargetRandState, ZoneRandState, Ledger, Vars) ->
...
     Vars,
     HexList,
-    InitHex,
-    InitHexRandState
+    InitHex
 ) ->
-    case gateways_for_zone(ChallengerPubkeyBin,
-        Ledger, Vars, HexList, [{InitHex, InitHexRandState}]) of
-
-        {ok, []} ->
```
andymck, line 149: If gateways_for_zone returns anything other than {ok, List}, Sibyl will need to be updated to accommodate and will need to handle both the {ok, List} and the newer {error, Reason} response

https://github.com/helium/sibyl/blob/master/src/sibyl_poc_mgr.erl#L163

If we want to settle on {error, Reason} response, then the v5 targeting should also be updated so that we can keeps the APIs consistent and support one format in sibyl

src/poc/blockchain_poc_target_v6.erl
Vagabond: I removed the dependence on this api in helium/sibyl#58

Vagabond:
This seems to re-sync fine

PR #1385, opened 2022-06-07
Don't invalidate h3->region lookups on var nonce change

Vagabond:
Previously all the h3->region lookups would be invalidated when any
chain var was changed. This was very expensive when unrelated vars
changed.

This patch changes things so that the lookup is keyed by the hash of the
h3 region binaries, so as long as those do not change the cache should
not be invalidated. It also adds caching for the h3 region binary hashes
and for the h3 binaries themselves.

### DIFF
--- a/src/region/blockchain_region_v1.erl
+++ b/src/region/blockchain_region_v1.erl
@@ -38,21 +38,28 @@ get_all_regions(Ledger) ->
 -spec get_all_region_bins(Ledger :: blockchain_ledger_v1:ledger()) ->
     {ok, [{atom(), binary() | {error, any()}}]} | {error, any()}.
 get_all_region_bins(Ledger) ->
-    case get_all_regions(Ledger) of
-        {ok, Regions} ->
-            Map = lists:foldl(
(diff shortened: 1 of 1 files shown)


src/region/blockchain_region_v1.erl
```diff
@@ -38,21 +38,28 @@ get_all_regions(Ledger) ->
...
+                              fun(Reg, Acc) ->
+                                      case blockchain:config(Reg, Ledger) of
+                                          {ok, Bin} ->
+                                              [{Reg, Bin}|Acc];
+                                          _ ->
+                                              [{error, {region_var_not_set, Reg}}|Acc]
+                                      end
+                              end, [], Regions),
+                      {ok, lists:reverse(Map)};
+                  Error ->
```
jeffgrunewald, line 59: do we want to cache this error when it happens or lift the initial case statement outside of the call to `e2qc:cache/3`?

evanmcc (approved):
I think this should do what we want it to without too much extra cost.

vihu (approved):

PR #1388, opened 2022-06-08
make sure we're running the latest version of the min otp22

jeffgrunewald:


jeffgrunewald:
superseded when we moved to otp 24

PR #1393, opened 2022-06-14
Add validation for stale heartbeats

PaulVMo:
This change proposes adding a new validation to validator heartbeat transactions to ensure they are not older than the heartbeat interval + grace period. This change will help ensure that 1) validators who are significantly behind the blockchain (i.e. still syncing) do not submit PoC proposals or gateway reactivations and 2) PoC public keys stored in the ledger do not outlive the private keys stored locally by the validator (and thus causing invalid PoCs).

Currently, heartbeats are validated 1) to make sure they are not ahead of the consensus group validating the transaction and 2) to make sure they are not too soon after the previous heartbeat for that validator. The proposed change in this PR adds another check to make sure that the transaction height is not too old, using the heartbeat interval + grace period as the maximum age. While the miner includes a check to not submit a heartbeat if still synching, this is not currently enforced on chain.

This is likely an edge case, as most validator heartbeats are added to the chain within 2-3 blocks (see below for stats). However, during times of heavily load or in the case of an improperly functioning validator, the difference between the heartbeat transaction height and the height at which it is absorbed, may be significant. The table below shows the count of heartbeat transactions grouped by the difference between the chain height and the heartbeat transaction height in blocks for the last 100000 heartbeats.  
```
Diff.   Count
2       45541
3       33080
4       11860
5        4341
6        1749
7         854
8         433
9         321
10        229
11-150    117
151+       94
```

### DIFF
--- a/src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
+++ b/src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
@@ -131,17 +131,26 @@ is_valid(Txn, Chain) ->
                         ok;
                     _ -> throw(unsupported_txn)
                 end,
-                %% make sure that this validator exists and is staked
+                %% make sure that this validator exists and is staked, and that the transaction height is
+                %% not too soon after previous HB, not in the future, and (if chain var set) not too old
                 case blockchain_ledger_v1:get_validator(Validator, Ledger) of
                     {ok, V} ->
                         {ok, Interval} = ?get_var(?validator_liveness_interval, Ledger),
+                        {ok, GracePeriod} = ?get_var(?validator_liveness_grace_period, Ledger),
+                        NotStaleCheck =
+                            %% if chain var set, check for too old of heartbeat
+                            case ?get_var(?validator_stale_heartbeat_check, Ledger) of
+                                {ok, true} -> TxnHeight > (Height - (Interval + GracePeriod));
+                                _ -> true
+                            end,
                         Status = blockchain_ledger_validator_v1:status(V),
                         HB = blockchain_ledger_validator_v1:last_heartbeat(V),
                         case Status == staked
-                            andalso TxnHeight >= (Interval + HB)
-                            andalso TxnHeight =< Height of
+                            andalso TxnHeight >= (Interval + HB)   %% not too soon after last HB
+                            andalso TxnHeight =< Height            %% not in the future
+                            andalso NotStaleCheck of               %% not too old
                             true -> ok;
-                            _ -> throw({bad_height, prev, HB, height, Height, got, TxnHeight})
+                            _ -> throw({bad_height, previous_hb, HB, current_height, Height, got, TxnHeight})
                         end;
                     {error, not_found} -> throw(nonexistent_validator);
                     {error, Reason} -> throw({validator_fetch_error, Reason})
--- a/test/blockchain_validator_SUITE.erl
+++ b/test/blockchain_validator_SUITE.erl
@@ -20,7 +20,9 @@
     unstake_fail_already_unstaked/1,
     transfer_ok_in_account/1,
     transfer_ok/1,
-    unstake_ok_at_same_height/1
+    unstake_ok_at_same_height/1,
+    heartbeat_ok/1,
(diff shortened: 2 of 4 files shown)


evanmcc (commented):
This looks good but needs a test.

PaulVMo:
thanks evan. let me know if this tests are what you had in mind.

PaulVMo:
thanks evan. let me know if this tests are what you had in mind.

test/blockchain_validator_SUITE.erl
```diff
@@ -511,3 +516,129 @@ unstake_ok_at_same_height(Config) ->
...
+    {ok, B} = test_utils:create_block(Consensus, []),
+    _ = blockchain:add_block(B, Chain), %% new height = 27
+
+    %% confirm that heartbeat is invalid with txn height = 12, current height = 28
+    ?assertEqual(blockchain:height(Chain), {ok, 27}),
+    ?assertEqual(blockchain_txn_validator_heartbeat_v1:height(SignedHeartbeatTxn),12),
+    case blockchain_txn_validator_heartbeat_v1:is_valid(SignedHeartbeatTxn, Chain) of
+        ok -> ct:fail("expect invalid heartbeat, got valid one");
+        _Error -> ok
+    end.
```
xandkar, line 644:
```suggestion
    ?assertMatch(
        {error, _},
        blockchain_txn_validator_heartbeat_v1:is_valid(SignedHeartbeatTxn, Chain)
    ).
```

And even better than this would be to include the expected reason for the error.

test/blockchain_validator_SUITE.erl
```diff
@@ -511,3 +516,129 @@ unstake_ok_at_same_height(Config) ->
...
+        OwnerPubkeyBin,
+        ?bones(10000),
+        ?bones(5)
+    ),
+    SignedStakeTxn = blockchain_txn_stake_validator_v1:sign(Txn, OwnerSigFun),
+    {ok, Block} = test_utils:create_block(Consensus, [SignedStakeTxn]),
+    ok = blockchain:add_block(Block, Chain), 
+
+    %% add HB interval (10 blocks) to chain before heartbeating
+    _ = lists:map(
```
xandkar, line 545:
```suggestion
    _ = lists:foreach(
```

src/transactions/v1/blockchain_txn_validator_heartbeat_v1.erl
```diff
@@ -131,17 +131,28 @@ is_valid(Txn, Chain) ->
...
+                                _ -> false
+                            end,
                         Status = blockchain_ledger_validator_v1:status(V),
                         HB = blockchain_ledger_validator_v1:last_heartbeat(V),
                         case Status == staked
-                            andalso TxnHeight >= (Interval + HB)
-                            andalso TxnHeight =< Height of
+                            andalso TxnHeight >= (Interval + HB)   %% not too soon after last HB
+                            andalso TxnHeight =< Height            %% not in the future
+                            andalso case StaleHeartbeatCheck of    %% not too old
```
Vagabond, line 150: I don't love the `andalso case`, it might be nicer to bind a var above and check it here.

PaulVMo: how about something like this? i don't totally love the var name choices now so open to more feedback.
```
                        NotStaleCheck =
                            %% if chain var set, check for too old of heartbeat
                            case blockchain_ledger_v1:config(?validator_stale_heartbeat_check, Ledger) of
                                {ok, true} -> TxnHeight > (Height - (Interval + GracePeriod));
                                _ -> true
                            end,
                        Status = blockchain_ledger_validator_v1:status(V),
                        HB = blockchain_ledger_validator_v1:last_heartbeat(V),
                        case Status == staked
                            andalso TxnHeight >= (Interval + HB)   %% not too soon after last HB
                            andalso TxnHeight =< Height            %% not in the future
                            andalso NotStaleCheck of         %% not too old
```

andymck (approved):

PaulVMo:
Resolved merge conflict and squashed including re-running ct suite for validator heartbeats.

evanmcc (approved):

PR #1396, opened 2022-06-20
sc handler optimisations

andymck:
Provides optimisations/improvements of the state channel GRPC handler and the shared `blockchain_state_channel_common` module.  This mod is shared across both grpc and libp2p transport layers for state channel requests.

Optimisation/improvements include:

- Remove cached versions of the `blockchain` and `ledger` in the state channel handler state
- Add a simple ETS cache to blockchain_worker to hold the current blockchain.  Avoids blocking calls to the worker by spawned grpc handlers.

### DIFF
--- a/src/blockchain_worker.erl
+++ b/src/blockchain_worker.erl
@@ -15,7 +15,9 @@
 %% ------------------------------------------------------------------
 -export([
     start_link/1,
+    make_ets_table/0,
     blockchain/0, blockchain/1,
+    cached_blockchain/0,
     num_consensus_members/0,
     consensus_addrs/0,
     integrate_genesis_block/1,
@@ -78,6 +80,9 @@
 -define(WEEK_OLD_SECONDS, 7*24*60*60). %% a week's worth of seconds
 -define(MAX_ATTEMPTS, 3).
 
+-define(CACHE, worker_cache).
+-define(CHAIN, chain).
+
 -ifdef(TEST).
 -define(SYNC_TIME, 1000).
 -else.
@@ -121,7 +126,29 @@
 %% API Function Definitions
 %% ------------------------------------------------------------------
 start_link(Args) ->
-    gen_server:start_link({local, ?SERVER}, ?SERVER, Args, [{hibernate_after, 5000}]).
+    Res = server:start_link({local, ?SERVER}, ?SERVER, Args, [{hibernate_after, 5000}]),
+    case Res of
+        {ok, Pid} ->
+            %% if we have an ETS table reference, give ownership to the new process
+            %% we likely are the `heir', so we'll get it back if this process dies
+            case proplists:get_value(ets_cache, Args, not_found) of
+                not_found ->
+                    %% should ever hit here but ok....
+                    ok;
+                Tab ->
+                    true = ets:give_away(Tab, Pid, undefined)
+            end;
+        _ ->
+            ok
+    end,
+    Res.
+
+
+make_ets_table() ->
+    ets:new(?CACHE,
+            [named_table,
+             protected,
+             {heir, self(), undefined}]).
 
 %%--------------------------------------------------------------------
 %% @doc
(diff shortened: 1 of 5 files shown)


src/blockchain_worker.erl
```diff
@@ -387,6 +422,7 @@ init(Args) ->
                      [ libp2p_swarm:listen(SwarmTID, Addr) || Addr <- ListenAddrs ]),
     NewState = #state{swarm_tid = SwarmTID, blockchain = Blockchain,
                 gossip_ref = Ref},
+    true = ets:insert(?CACHE, {?CHAIN, Blockchain}),
```
jeffgrunewald, line 425: should be pretty easy to move this into a `handle_continue/2` so we load the chain into the cache table before receiving any other messages post-init. would it make sense (perhaps later) to consolidate these "blockchain worker cache items" into a single shared globally readable cache, owned and updated by the worker, and consolidate the "current height" cache that's in review to merge into the txn mgr? any others doing similar work?

andymck: we def need to get loading the chain out of the init, using handle_continue would work well there.  I dont want to address that in this PR as its dramatically increases the impact and thus the risk.

We can address in a follow up PR and that could also include replacing all blocking calls to the worker to return the chain with the new non blocking call.  Extending the cache to also include the height and additional data does make sense too.

src/blockchain_worker.erl
```diff
@@ -506,6 +545,7 @@ handle_call({install_aux_snapshot, Snapshot}, _From,
     {ok, GossipRef} = add_handlers(SwarmTID, NewChain),
     notify({new_chain, NewChain}),
     blockchain_lock:release(),
+    true = ets:insert(?CACHE, {?CHAIN, NewChain}),
```
evanmcc, line 548: don't think that we need this one here, this is for the aux ledger.

src/blockchain_worker.erl
```diff
@@ -538,16 +578,19 @@ handle_call({add_commit_hook, CF, HookIncFun, HookEndFun} , _From, #state{blockc
     Ledger = blockchain:ledger(Chain),
     {Ref, Ledger1} = blockchain_ledger_v1:add_commit_hook(CF, HookIncFun, HookEndFun, Ledger),
     Chain1 = blockchain:ledger(Ledger1, Chain),
+    true = ets:insert(?CACHE, {?CHAIN, Chain1}),
```
evanmcc, line 581: nor here

src/blockchain_worker.erl
```diff
@@ -538,16 +578,19 @@ handle_call({add_commit_hook, CF, HookIncFun, HookEndFun} , _From, #state{blockc
...
     {reply, Ref, State#state{blockchain = Chain1}};
 handle_call({add_commit_hook, CF, HookIncFun, HookEndFun, Pred} , _From, #state{blockchain = Chain} = State) ->
     Ledger = blockchain:ledger(Chain),
     {Ref, Ledger1} = blockchain_ledger_v1:add_commit_hook(CF, HookIncFun, HookEndFun, Pred, Ledger),
     Chain1 = blockchain:ledger(Ledger1, Chain),
+    true = ets:insert(?CACHE, {?CHAIN, Chain1}),
```
evanmcc, line 587: nor here

src/blockchain_worker.erl
```diff
@@ -538,16 +578,19 @@ handle_call({add_commit_hook, CF, HookIncFun, HookEndFun} , _From, #state{blockc
...
     {reply, Ref, State#state{blockchain = Chain1}};
 handle_call({remove_commit_hook, RefOrCF} , _From, #state{blockchain = Chain} = State) ->
     Ledger = blockchain:ledger(Chain),
     Ledger1 = blockchain_ledger_v1:remove_commit_hook(RefOrCF, Ledger),
     Chain1 = blockchain:ledger(Ledger1, Chain),
+    true = ets:insert(?CACHE, {?CHAIN, Chain1}),
```
evanmcc, line 593: nor here

src/blockchain_worker.erl
```diff
@@ -743,6 +787,7 @@ handle_info({'DOWN', RocksGCRef, process, RocksGCPid, Reason},
     {noreply, State#state{rocksdb_gc_mref = undefined}};
 
 handle_info({blockchain_event, {new_chain, NC}}, State) ->
+    true = ets:insert(?CACHE, {?CHAIN, NC}),
```
evanmcc, line 790: pretty sure this is duplicative

evanmcc (changes requested):
I'm worried about how many times we insert the chain into the cache.  I've added some that I think are duplicated, but I think that someone should add logging to each of these inserts and make sure that we're only calling it once on any given code path.  the important ones are:
1. Startup with existing chain
2. Load new genesis
3. load snapshot

andymck:
So the approach I took was to insert to the cache anytime the chain in state is being updated

PR #1397, opened 2022-06-20
Fix unclosed rocks iterators

xandkar:


src/blockchain.erl
```diff
@@ -892,12 +892,18 @@ find_first_height_after(MinHeight0, #blockchain{db=DB, heights=HeightsCF}) ->
...
+        case rocksdb:iterator_move(Iter, next) of
+            {ok, <<Height:64/integer-unsigned-big>>, Hash} ->
+                {ok, Height, Hash};
+            {error, invalid_iterator} ->
+                {error, not_found};
+            {error, Reason} ->
+                lager:error("Unexpected iterator error: ~p", [Reason]),
+                {error, not_found}
+        end,
+    catch rocksdb:iterator_close(Iter),
```
jeffgrunewald, line 905: thought we were trying to avoid these types of catches since they swallowed errors

xandkar: Yeah. I hate it. I've done a bunch of rocks iteration experiments recently and stumbled on enough seemingly-nondeterministic iterator-access errors to make me nervous.

While I lean on removing these `catch`s here, I'm not sure what will happen if you try to close an iterator after an error other than `invalid_iterator` - hence the recon logging.

I wasn't able to reproduce those not-`invalid_iterator` errors.

xandkar: @jeffgrunewald OK, I replaced all the `catch rocksdb:iterator_close`s with `try rocksdb:iterator_close catch log end`s.

jeffgrunewald (approved):
Love the macro!

src/blockchain_rocks.hrl
```diff
@@ -0,0 +1,10 @@
+-define(ROCKSDB_ITERATOR_CLOSE(IteratorHandle),
+    %% Wrapping in fun to avoid scope leakage:
+    (fun() ->
+        try
+            rocksdb:iterator_close(IteratorHandle)
+        catch ErrorClass:ErrorReason ->
+            lager:error("rocksdb:iterator_close error: ~p:~p", [ErrorClass, ErrorReason])
```
evanmcc, line 7: change to warning, please

evanmcc:
suspect that the catch logging is going to be noisy, we might want to add a version of the macro that allows us not to log.

in any case, we want to log at `warning`, since error writes to the error log with `fsync`

src/blockchain_rocks.hrl
xandkar: done.

xandkar:
> suspect that the catch logging is going to be noisy, we might want to add a version of the macro that allows us not to log.

added the an an app env var to disable the logging.

> in any case, we want to log at `warning`

switched it to warning.

> since error writes to the error log with `fsync`

TIL!

PR #1377, opened 2022-06-01
Add support for multitoken rewards and subnetworks

evanmcc:
Summary
----
This PR primarily adds support for adding subnetworks and subnetwork rewards.

Notes
----
- Add `blockchain_ledger_subnetwork_v1` module
- Modify `blockchain_ledger_v1` to support adding, updating and finding subnetworks.
- Add support for `blockchain_txn_add_subnetwork_v1:txn_add_subnetwork_v1`, `blockchain_txn_update_subnetwork_v1:txn_update_subnetwork_v1`, `blockchain_txn_subnetwork_rewards_v1:txn_subnetwork_rewards_v1` and `blockchain_txn_token_convert_v1:txn_token_convert_v1`.
- Add `allowed_reward_server_key_length` chain var
- Add subnetwork rewards test.

PR #1398, opened 2022-06-21
handling txn fees in balance clearing

jeffgrunewald:
This change should allow the proper handling of transaction fees when making a balance clearing transaction. In this case, the paying address's DC ledger entry balance is checked to cover the expected fee and there is a shortfall, attempts to make up the difference by implicit burning HNT. This is contrary to the previous behavior wherein only a DC balance or an implicit burn would be used to cover transaction fees.

This also properly handling converting the fee to HNT in bones when calculating the paying transaction's balance to deduct and send when making a balance clearing transaction. The previous incorrect behavior was to deduct the exact amount of the transaction fee from the payment amount.


jeffgrunewald:
Assuming this is merged I will apply the same necessary changes to the balance clearing payment handling in https://github.com/helium/blockchain-core/pull/1377

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3196,10 +3196,18 @@ debit_fee(Address, Fee, Ledger, MaybeTryImplicitBurn, TxnHash, Chain) ->
                     EntriesCF = dc_entries_cf(Ledger),
                     cache_put(Ledger, EntriesCF, Address, Bin);
                 {false, true} ->
-                    %% user does not have sufficient DC balance, try to do an implicit hnt burn instead
-                    {ok, FeeInHNT} = ?MODULE:dc_to_hnt(Fee, Ledger),
-                    ok = add_hnt_burned(FeeInHNT, Ledger),
-                    ?MODULE:debit_fee_from_account(Address, FeeInHNT, Ledger, TxnHash, Chain);
+                    Entry1 = blockchain_ledger_data_credits_entry_v1:new(
```
andymck, line 3199: this will need to be behind a var otherwise sync will be impacted for all fee based txns ?

andymck:
I implemented the txn fees and am pretty sure that it was deliberate at the time to not split the fee across DC and HNT but to do one or the other.  Is there a driver for this change ? I mean it probably is better to split it but I would like to understand the driver

src/ledger/v1/blockchain_ledger_v1.erl
jeffgrunewald: Crap; good call. I’ll add if needed but per your broader question I’ll wait to get confirmation that this is actually desired.

jeffgrunewald:
@andymck the drive for the change was just that I noticed it when I was fixing the way balance clearing txns handle deducting fees and the response seemed to be that it was a desirable change but I’m not sure how closely anyone has looked at it since I asked in the channel. Will wait for authoritative word but can easily back that part out if I was mistaken.

src/ledger/v1/blockchain_ledger_v1.erl
jeffgrunewald: per discussion in the community channel, this one-or-the-other approach is still preferred so reverting this split fee payment change. making a note for the future in case it's desirable to revisit.

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -392,12 +392,29 @@ split_payment_amounts(#blockchain_txn_payment_v2_pb{payer=Payer, fee=Fee}=Txn, L
         true ->
             {ok, PayerEntry} = blockchain_ledger_v1:find_entry(Payer, Ledger),
             Balance = blockchain_ledger_entry_v1:balance(PayerEntry),
-            MaxPaymentAmount = Balance - lists:sum(SpecifiedAmounts) - Fee,
+            MaxPaymentAmount = Balance - lists:sum(SpecifiedAmounts) - calculate_hnt_fee(Payer, Fee, Ledger),
```
andymck, line 395: not really familiar with split_payment_amounts function but why is fee changing from DC to HNT here ?

Also is there anything that needs considered as part of the absorb associated with this change ?

jeffgrunewald: The split payment amounts function will take all the payments in a v2 payments txn and split the balance clearing payment from the others (there can be no more than one) if it exists. Then it figures out what that payment should be in order to "clear the account balance" after deducting any other payments and any amount of HNT needed to cover the fee of the txn.

The fee is converted to HNT _iff_ the txn would require an implicit burn to ensure that we don't double-count HNT to cover the fee _and_ try to include it in the balance clearing payment (resulting in an invalid txn). If the account has enough dc to cover the fee the result in the line you've highlighted above evaluates to 0.

I'm not sure what else to consider in the txn absorb; it recomputes at the time of absorb what the split payment is, so it'll ensure that balance clearing payment is whatever's left over _after_ covering the fee

andymck: Ok, makes sense now

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -392,12 +392,29 @@ split_payment_amounts(#blockchain_txn_payment_v2_pb{payer=Payer, fee=Fee}=Txn, L
...
             {undefined, SpecifiedAmounts}
     end.
 
+-spec calculate_hnt_fee(Payer :: libp2p_crypto:pubkey_bin(), Fee :: non_neg_integer(), Ledger :: blockchain_ledger_v1:ledger()) -> non_neg_integer().
+calculate_hnt_fee(_, 0, _) -> 0;
+calculate_hnt_fee(Payer, Fee, Ledger) when Fee > 0 ->
+    case blockchain_ledger_v1:find_dc_entry(Payer, Ledger) of
+        {error, dc_entry_not_found} ->
+            {ok, AmountInHNT} = blockchain_ledger_v1:dc_to_hnt(Fee, Ledger),
+            AmountInHNT;
```
jeffgrunewald, line 407:
```suggestion
            {ok, FeeInHNT} = blockchain_ledger_v1:dc_to_hnt(Fee, Ledger),
            FeeInHNT;
```

evanmcc (approved):

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -569,7 +569,7 @@ calculate_hnt_fee(Payer, Fee, Ledger) when Fee > 0 ->
             FeeInHNT;
         {ok, Entry} ->
             DCBalance = blockchain_ledger_data_credits_entry_v1:balance(Entry),
-            case (DCBalance - Fee >= 0) of
+            case (DCBalance - Fee) >= 0 of
```
Vagabond, line 572: I believe the precedence rules here make the parentheses in both forms irrelevant?

src/transactions/v2/blockchain_txn_payment_v2.erl
```diff
@@ -183,6 +183,7 @@ absorb_v2_(Txn, Ledger, Chain) ->
     Nonce = ?MODULE:nonce(Txn),
     TotalAmounts = ?MODULE:total_amounts(Txn, Ledger),
     ShouldImplicitBurn = blockchain_ledger_v1:txn_fees_active(Ledger),
+    {MaxAmounts, _} = split_payment_amounts(Txn, Ledger),
```
Vagabond, line 186: Why was this hoisted if it's still used in only the one sub clause?

vihu (approved):

PR #1414, opened 2022-07-05
Await CLI sync stream response

jadeallenx:
Problem to solve: The cli sync command over optimistically assumes that sync starts once a stream is established.

Solution: We will block on a response from the sync stream, and report the outcome to the shell.


jadeallenx:
```_build/dev/rel/miner/bin/miner peer sync /p2p/115PmCR6fpFihdjw626JXYdUEdzwjh66yoWzWkMvB9CRGEx1U6G
ok```

Vagabond (approved):

PR #1419, opened 2022-07-07
add per-block limit to rewards

evanmcc:
to limit the impact of potential reward server takeovers, add a chain var to limit the amount of rewards that can be emitted per block.

### DIFF
--- a/include/blockchain_vars.hrl
+++ b/include/blockchain_vars.hrl
@@ -638,4 +638,4 @@
 %% How many reward server keys to allow
 -define(allowed_num_reward_server_keys, allowed_num_reward_server_keys).
 %% limit per-block payout of l2 tokens
--define(limit_subnetwork_payout, limit_subnetwork_payout).
+-define(subnetwork_reward_per_block_limit, subnetwork_reward_per_block_limit).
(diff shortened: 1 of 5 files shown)


include/blockchain_vars.hrl
```diff
@@ -637,4 +637,5 @@
 -define(deprecate_security_exchange_v1, deprecate_security_exchange_v1).
 %% How many reward server keys to allow
 -define(allowed_num_reward_server_keys, allowed_num_reward_server_keys).
-
+%% limit per-block payout of l2 tokesn
```
vihu, line 640:
```suggestion
%% limit per-block payout of l2 tokens
```

include/blockchain_vars.hrl
```diff
@@ -637,4 +637,5 @@
...
+-define(limit_subnetwork_payout, limit_subnetwork_payout).
```
vihu, line 641: Should we rename this to `subnetwork_payout_limit`?

src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1260,6 +1260,9 @@ validate_var(?allowed_num_reward_server_keys, Value) ->
         _ -> throw({error, {invalid_allowed_num_reward_server_keys, Value}})
     end;
 
+validate_var(?limit_subnetwork_payout, Value) ->
+    validate_int(Value, "limit_subnetwork_payout", 0, 10000000, false);
```
vihu, line 1264: Just wanted to double check on the max limit here, can we add a comment about it? Also is this supposed to be a `10^8` thing?

PR #1429, opened 2022-07-16
use full name for lorawan-lib dep

michaeldjeffrey:
and include in app.src


ke6jjj (approved):
I tested this branch on a miner that was experiencing a problem loading the module `lora_subnet` and subsequently crashing. This fixes it and doesn't seem to cause any other issues.

madninja (approved):

PR #1440, opened 2022-07-26
fix balance bug and add var

evanmcc:
this PR fixes a potential issue with token type v2 that could result in lost balances for non-hnt token types after the entry_v2 activation.


src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -3290,7 +3290,13 @@ debit_account(Address, AmountOrAmounts, Nonce, Ledger) when is_integer(AmountOrA
                     Balance = EntryMod:balance(Entry),
                     case (Balance - AmountOrAmounts) >= 0 of
                         true ->
-                            Entry1 = EntryMod:new( Nonce, (Balance - AmountOrAmounts)),
+                             Entry1 =
```
vihu, line 3293: There should be a guard to check whether `balance_erase_bugfix=true` somewhere here?

vihu (approved):
lgtm

Vagabond (approved):

PR #1442, opened 2022-07-27
Fix bug in h3dex targeting lookup call

PaulVMo:
Fixes a bug in the call to build_random_hex_targeting_lookup when adding or removing hexes from the h3dex.


evanmcc (approved):

Vagabond:
How did this ever work and is fixing it without a var safe?

PaulVMo:
I do not believe it needs a var. It worked because the only case that was
correct was the only one that actually got called. Two cases were protected
by a var already, and the third in the remove_gw wasn't ever called because
the parent hex is checked to see if it is populated BEFORE the gateway's
hex is actually removed from the h3dex - meaning it will always be
populated and never trigger a rebuild.

PR #1444, opened 2022-08-01
change cache from e2qc to cream

evanmcc:
e2qc had some locking, contention, and speed issues, this is an attempt to swap out the cache library.


rebar.lock
```diff
@@ -15,15 +15,15 @@
   {git,"https://github.com/helium/clique.git",
        {ref,"e4be0dae150061bec080502ed530091672880867"}},
   0},
+ {<<"cream">>,
+  {git,"https://github.com/JayKickliter/cream",
```
JayKickliter, line 19: We should probably transfer the repo to the helium org if everything looks good

JayKickliter:
Some context for reviewers: the new caching module is [cream](https://github.com/jaykickliter/cream), which is a NIF wrapper around [Moka](https://github.com/moka-rs/moka).

`e2qc`'s API has some inherent flaws that are not easily worked around: individual caches are referenced by an atom (necessitating a globally locked map holding all caches). Additionally, e2qc caches are lazily constructed when referenced by their atom name, which means that they use default parameters unless you choose provide cache parameters with every single caching call.

src/blockchain_sup.erl
```diff
@@ -143,3 +148,42 @@ init(Args) ->
 %% ------------------------------------------------------------------
 %% Internal Function Definitions
 %% ------------------------------------------------------------------
+
+
+cream_caches_init() ->
+    MaxCapacity = 10000,
+    {ok, VarCache}
+        = cream:new(MaxCapacity,
+                    [{initial_capacity, 1000},
```
Vagabond, line 157: Are these in byte sizes or in counts?

src/blockchain_utils.erl
```diff
@@ -432,17 +432,7 @@ hex_adjustment(Loc) ->
 -spec score_gateways(Ledger :: blockchain_ledger_v1:ledger()) -> gateway_score_map().
 score_gateways(Ledger) ->
     {ok, Height} = blockchain_ledger_v1:current_height(Ledger),
-    case blockchain_ledger_v1:mode(Ledger) of
```
Vagabond, line 435: No more cache here?

src/blockchain_utils.erl
```diff
@@ -812,36 +802,40 @@ get_vars(VarList, Ledger) ->
...
+      fun() ->
+              get_var_(VarName, Ledger)
+      end
     ).
 
 -spec get_var(VarName :: atom(), Ledger :: blockchain_ledger_v1:ledger()) -> {ok, any()} | {error, any()}.
 get_var(VarName, Ledger) ->
-    get_var_(VarName, Ledger).
+    VarsNonce =
+        case blockchain_ledger_v1:vars_nonce(Ledger) of
```
Vagabond, line 817: Did you ever get this to be a optimized lookup value?

src/blockchain_sup.erl
JayKickliter: item count. cream/main now has the option to bound the cache by memory

rebar.config
```diff
@@ -38,7 +38,7 @@
     {h3, ".*", {git, "https://github.com/helium/erlang-h3.git", {branch, "master"}}},
     {erl_angry_purple_tiger, ".*", {git, "https://github.com/helium/erl_angry_purple_tiger.git", {branch, "master"}}},
     {erlang_stats, ".*", {git, "https://github.com/helium/erlang-stats.git", {branch, "master"}}},
-    {e2qc, ".*", {git, "https://github.com/helium/e2qc", {branch, "master"}}},
+    {cream, ".*", {git, "https://github.com/JayKickliter/cream", {branch, "main"}}},
```
JayKickliter, line 41:
```suggestion
    {cream, ".*", {git, "https://github.com/helium/cream", {branch, "main"}}},
```

JayKickliter (changes requested):
After updating rebar.config to source cream from the helium or, please bump it so that you get the latest API. I don't think you're using anything that changed, but a couple things may break. The biggest change is that you can now bound the cache my [memory] usage.

[memory]: https://github.com/helium/cream/blob/main/README.md?plain=1#L56-L75

src/blockchain_sup.erl
evanmcc: object count, as I understand it

Vagabond:
I think it would be better to bound by memory and make the memory config tunable from the sys.config

JayKickliter:
> and make the memory config tunable from the sys.config

A global cache size config var or per-cache var?

rebar.lock
JayKickliter: It's been transferred.

Vagabond:
> > and make the memory config tunable from the sys.config
> 
> A global cache size config var or per-cache var?

I'd prefer a per-cache var

src/blockchain_sup.erl
```diff
@@ -151,21 +151,22 @@ init(Args) ->
 
 
 cream_caches_init() ->
-    MaxCapacity = 10000,
+    MaxItemCapacity = 10000,
+    MB = 1024 * 1024,
     {ok, VarCache}
-        = cream:new(MaxCapacity,
+        = cream:new(MaxItemCapacity,
                     [{initial_capacity, 1000},
```
JayKickliter, line 158: I added `initial_capacity` for completeness sake, but I'm not sure I'd bother setting it unless you have strong insight into growth. Reallocation is _probably_ fast enough compared to a cache miss.

src/grpc/blockchain_grpc_sc_server_handler.erl
```diff
@@ -115,17 +115,16 @@ maybe_initialize_state(undefined) ->
...
         {ok, 2} ->
             ActiveSCs =
-                e2qc:cache(
-                    ?MODULE,
-                    active_list,
-                   30,
-                    fun() -> maps:to_list(blockchain_state_channels_server:get_actives()) end
-                ),
+                cream:cache(
+                  ?sc_server_cache,
```
JayKickliter, line 124: Missing a `persistent_term:get(?sc_server_cache)`?

JayKickliter, line 124: Missing a pt:get

```suggestion
                Cache = persistent_term:get(?sc_server_cache),
                cream:cache(
                  Cache,
```

JayKickliter (changes requested):

src/blockchain_utils.erl
evanmcc: no, I think that this should be a separate PR, as it's a bit tricky to get right.

src/blockchain_utils.erl
evanmcc: I can't remember my logic here, I'll revert this.

Vagabond (approved):

PR #1447, opened 2022-08-04
Expose payment amounts when max=true

vihu:
Summary
----
This PR adds amount when `max=true` for payment_v2 transactions.
It requires that `ledger` be passed in when calling `to_json` for payment_v2 transactions implying that ETL needs to be upgraded to do the same.

TODO:
- [x] Test existing stuff
- [x] Extend json test


madninja:
Unfortunately this won’t work during submission to pending txns. There is no ledger available when submitting a pending payment txn and it needs some form of json to represent it to

madninja (approved):

PR #1448, opened 2022-08-06
Make block validation and absorb run in parallel

Vagabond:


src/transactions/blockchain_txn.erl
```diff
@@ -449,20 +449,27 @@ absorb_and_commit(Block, Chain0, BeforeCommit) ->
...
     Transactions = lists:sort(fun sort/2, (Transactions0)),
     Start = erlang:monotonic_time(millisecond),
-    case ?MODULE:validate(Transactions, Chain1, Rescue) of
+    Ref = make_ref(),
+    Parent = self(),
+    {AbsorbPid, AbsorbRef} = spawn_monitor(fun() ->
+                          Res = ?MODULE:absorb_block(Block, Rescue, Chain1),
+                          Parent ! {Ref, Res}
+                  end),
+    case ?MODULE:validate(Transactions, Chain0, Rescue) of
```
jeffgrunewald, line 465: If you’re passing `Chain0` in here doesn’t that mean the validation is happening with the version of the ledger before the `Ledger1` context and then you’re only ever committing `Ledger2` which came out of the result of the validation? 

It also looks like you’re doing `absorb_delayed_async/2` with `Chain0` instead of `Chain1`; is it okay to have those two absorbs using different versions of the chain?

evanmcc: Validation creates its own context (which is discarded on exit), and the delayed ledger is safe to write to in parallel, because it's all writing to different column families, so the exact version of the chain/ledger that's getting passed in to all these bits doesn't matter, or at least is correct as written in this PR.

evanmcc (approved):

src/transactions/blockchain_txn.erl
jeffgrunewald: Cool; thanks!

jeffgrunewald (approved):

PR #1449, opened 2022-08-08
Use a CREAM cache for valid witnesses, not pdict

Vagabond:
CREAM will evict naturally based on idle time and size, so use it to
replace the pdict cache that was never collected and would leak memory.


jeffgrunewald (approved):

PaulVMo:
Go a little bigger on the max size? Aspirationally if we have have 1000 PoC per block (3x current rate) and conservatively if epochs are a little long at 50 blocks, we could easily need 50,000 items in the cache.

evanmcc (approved):

PR #1450, opened 2022-08-08
Fix miner#1779 by reverting PR 1412

dpezely:
Fix https://github.com/helium/miner/issues/1779 by reverting https://github.com/helium/blockchain-core/pull/1412
per conversation with @mikev 

- The lora library call reporting as undefined was invoked from the mentioned PR
- Since that feature requires a future chain-var that has been put on hold, we can revert the feature by reversing its diff
- Later, we can track why that lora library call is no longer available; e.g., API changed?
- Preserved Pierre's additional fixes to pre-existing tests:
  + in `test/blockchain_state_channel_SUITE.erl`
 
```bash
wget https://github.com/helium/blockchain-core/pull/1412.diff
patch -p1 --reverse < 1412.diff
```


dpezely:
> We added the dep to the app file so this maybe doesn't need to be reverted?

Now, I'm really confused.  New tests wouldn't have passed prior to creating the original PR if there was such a missing dependency.

Everything with blockchain-core remain a bit alien to me since I don't follow the relevant Discord channels or its PRs closely enough.  Therefore, I'll leave that decision to others.  My rationale was that if this feature requires a chain var which has been put on hold, there's no reason for this feature code to exist (for now).

PR #1463, opened 2022-08-12
Fix security rewards for ledger_entry_v2

vihu:
- Add security_reward_bugfix chain var
- Guard security rewards behing security_reward_bugfix

### DIFF
--- a/src/transactions/v1/blockchain_txn_vars_v1.erl
+++ b/src/transactions/v1/blockchain_txn_vars_v1.erl
@@ -1271,7 +1271,7 @@ validate_var(?balance_erase_bugfix, Value) ->
 
 validate_var(?security_reward_bugfix, Value) ->
     case Value of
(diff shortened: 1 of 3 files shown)


src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1269,6 +1269,12 @@ validate_var(?balance_erase_bugfix, Value) ->
         _ -> throw({error, {invalid_balance_erase_bugfix, Value}})
     end;
 
+validate_var(?security_reward_bugfix, Value) ->
+    case Value of
+        Val when is_boolean(Val) andalso Val == true -> ok;
```
jeffgrunewald, line 1274:
```suggestion
        true -> ok;
```
seems like `true` is the only viable option here

jeffgrunewald (approved):

evanmcc (approved):
I like jeff's suggestion, otherwise this looks fine

PR #1465, opened 2022-08-14
Fix subnetwork reward txn hash

vihu:
Problem
----
Subnetwork reward txn hashes are not consistent because the signature field was still present when proto encoding the txn.

Solution
----
Unset the signature before calculating the txn hash.

Caveats
----
I _think_ this is largely okay and would not cause any chain related problems since we don't technically use the txn hash for any chain consistency. However, APIs may have a tough time querying the few subnetwork reward txns which are already on chain. AFAICT there's no workaround for that other than to manually fix the external database(s) where the inconsistent hashes were stored.

### DIFF
--- a/src/transactions/v1/blockchain_txn_subnetwork_rewards_v1.erl
+++ b/src/transactions/v1/blockchain_txn_subnetwork_rewards_v1.erl
@@ -65,7 +65,7 @@ new(Type, Start, End, Rewards) ->
 
 -spec hash(txn_subnetwork_rewards_v1()) -> blockchain_txn:hash().
 hash(Txn) ->
-    EncodedTxn = blockchain_txn_subnetwork_rewards_v1_pb:encode_msg(Txn),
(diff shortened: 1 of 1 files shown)


src/transactions/v1/blockchain_txn_subnetwork_rewards_v1.erl
```diff
@@ -264,6 +263,10 @@ to_json(Txn, _Opts) ->
         rewards => Rewards
     }.
 
+-spec base(Txn :: txn_subnetwork_rewards_v1()) -> txn_subnetwork_rewards_v1().
```
jeffgrunewald, line 266:
```suggestion
-spec base(Txn) -> Txn when Txn :: txn_subnetwork_rewards_v1().
```

jeffgrunewald (approved):

madninja (approved):

PR #1475, opened 2022-09-19
Allow for off-chain oracle poc receipts

Vagabond:


include/blockchain_vars.hrl
```diff
@@ -238,6 +238,9 @@
 %% whether or not to process validation of poc receipt witnesses
 -define(poc_receipt_witness_validation, poc_receipt_witness_validation).
 
+%% for off-chain poc, who should be signing receipts
+-define(poc_oracle_key, poc_oracle_key).
```
abhay, line 242: maybe an array so we can swap them as needed?

Vagabond: would you not just update the var, or are you proposing having a set of them?

abhay: proposing having a list of them so we can roll keys if we need to.

Vagabond: we could also use a multisig key and allow 1/M signing?

abhay: > we could also use a multisig key and allow 1/M signing?

not a bad idea but then we'd have to generate the keys in advance. i'm okay with whatever implementation you think is best.

Vagabond: I mean if it's a chain var we can change the "oracle public key" at any time to reflect keys being added/removed from the rotation?

rebar.config
```diff
@@ -40,7 +40,7 @@
     {erlang_stats, ".*", {git, "https://github.com/helium/erlang-stats.git", {branch, "master"}}},
     {cream, ".*", {git, "https://github.com/helium/cream", {branch, "main"}}},
     {vincenty, ".*", {git, "https://github.com/helium/vincenty", {branch, "master"}}},
-    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "master"}}},
+    {helium_proto, {git, "https://github.com/helium/proto.git", {branch, "adt/off-chain-poc-verification"}}},
```
jeffgrunewald, line 43: i think this has been merged so this needs updated which will probably fix the rebase conflict

src/transactions/v2/blockchain_txn_poc_receipts_v2.erl
```diff
@@ -156,20 +156,46 @@ is_valid(Txn, Chain) ->
...
-                            check_is_valid_poc(POCVersion, Txn, Chain)
+            {error, empty_path};
+        false ->
+            case libp2p_crypto:verify(EncodedTxn, Signature, PubKey) of
+                false ->
+                    {error, bad_signature};
+                true ->
+                    case {?get_var(?poc_challenger_type, Ledger), ?get_var(?poc_oracle_key, Ledger)} of
+                        {{ok, oracle}, {ok, OracleKey}} ->
+                            case OracleKey == Challenger of
```
jeffgrunewald, line 169: are we just setting the oracle signing key in the poc receipts coming from the oracle?

jeffgrunewald (approved):

vihu (approved):

PR #1476, opened 2022-09-20
implement stream of active gateways

jeffgrunewald:
implement a stream of records from the active gateways column family on the ledger in order to allow lazy streaming of large volumes of records out of the ledger downstream via blockchain-node or similar.

### DIFF
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -878,14 +878,18 @@ get_vars(VarList, Ledger) ->
                VarsNonce :: non_neg_integer(),
                Ledger :: blockchain_ledger_v1:ledger()) -> {ok, any()} | {error, any()}.
 get_var_(VarName, HasAux, VarsNonce, Ledger) ->
-    Cache = persistent_term:get(?var_cache),
-    cream:cache(
-      Cache,
-      {HasAux, VarsNonce, VarName},
-      fun() ->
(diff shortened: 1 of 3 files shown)


src/blockchain_utils.erl
```diff
@@ -878,14 +878,18 @@ get_vars(VarList, Ledger) ->
...
 get_var_(VarName, HasAux, VarsNonce, Ledger) ->
-    Cache = persistent_term:get(?var_cache),
-    cream:cache(
-      Cache,
-      {HasAux, VarsNonce, VarName},
-      fun() ->
-              get_var_(VarName, Ledger)
-      end
-    ).
+    try persistent_term:get(?var_cache) of
```
jeffgrunewald, line 881: fixes tests that blow up when no var cache has been created (triggered by `blockchain_ledger_v1:new/1`)

src/blockchain_utils.erl
```diff
@@ -908,9 +912,13 @@ var_cache_stats() ->
 
 -spec teardown_var_cache() -> ok.
 teardown_var_cache() ->
-    Cache = persistent_term:get(?var_cache),
-    cream:drain(Cache),
-    ok.
+    try persistent_term:get(?var_cache) of
```
jeffgrunewald, line 915: same deal as above

PR #1477, opened 2022-09-23
Don't store 0 security token rewards

Vagabond:


src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -952,9 +952,16 @@ securities_rewards(Ledger, #{epoch_reward := EpochReward,
...
                       Balance = blockchain_ledger_entry_v2:balance(Entry, hst),
-                      PercentofReward = (Balance*100/TotalSecurities)/100,
-                      Amount = erlang:round(PercentofReward*SecuritiesReward),
-                      maps:put({owner, securities, Key}, Amount, Acc)
+                      case Balance =< 0 andalso ?get_var(?security_zero_reward_bugfix, Ledger) == {ok, true} of
+                          true ->
+                              %% no security tokens AND the zero reward bugfix var is active
+                              Acc;
+                          false ->
+                              %% either a security reward, or the zero reward bugfix is not active
```
jeffgrunewald, line 960: Doesn’t this also presuppose the original `security_reward_bugfix` _is_ active since we don’t currently have the problem with that var unset?

Vagabond: yes, it's behind that var, so to enable this we'd have to set both vars

jeffgrunewald: so wouldn't we want the above to be something like:
```
case Balance =< 0 andalso (?get_var(?security_zero_reward_bugfix, Ledger) == {ok, true} andalso ?get_var(?security_reward_bugfix, Ledger) == {ok, true}) of
```
to guarantee both vars are set appropriately?

andymck: the security_reward_bugfix is checked prior at line 944

andymck (approved):

jeffgrunewald (approved):

vihu (approved):

PR #1485, opened 2022-10-24
Only snapshot deterministic h3dex keys, precalc on h3dex load

Vagabond:
This should fix the current snapshot agreement issues.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_v1.erl
@@ -6217,9 +6217,20 @@ load_pocs(_, PoCs, Ledger) ->
             CF :: rocksdb:cf_handle()
         }.
 snapshot_raw(CF, L) ->
+    snapshot_raw(CF, L, fun(_) -> true end).
+
+
+snapshot_raw(CF, L, Filter) ->
     %% XXX Preserve order to ensure that snapshots hash deterministically.
     %% Since rocks folds are lexicographic - we can just reverse:
-    lists:reverse(cache_fold(L, CF, fun({_, _}=KV, KVs) -> [KV | KVs] end, [])).
+    lists:reverse(cache_fold(L, CF, fun({_, _}=KV, KVs) ->
+                                            case Filter(KV) of
+                                                true ->
+                                                    [KV | KVs];
+                                                false ->
+                                                    KVs
+                                            end
+                                    end, [])).
 
 -spec load_raw([{binary(), binary()}] | function(), CFSpec, ledger()) -> ok when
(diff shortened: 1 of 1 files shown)


madninja:
how the heck did you figure out it was only those two keys!

Vagabond:
> how the heck did you figure out it was only those two keys!

a patched `snapshot diff`

Vagabond:
To elaborate, the `h3dex cache` was calculated at a point in time depending on when a validator upgraded, thus there was some drift. Some hexes, depending on when the validator upgraded, would or would not have a cache associated and the cache key would not be re-calculated until there was "activity" on the hex.

So the change in this PR is to exclude those non-deterministic keys from the snapshot, and force them to be recalculated on snapshot load.

Vagabond:
Also, I tested this with a drifted light gateway validator and was able to reconcile the drift with the new snapshots (there's still drift when the non-deterministic keys are not excluded).

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -6473,7 +6484,10 @@ snapshot_h3dex(Ledger) ->
     case config(?poc_targeting_version, Ledger) of
         {ok, N} when N >= 6 ->
             H3CF = h3dex_cf(Ledger),
-            snapshot_raw(H3CF, Ledger);
+            snapshot_raw(H3CF, Ledger, fun({<<"count-", _/binary>>, _}) -> false;
+                                           ({<<"gws-", _/binary>>, _}) -> false;
```
xandkar, line 6488: Would be good to leave a comment for posterity - why these two.

src/ledger/v1/blockchain_ledger_v1.erl
```diff
@@ -6217,9 +6217,20 @@ load_pocs(_, PoCs, Ledger) ->
...
+
+snapshot_raw(CF, L, Filter) ->
     %% XXX Preserve order to ensure that snapshots hash deterministically.
     %% Since rocks folds are lexicographic - we can just reverse:
-    lists:reverse(cache_fold(L, CF, fun({_, _}=KV, KVs) -> [KV | KVs] end, [])).
+    lists:reverse(cache_fold(L, CF, fun({_, _}=KV, KVs) ->
+                                            case Filter(KV) of
+                                                true ->
+                                                    [KV | KVs];
+                                            false ->
```
xandkar, line 6230:
```suggestion
                                                false ->
```

jadeallenx (approved):

andymck (approved):

PR #1487, opened 2022-10-29
Honor blockchain_txn:block_delay() when using ledger_at

Vagabond:
For some unknown reason, some nodes may have an older delayed ledger than the prescribed 50 blocks, this was leading to block disagreement when a transaction whose validity was bounded by ledger availability was included in a block and the validity needed a ledger 51+ blocks back.

This patch forces ledger_at to only ever return 50 blocks back, even if the lagging ledger is lagging further behind.

### DIFF
--- a/src/blockchain.erl
+++ b/src/blockchain.erl
@@ -583,6 +583,7 @@ ledger_at(Height, Chain0, ForceRecalc) ->
         _ ->
             Ledger0
     end,
+    Delay = blockchain_txn:block_delay(),
     case blockchain_ledger_v1:current_height(Ledger) of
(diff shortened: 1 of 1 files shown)


jeffgrunewald (approved):

src/blockchain.erl
```diff
@@ -614,7 +616,9 @@ ledger_at(Height, Chain0, ForceRecalc) ->
                                     Error
                             end
                     end;
-                {ok, DelayedHeight} when Height < DelayedHeight ->
+                {ok, _DelayedHeight} when Height < CurrentHeight - Delay ->
```
andymck, line 619: is this not likely to break sync ?

PR #1489, opened 2022-11-01
Use ledger checkpoint/snapshot at currentheight - block_delay for snaps

Vagabond:
Prior to this the default way to take a snapshot was to use the delayed ledger. However recent events has revealed delayed ledgers lagging too far behind and delayed ledger corruption. In the time since the delayed ledger was added the chain has switched to using rocksdb snapshots/checkpoints which offer a more reliable and less vulnerable to corruption source for snapshot data.

### DIFF
--- a/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
+++ b/src/ledger/v1/blockchain_ledger_snapshot_v1.erl
@@ -166,7 +166,13 @@
     {ok, snapshot()}
     | {error, killed | snapshot_error()}.
 snapshot(Ledger0, Blocks, Infos) ->
-    snapshot(Ledger0, Blocks, Infos, delayed).
+    %% by default use the ledger height - block_delay to emulate the lagging ledger but not actually
+    %% using the lagging ledger because that has recently been found subject to corruption or lagging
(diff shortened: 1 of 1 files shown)


src/ledger/v1/blockchain_ledger_snapshot_v1.erl
```diff
@@ -166,7 +166,13 @@
     {ok, snapshot()}
     | {error, killed | snapshot_error()}.
 snapshot(Ledger0, Blocks, Infos) ->
-    snapshot(Ledger0, Blocks, Infos, delayed).
+    %% by default use the ledger height - block_delay to emulate the lagging ledger but not actually
+    %% using the lagging ledger because that has recently been found subject to corruption or lagging
+    %% too far behind. Nodes that are reliably syncing will have a rocksdb checkpoint/snapshot at this
+    %% height which is safer to use.
```
jadeallenx, line 172: Hate to be that girl but what about nodes that have not been reliably syncing? Will this blow up for them?

Vagabond: no because they're guaranteed to have a ledger checkpoint at this height

jadeallenx (approved):

PR #1492, opened 2022-11-21
Teardown region cache on delayed ledger absorb

vihu:
# Summary

Sooner rather than later we'll have to issue some more region related vars, there's already some reports of region misalignment (Kazakhstan still on EU433 for example, instead of IN865). I **think** this change will help to hopefully avoid the delayed ledger cache being inconsistent on a region var change.

### DIFF
--- a/src/blockchain_utils.erl
+++ b/src/blockchain_utils.erl
@@ -63,6 +63,7 @@
     get_vars/2, get_var/2,
     var_cache_stats/0,
     teardown_var_cache/0,
+    teardown_region_cache/0,
     target_v_to_mod/1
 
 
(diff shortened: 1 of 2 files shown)


src/blockchain_utils.erl
```diff
@@ -912,6 +913,12 @@ teardown_var_cache() ->
     cream:drain(Cache),
     ok.
 
+-spec teardown_region_cache() -> ok.
+teardown_region_cache() ->
```
Vagabond, line 917: Isn't there already a teardown function in blockchain_region_v1?

Vagabond:
I don't think this is the fix we need, I believe the problem comes from an inappropriate call to `blockchain_region_v1:h3_to_region/3` with a regionvars from before the var change, which corrupts the cache. We should try to check the vars nonce for the regionvars is the same as the one in the ledger somehow.

vihu:
> I don't think this is the fix we need, I believe the problem comes from an inappropriate call to `blockchain_region_v1:h3_to_region/3` with a regionvars from before the var change, which corrupts the cache. We should try to check the vars nonce for the regionvars is the same as the one in the ledger somehow.

Hmm, fair. Closing this one then.

PR #1493, opened 2022-11-29
Fix ledger exporter

vihu:
- Add token type support
- Remove specific security entry fun
- Export individual funs


abhay:
is this ready to merge?

src/ledger/v1/blockchain_ledger_exporter_v1.erl
```diff
@@ -34,36 +84,261 @@ export_accounts(Ledger) ->
...
-              {location, blockchain_ledger_gateway_v2:location(Gateway)},
-              {nonce, blockchain_ledger_gateway_v2:nonce(Gateway)}] | Acc]
-        end,
-        [],
-        maps:to_list(blockchain_ledger_v1:active_gateways(Ledger))
-    ).
+            Loc =
+            case blockchain_ledger_gateway_v2:location(Gateway) of
+                undefined -> "null";
+                L -> integer_to_list(L)
```
Vagabond, line 90: would it not be better to use h3:to_string() or whatever to get the hexadecimal h3 string here?

Vagabond (approved):
If this is producing the correct output, then I'm ok with merging it.

PR #1496, opened 2022-12-05
Handle off-chain oracle poc rewards by not paying anything for them

Vagabond:


src/transactions/v2/blockchain_txn_rewards_v2.erl
```diff
@@ -677,7 +677,10 @@ prepare_rewards_v2_txns(Results, Ledger) ->
                                                                  fun(Balance) -> Balance + Amt end,
                                                                  Amt,
                                                                  Acc)
-                                         end
+                                         end;
+                                    {oracle, _Type, _O} ->
+                                        %% no rewards for offchain poc oracles
```
jeffgrunewald, line 682: perhaps "no _hnt_ rewards for offchain poc oracles" as i believe it is accounted for in the hip 51/52 emissions curves for offchain oracles to be rewards iot/mobile tokens for their operators' efforts?

jeffgrunewald (approved):

src/transactions/v2/blockchain_txn_rewards_v2.erl
Vagabond: yes, but that will never happen in this codebase

PR #1502, opened 2022-12-19
define custom regions

mikev:
Problem Statement:
Often a country or several countries need a custom region, which defines slightly different parameters. For example in Malaysia any frequencies in 924 Mhz are not legal.

Solution:
So we defined a new region AS923_1B which did not use frequency within 924 Mhz. Now Kenya needs this same custom region, because some of our EU868 frequencies are not legal. This will be a recurring pattern. The PR, review, commit process is painful so lets just do this once across all our repos and define EU868_A to _F and AS923_1A .. 1F

Only AS923 and EU868 will require custom regions.


ke6jjj (approved):

madninja:
Note that this unlikely be used in a miner release, unless we need to do another miner release in January, 2023 right?

PR #1519, opened 2023-01-23
Add a chain var to isolate var txns to their own blockchain_vars

Vagabond:
Due to several observed cache issues when a var txn appears in a block with other transactions that may have been affected by the var changing, this variable has been introduced to force var txns to appear in their own block. This is done as a var so that it can only be activated once a majority of the consensus group understands the var, so a split block will not be produced.

### DIFF
--- a/src/transactions/v1/blockchain_txn_vars_v1.erl
+++ b/src/transactions/v1/blockchain_txn_vars_v1.erl
@@ -1647,6 +1647,12 @@ validate_var(?increment_var_nonce_in_rescue_block, Value) ->
         _ -> throw({error, {invalid_increment_var_nonce_in_rescue_block, Value}})
     end;
 
+validate_var(?isolate_var_txns, Value) ->
+    case Value of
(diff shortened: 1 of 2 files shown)


src/transactions/v1/blockchain_txn_vars_v1.erl
```diff
@@ -1633,6 +1633,12 @@ validate_var(?zero_reward_shares_fix, Value) ->
         _ -> throw({error, {invalid_zero_reward_shares_fix, Value}})
     end;
 
+validate_var(?isolate_var_txns, Value) ->
+    case Value of
+        Val when is_boolean(Val) -> ok;
+        _ -> throw({error, {invalid_zero_reward_shares_fix, Value}})
```
vihu, line 1639:
```suggestion
        _ -> throw({error, {invalid_isolate_var_txns, Value}})
```

vihu (approved):

PR #1525, opened 2023-02-01
include witness invalid reason in tagged witnesses

andymck:
Enable explorer, via ETL, to depict the reason a witness is invalid.  The iot injector is overriding the witness_v1 pb's channel field to represent the invalid reason.  Cast that int to a binary message for use on explorer

linked injector pr: https://github.com/helium/oracles/pull/325

NOTE: this is a known hack.  It is only meant to be temporary and will be dropped post current l1 transition


vihu (approved):

ke6jjj (approved):
Checked these constants against poc_lora.proto. They match.

Vagabond (approved):

